PR #7, opened 2017-03-08
add some minor improvements to code/docs

GlenDC:
These are pretty trivial and minor code/doc changes that I did, while reading through the codebase to understand it all better. I think they are save to merge in, but it's probably a good idea if they pass through a second pair of eyes, so feel free to review and merge this in @robvanmieghem.


nbdserver/redispool.go
```diff
@@ -63,9 +63,10 @@ func (p *RedisPool) Get(connectionString string) redis.Conn {
 // Close releases the resources used by the pool.
 func (p *RedisPool) Close() {
 	p.lock.Lock()
+	defer p.lock.Unlock()
+
```
robvanmieghem: pfew, was pretty bad, this one

PR #44, opened 2017-04-07
add content ref. and gc to dedupedStorage

### LINKED ISSUE #42: Keep track of volumes referencing the content and clean up if 0
Only track 10 volumes, if more than 10, we never delete.

GlenDC:
this PR fixes #42 

**TODO**:

- [x] add support for SET operations in the Redis STUB;
- [x] ensure that referencing is updated at all times (and add unit-tests to guarantee this):
   - [x] SetContent: reference new content, dereference previous content if it existed;
   - [x] DelContent: dereference content;
   - [x] MergeContent: reference new merged content, dereference previous content if it existed;
- [x] ensure that a single volume can reference content multiple times (using a `hashMap` and `HINCRBY` might be better at this point);
- [x] benchmark & investigate impact this PR has on the performance;
- [x] adapt design and implementation where needed;

**NOTE**: Not sure what to do with the copy volume lua script. As any volumes creates with that script, are not added to the reference sets for now, and thus these volumes might reference non-existing content, at some point.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -34,7 +34,7 @@
 		},
 		{
 			"ImportPath": "github.com/g8os/gonbdserver/nbd",
-			"Rev": "585d2378301579044ed1d2bac176b5ba431d56f3"
+			"Rev": "b6a81a628665689535d04401b61a5773dc0fd836"
 		},
 		{
 			"ImportPath": "github.com/garyburd/redigo/internal",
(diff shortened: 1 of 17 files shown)


GlenDC:
At this point I start to think it becomes cheaper to add the logic to the `LBA` implementation. However it feels a bit dirty, as it would mean it knows about the logic of `dedupedStorage` and how its storing the content.

GlenDC:
I'm not really happy with the current implementation of the WIP state of this PR. But if we do end up going with this _exact_ approach, we can no longer use the in-memory Redis stub client, as it does not implement the `EVAL` and `EVALSHA` command, and I do not think we really can/want to support that command, as that starts to become really a (big) project of its own. Seems like a rabbit hole that never ends (or at least not soon). That command is currently required because I'm using some scripts to speed up processing. But I think the design is getting pretty convoluted, and it might be time to go back to the design board and see how we can organize this better and cleaner. Or perhaps it's all fine, and I'm overreacting.

GlenDC:
Rebased latest of master into this PR's branch. Merge conflicts resolved.

GlenDC:
Had an idea and implemented it to try it out, but it turns out to actually make this PR worse. Giving the patch as a diff below, to track it in the history, but it really doesn't seem to be worth it. The goal was to store the block and reference counters all in 1 map, rather than have the content as a separate key. The goal was that this would reduce space, but in fact it seems to require 0.04G more space in my test (1.15G instead of 1.11G). It does allow us to allocate less heap memory in Go, as the reference counter key would no longer needed to be computed on the fly, but I guess that work is now simply delegated to redis. Performance wise we also don't really seem to get any improvement (which was more or less expected).

This patch does make it easier to copy the content and reference counting all at once, and less error-prone, but that's about as its benefits seem to take us.

**TLDR**: had an idea, implemented it, didn't work, giving patch as a diff anyway;

```diff
diff --git a/nbdserver/ardb/deduped.go b/nbdserver/ardb/deduped.go
index cf7cfff..ff26c68 100644
--- a/nbdserver/ardb/deduped.go
+++ b/nbdserver/ardb/deduped.go
@@ -182,7 +182,7 @@ func (ds *dedupedStorage) getContent(hash lba.Hash) (content []byte, err error)
 	}
 	defer conn.Close()
 
-	content, err = redis.Bytes(conn.Do("GET", hash))
+	content, err = redis.Bytes(conn.Do("HGET", hash, dedupedContentField))
 	// This could happen in case the block doesn't exist,
 	// or in case the block is a nullblock.
 	// in both cases we want to simply return it as a null block.
@@ -205,7 +205,7 @@ func (ds *dedupedStorage) setContent(prevHash, curHash lba.Hash, content []byte)
 
 		exists, err := redis.Bool(conn.Do("EXISTS", curHash))
 		if err == nil && !exists {
-			_, err = conn.Do("SET", curHash, content)
+			_, err = conn.Do("HSET", curHash, dedupedContentField, content)
 		}
 
 		return
@@ -240,7 +240,7 @@ func (ds *dedupedStorage) goReferenceContent(hash lba.Hash) (err error) {
 	}
 	defer conn.Close()
 
-	_, err = conn.Do("HINCRBY", dsReferenceKey(hash), ds.vdiskID, 1)
+	_, err = conn.Do("HINCRBY", hash, ds.vdiskID, 1)
 	return
 }
 
@@ -254,29 +254,18 @@ func (ds *dedupedStorage) dereferenceContent(hash lba.Hash) {
 
 // receiver of the dereference content action
 func (ds *dedupedStorage) goDereferenceContent(hash lba.Hash) (err error) {
-	key := dsReferenceKey(hash)
-
 	conn, err := ds.getRedisConnection(hash)
 	if err != nil {
 		return
 	}
 	defer conn.Close()
 
-	_, err = delContentScript.Do(conn, key, ds.vdiskID, hash)
-	return
-}
-
-// get the unique key for a reference content set
-func dsReferenceKey(hash lba.Hash) (key []byte) {
-	key = make([]byte, rcKeyPrefixLength+lba.HashSize)
-	copy(key, rcKeyPrefix[:])
-	copy(key[rcKeyPrefixLength:], hash[:])
+	_, err = delContentScript.Do(conn, hash, ds.vdiskID)
 	return
 }
 
 const (
-	rcKeyPrefix       = "rc:"
-	rcKeyPrefixLength = len(rcKeyPrefix)
+	dedupedContentField = "+deduped+block+"
 )
 
 type rcAction struct {
@@ -296,17 +285,16 @@ const (
 // lua script used to dereference a vdisk from a hash,
 // and delete the content if the new reference count is < 1
 var delContentScript = redis.NewScript(0, `
-local key = ARGV[1]
+local hash = ARGV[1]
 local vdiskID = ARGV[2]
-local hash = ARGV[3]
 
-redis.call('HINCRBY', key, vdiskID, -1)
-if tonumber(redis.call('HGET', key, vdiskID)) < 1 then
-	redis.call('HDEL', key, vdiskID)
+redis.call('HINCRBY', hash, vdiskID, -1)
+if tonumber(redis.call('HGET', hash, vdiskID)) < 1 then
+	redis.call('HDEL', hash, vdiskID)
 end
 
-local rc = redis.call('HLEN', key)
-if rc < 1 then
+local rc = redis.call('HLEN', hash)
+if rc < 2 then
 	redis.call('DEL', hash, key)
 end
 
```

(profiling while running command: `qemu-img convert -p -n ubuntu-16.04-server-cloudimg-amd64-disk1.img -O nbd nbd+tcp://localhost:6666/a`)

GlenDC:
@robvanmieghem added a new commit, which improves the reference counting. It is now faster and takes less space in Redis. It's however still (40%) slower than the current master branch version.

GlenDC:
@robvanmieghem ok some more improvements. Making sure that all hashes are now send as a bytes. We also now don't read replies from a redis.Conn, if we don't need it.

It's in pretty OK state now I think. Feel free to review the PR as it is. As the deadline is today.

> if more than 10, we never delete.

This isn't a thing for now. Is this important? Because if it is, I'll need to add this feature to this PR.

GlenDC:
With the current state of this PR we also no longer require any Lua scripts. We could bring it back for the dereference logic, but I think the way it is right now it is probably OK.

PR #50, opened 2017-04-12
WIP: add redis stub (using ledisdb)

GlenDC:
**WIP**

fixes #46

PR #51, opened 2017-04-12
WIP: add redisstub, using an embedded ledisdb

### LINKED ISSUE #46: Better redis compliancy for the inmemory stub.
Currently the in-memory Redis stub client doesn't support pipelining.

+ `MULTI` and `EXEC` commands are recognized but ignored;
+ `SEND` only supports a small subset of the supported `Do` commands;
+ `SEND` does not collect its results for later use (eg. when `EXEC` gets executed);
+ `Flush` doesn't do anything either for now;

Support for this would be greatly welcome for the WIP code of https://github.com/g8os/blockstor/pull/44.

**EDIT**:

TODO:

- [x] Create InMemoryRedis Conn, using [LedisDB](https://github.com/siddontang/ledisdb);
  - [x] Support Lua Scripting (`EVAL` command);
  - [x] Support pipelining (`Send`, `Flush`, `MULTI`, `EXEC` (fake it));
- [x] Support `redis` global variable _ledisdb_'s Lua support (as an alias of `ledis` if possible), see: https://github.com/siddontang/ledisdb/pull/289 (merged);
- [x] Replace [CGO Lua](https://github.com/siddontang/golua) (in LedisDB) with [Gopher Lua](https://github.com/yuin/gopher-lua) (see: https://github.com/siddontang/ledisdb/issues/288) (PR: https://github.com/siddontang/ledisdb/pull/291);
- [x] Allocate a port dynamically for ledisdb, when no address is specified (see: https://github.com/siddontang/ledisdb/issues/290) (PR: https://github.com/siddontang/ledisdb/pull/292);
- [ ] Improve InMemoryRedis performance, as it seems to account for most of the cpu usage when used (ledisdb, and especially `net(*conn).Write/Read`, seem responsible) (solution proposed in open PR);
- [x] Vendor dependencies;

GlenDC:
**WIP**

replaces and closes PR #50

fixes #46


GlenDC:
With this PR merged, we are limited to the commands available in ledis, which is a subset of the available redis commands. The full list of supported commands by ledis can be found at: https://github.com/siddontang/ledisdb/blob/master/doc/commands.md

In theory this shouldn't ever be a problem, as I don't think we'll ever need a command that is not supported by ledis. In practice it might mean that we might need to contribute in the future to ledis (and goleveldb), to support the missing redis command we desire so much.

GlenDC:
This PR is good to go, except for the fact that the performance of the `memorystorage` option using the embedded `ledisdb` really seems to be awful.

Using `redis-server`:

```
# time
qemu-img convert -p -n ubuntu-16.04-server-cloudimg-amd64-disk1.img -O nbd
3.95s user 1.32s system 22% cpu 23.941 total
```

Using the currently embedded `ledisdb` as integrated in this PR:

```
# time
qemu-img convert -p -n ubuntu-16.04-server-cloudimg-amd64-disk1.img -O nbd
4.49s user 1.82s system 8% cpu 1:18.24 total
```

When profiling the heap we can see that almost all of the storage goes to ledisdb and its underlying backend (goleveldb), which is not surprising:

screen shot 2017-04-17 at 11 27 31

When profiling the CPU you'll notice that about 50% of the available CPU power for this process goes to `net.(*conn).Write+Read` (, which is not surprising as even though the ledisdb service is embedded, it is still all going over a TCP socket. Even worse is that now both the reading and the writing happens in the same process.

@robvanmieghem I do wonder if it wouldn't be worth merging over my RedisStub struct (which conforms to the `redigo.Conn` interface, as that would allow us to let our nbdserver talk _directly_ to _ledisdb_ rather than going via a TCP connection (which hogs most of the CPU power at the moment). On a very powerful machine this problem might be neglectable of course, but on my dev machine (2.7 GHz Intel Core i5; 8 GB 1867 MHz DDR3) I do experience it as a world of difference :|

What do you think?

robvanmieghem:
@GlenDC 
Honestly, I like it this way, it keeps the memory implementation very similar to the normal one. An improvement using a non tcp connection would increase performance but is never to be used in real life scenario's anyway.

GlenDC:
Fair enough. It definitely was a non-blocking issue. Would it become a problem we can always resolve it as a separate issue.

PR #54, opened 2017-04-16
Fix better logging

### LINKED ISSUE #52: Echo commandline parameters
For debugging purposes, it would be better to echo the commandline parameters to the output. The nbdserver is started in a g8os and if two listen on the same socket for example, this would immediately show in the logs this way.

GlenDC:
This PR aims to simply logging, while add the same time add the capability to log Debug statements.

This PR also fixes #52 by logging nbdserver's command line arguments in debug mode.


GlenDC:
@robvanmieghem if this is not an approach you are happy with, I'm more than happy to talk about it.

The standard logging package is sadly not sufficient as it does not allow for a separation of debug and info logs. Which is odd as it does allow for other weird Log levels. I looked through all the available log packages, and none of them keep it simple, only adding more log levels. On top of that they omit the feature of being able to log the file and line number where the log came from, which is a feature I appreciate very much in logging.

Therefore I "forked" the pkg/log package and improved what I thought was needed. The fork you can find here: http://github.com/glendc/go-mini-log

The end result is pretty elegant I think, and does the job. And we now also echo the command line arguments in nbdserver, which you asked :)

GlenDC:
If you approve this PR and merge it into master. I'll merge the gonbdserver change over to the gonbdserver repo, and update the godep commit (of gonbdserver) in a new commit straight to master (of this repo).

GlenDC:
If you think this change is too invasive or you don't agree with my opinion about how minimal a logging system can/should be, than I can also simply close this PR, to make a simpler change that would just involve fixing issue #52.

We could for example work with a build tag (`debug`), and use that for conditional print statements, of which we only have one for now. Which could result in a patch as simple as the following:

```diff
diff --git a/build/debug_off.go b/build/debug_off.go
new file mode 100644
index 0000000..4c40eea
--- /dev/null
+++ b/build/debug_off.go
@@ -0,0 +1,5 @@
+// +build !debug
+
+package build
+
+const DEBUG = false
diff --git a/build/debug_on.go b/build/debug_on.go
new file mode 100644
index 0000000..e3e9369
--- /dev/null
+++ b/build/debug_on.go
@@ -0,0 +1,5 @@
+// +build debug
+
+package build
+
+const DEBUG = true
\ No newline at end of file
diff --git a/nbdserver/main.go b/nbdserver/main.go
index 2a21457..f9cf43a 100644
--- a/nbdserver/main.go
+++ b/nbdserver/main.go
@@ -17,6 +17,7 @@ import (
 	"github.com/g8os/blockstor/nbdserver/lba"
 	"github.com/g8os/blockstor/storagecluster"
 	"github.com/g8os/gonbdserver/nbd"
+	"github.com/glendc/Sia/build"
 	"golang.org/x/net/context"
 )
 
@@ -44,6 +45,18 @@ func main() {
 		fmt.Sprintf("Cache limit of LBA in bytes, needs to be higher then %d (bytes in 1 shard)", lba.BytesPerShard))
 	flag.Parse()
 
+	if build.DEBUG {
+		log.Printf("[DEBUG] flags parsed: memorystorage=%t tslonly=%t profileaddress=%q protocol=%q address=%q gridapi=%q nondeduped=%q testardbs=%q export=%q lbacachelimit=%d",
+			inMemoryStorage, tslonly,
+			profileAddress,
+			protocol, address,
+			gridapiaddress,
+			nonDedupedExports,
+			testArdbConnectionSrings,
+			exports,
+			lbacachelimit)
+	}
+
 	exportArray := strings.Split(exports, ",")
 	if len(exportArray) == 0 {
 		exportArray = []string{"default"}
```

@robvanmieghem, you let me know with which one of the two approaches (this PR or this patch) you want to go, and I'll make it happen. Or perhaps you have a different (and thus third) approach in mind that you want us to go for.

GlenDC:
@robvanmieghem because logrus does not have the capability to log the files. While with the standard package (and thus this go-mini-log) package we do have that option. I think logging the file (and line number) next to the log is very helpful.

Also logrus ads a lot of extra features, which I don't really think is necessary. Personally I think there should only be 2 different levels. Info and Debug, where info is for logs that remain in code (even in production) and thus could even be seen by the user. And Debug is just for the developers.

I've worked with logrus before though, and it's a nice library. But it's sad that it adds so many features that I don't think are needed, while it loses a very valuable feature, and that is logging the file and number it logs from.

But like I said before, that's just an opinion. If you want to keep the dependencies as low as it is right now. And replace this PR with the patch that I've shared earlier as a comment (the one where we simply use a `DEBUG bool` const, could work just as well.

robvanmieghem:
@GlenDC Ok, you're right. Can you resolve the conflicts for this PR so I can merge it?

GlenDC:
Merged logger into `g8os/gonbdserver` and rebased master into this PR.

PR #64, opened 2017-04-22
add support for root (fallback) storage

### LINKED ISSUE #57: Remote root storagecluster support
For deduped volumes, having a remote storagecluster available enables a newly installed g8os grid to immediately start volumes without the need to fully sync first. 
First the get is tried on the local cluster but when it is not there, it should be fetched from the remote root storage cluster. After a block is fetched from the remote root storagecluster, it should also be stored in the local one so a future fetch would get it from the local cluster directly.

GlenDC:
This PR fixes #57 as a POC. For now it simply gets the root's grid api address from an optional flag passed by the user to the nbdserver CLI.

The new behavior of the deduped's storage `Get` method is now as visualized in following graph:

deduped_get

(the same behavior by the way applies to the getContent phase of the `Merge` method of the deduped storage, due to the shared internal method where this behavior is defined)


iwanbk:
Nice graph, what tool did you use to generate it?

GlenDC:
@iwanbk: https://en.wikipedia.org/wiki/DOT_%28graph_description_language%29, my graph is pretty basic though, I don't make use of all the available options, but it's good enough to get my point/idea across swiftly :)

GlenDC:
@robvanmieghem ah, I do remember you saying that you wanted to be able to specify the ardb connection strings. But I found this a bit strange. As these connection strings are normally given by the grid API. Why? Because they can change in the duration that the nbdserver is up. This POC will therefore break in a production environment I think. In case we want to be able to specify the root ardb connection string (for content storage) we do really want to fetch it dynamically from the grid API I think (as part of the vdisk fetching?).

GlenDC:
@robvanmieghem would it also be OK for me to commit the dot file in this repo, used to generate the graph above? This way we can easily modify it in the future if needed.

robvanmieghem:
Thing is we don't have that central infrastructure in place yet and I don't even know if this is going to be fast enough to be workable anyway.

robvanmieghem:
And indeed, adding the dot file (source of the image) is indeed necessary for other people to be able to modify if needed.

GlenDC:
@robvanmieghem added your changes as requested, and also added the flow graph (inlined above) to newly added documentation about the backendStorage.

PR #77, opened 2017-04-27
replace isa-l with blake2b(go) and go std aes

### LINKED ISSUE #67: tlogserver: replace isa-l with go AES and blake2
Related to some tasks in #49

### LINKED ISSUE #70: make tlog client pure-go by default
Depends on #69, related to #49

GlenDC:
+ isa-l crc32 is replaced with blake2b (go lib minio/blake2b-simd);
+ isa-l_crypto aes_cbc_enc_256 is replaced with Go std aes256+gcm;
+ fixes #67;
+ fixes #70;
+ replace the std `log` pkg in the tlog code with the shared `glendc/go-mini-log` logger;
+ move hash code to the root dir (`blockstor`), previously it was in the nbdserver `lba` module;


.travis.yml
```diff
@@ -7,9 +7,9 @@ go:
 
 install:
   - go get -u github.com/Jumpscale/go-raml
-  - ./tlog/tlogserver/install_dependencies.sh
+  - ./tlog/install_dependencies.sh
 
 script:
-  - go test $(go list ./... | grep -v vendor)
+  - go test $(go list ./... | grep -v vendor | grep -v tlogclient/examples)
```
GlenDC: this is done, because the `tlogclient` examples rely on cgo code

hash.go
```diff
@@ -1,4 +1,4 @@
-package lba
+package blockstor
```
GlenDC: The hash code is moved from LBA to the root `g8os/blockstor` repo. In a way the hashing code has nothing to do with LBA (specifically). So now both the tlog code as well as the nbdserver code makes use of this shared hash code.

.travis.yml
iwanbk: why we can't simply do `go test ./...`?

GlenDC: Because it would also test all code in the vendor folder, which we don't want. And in the travis are 2 lines where we test code, one line that tests the code without any cgo, and the other line where we do use cgo.

iwanbk:
```
Because it would also test all code in the vendor folder, which we don't want
```
there are no *_test.go files in vendor folder.
but it is true that `go test` still execute the vendor folder and then says :" ........[no test files]"

GlenDC: You're correct. I seem to remember that in the past it also took quite a while for the `go test` tool to conclude a directory has no tests, but now it seems to be pretty fast, as I just did a test in our repo and there was only a `0.2s` difference, so I guess we no longer need it.

So those 2 lines could change to:

```yaml
- go test $(go list ./... | grep -v tlogclient/examples)
- go test -tags isal ./...
```

Issue #92, opened 2017-05-02
tlog : use same redis package as nbdserver


iwanbk:

GlenDC:
Already fixed in integration of tlogserver into nbdserver.

@iwanbk is https://github.com/g8os/blockstor/commit/94fe8592515f731d78b5a3eff365d3158b216827 what you mean?

If you want I can already merge that commit into tlogserver, as that is not really specific to my work on integrating tlog into the nbdserver.

iwanbk:
sure

GlenDC:
@iwanbk opened a PR, if that's what you had in mind, you can merge it in. Otherwise I guess you could amend changes to that PR and merge it in afterwards, as I'm going to bed now.

iwanbk:
i merge that PR because that is a good PR and really helps for testing.

But this issue is about using 'https://github.com/g8os/blockstor/tree/master/nbdserver/ardb', instead of using redigo directly.

It is make sense to use that package for tlog?

GlenDC:
I assume you are referring to using [RedisPool](https://github.com/g8os/blockstor/blob/master/nbdserver/ardb/redispool.go#L14-L77)?

Dunno if it makes sense to use tlog, will think about that today, but if it does make sense, we could improve it a bit (if needed), and move it to the root blockstor module, so we can share it.

GlenDC:
@iwanbk I don't think sharing the redispool between nbdserver and tlogserver makes sense. We could copy the logic and adapt where needed, but both projects seem to want different things from what a redispool is supposed to be.

A big difference is for example the fact that tlogserver accesses the pools via an index (`int`), while nbdserver.ardb wants them per connection (`string`).

You could always refactor stuff to make them want similar things, but code sharing isn't always desired, as it could also lead to more static design.

So if you agree @iwanbk , I think we could close this issue.

iwanbk:
```
A big difference is for example the fact that tlogserver accesses the pools via an index (int), while nbdserver.ardb wants them per connection (string).
```
it is no a big difference, we can easily convert int to string.

But i agree with your other points.

I thought the pool provide a kind of failover mechanism as mention in the comment https://github.com/g8os/blockstor/issues/49#issue-221191198.

PR #93, opened 2017-05-02
add tlog server and client

### LINKED ISSUE #49: tlog server implementation in go
This would allow us to share some code between the tlog server and the nbdserver
like the pooling of the ardb's, failover mechanism, storagecluster configuration
...

the embedded ledisdb (on which Glen is working) for inmemory storage during development,...

and publishing the hashses to the rivine blockchain could import the required libraries directly from rivine without the need for reimplementation

since that is implemented in go as well

iwanbk:
Closes #49 

Some of the sub task at #49 are moved to new issue because i don't think anymore that we must do it.


GlenDC:
Are you sure this is already ready to merge into master? I thought we were still doing work on this? I'm certainly still working on some feature.

Also I think before this can be merged, we should clean up the commit history of this branch, as it's now pretty messy.

iwanbk:
```
Are you sure this is already ready to merge into master?
```
Yes, it won't break anything. And make the integration to be much easier because the ndbserver master still changing.

```
 I thought we were still doing work on this
```
Yes, we will always working on this.

```
Also I think before this can be merged, we should clean up the commit history of this branch, as it's now pretty messy.
```
right

GlenDC:
@iwanbk ok that does make sense.

I'm ok with this being merged into master then, as long as you clean up the commit history a bit, as it's now pretty messy as I said (especially all those commits at the beginning).

GlenDC:
BTW, I do think that it would be cool if issue #92 could be resolved before merging in this PR. I opened a PR for it.

iwanbk:
```
(especially all those commits at the beginning).
```
yes, it is very messy.
As it supposed to be cleaned before sending the PR, i never had any problem this way before but got many many conflicts in this branch, that involved code not written by me.
It seems because this branch was modified not only by me and involve rebase/merge from master.

 The command i use to squashed it
```
git rebase -i ced30d1b4c0c3b749603ec3fdbb78324c2d4bbd9
```
i got the hash from 
```
$ git merge-base tlogserver master
ced30d1b4c0c3b749603ec3fdbb78324c2d4bbd9
```

If someone really knows how to handle it, please do it or tell me. 

If not, we can merge this to master

GlenDC:
Yeah I don't think we can (and should) all squash it into one commit. I think it's better to squash the commits that clearly belong together, but as this is a very big PR, it's ok that there are a couple of different commits. Going to work on merging this later today.

iwanbk:
yep, i didn't try to make it a single commit.
I only tried to squash the commits from the first day (IIRC April 13th).

GlenDC:
Cleaned up the tlogserver branch as much as I ought reasonable. I left alone all master-to-tlogserver merge commits as to not mess up.

Will do a final self-review after lunch, and will merge it into master afterwards.

A backup (of the old tlogserver) branch is pushed to https://github.com/g8os/blockstor/tree/tlogserver-backup, that one will remain online for the next week or so, in case we notice something wrong because of the tlogserver-history-rewriting.

Should be fine though, as there weren't any unexpected merge conflicts. Only the ones that were already there.

tlog/install_dependencies.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+BUILD_DIR=/tmp/build_tlog
+rm -rf $BUILD_DIR
+mkdir $BUILD_DIR
+
+
+sudo apt-get install build-essential autoconf automake nasm yasm libsnappy-dev
```
GlenDC: Only works for Debian for now. This script should also work for MacOS and other required platforms once we know for sure that we know for sure that we want to keep the CGO code.

tlog/tlogclient/README.md
```diff
@@ -0,0 +1,17 @@
+# TLOG Client
+
+- TLOG Client library is a client lib that works on GIG BLOCKSTOR (NBD Server)
+- TLOG Client communicates with TLOG Server via dpdk stack.
+- Communication between client and server use binary capnp stream (no RPC).
+
+## Data send from the client to server:
+```
+Block:
```
GlenDC: Actualize these docs, as they seem to be out of date and incorrect.

tlog/tlogclient/client.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+// Send sends the transaction tlog to server.
+// It returns error in these cases:
+// - broken network
+// - failed to send all tlog
+// in case of errors, client is not in valid state,
+// shouldn't be used anymore
+func (c *Client) Send(vdiskID string, seq uint64,
+	lba, timestamp uint64, data []byte) error {
```
GlenDC: should move to Line 156. This is the only line in the entire codebase that we break up like this, so better be consistent and combine line 156 and 157 into one.

tlog/tlogclient/decoder/decoder.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// get the real len
+	var realLen uint64
+	if err := binary.Read(bytes.NewBuffer(merged), binary.LittleEndian, &realLen); err != nil {
+		return nil, err
+	}
+	merged = merged[8 : 8+realLen] /* 8 is the size of uint64 */
```
GlenDC: 8 should move to a constant and calculated using sizeof

PR #108, opened 2017-05-04
tlog : make the default config to not return random values.

iwanbk:
Make the default config to return constant instead of random value.
I think random value doesn't help anyone, except the embedded ledisdb, that supposed to be used only in unit test.

If we use constant, we can reuse the default config among other binary/component:
- decoder
- send tlog
- walk_history
- tlogreplay

### DIFF
--- a/tlog/tlogserver/server/config.go
+++ b/tlog/tlogserver/server/config.go
@@ -1,35 +1,25 @@
 package server
 
 import (
-	"crypto/rand"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"net"
 	"strconv"
 
 	"github.com/g8os/blockstor/redisstub"
-	"github.com/g8os/blockstor/tlog"
 	log "github.com/glendc/go-mini-log"
 )
 
 // DefaultConfig creates a new config, using sane defaults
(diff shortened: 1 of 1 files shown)


GlenDC:
@iwanbk sure, but a default private key and nonce is a potential attack vector, as that will be something that an attacker can easily try out, in case a production server didn't explicitly define a private key and nonce... At least with a random default, there is no possibility of that.

GlenDC:
On the other hand, if those values are random, than they are also pretty useless as we would have to print them out in case they are not set by the user.

@robvanmieghem you have any opinion about this? Otherwise I'll just merge this. It's not too important for now anyhow.

iwanbk:
```
but a default private key and nonce is a potential attack vector, as that will be something that an attacker can easily try out, in case a production server didn't explicitly define a private key and nonce... At least with a random default, there is no possibility of that.
```
yeah, i understand it

```
On the other hand, if those values are random, than they are also pretty useless as we would have to print them out in case they are not set by the user.
```
very true! Random value is useless.

Other alternative is to not give default value and return error if user didn't set it.

PR #109, opened 2017-05-04
Make the log optional for ardb backend.

iwanbk:
Sometimes we don't need the tlog, for example in tlog replay.


GlenDC:
@iwanbk what happened to this PR?

iwanbk:
I need it for tlogreplay and i don't think it needed by the other.
So i close it and included it in tlogreplay branch.
You can merge it if you want.

PR #110, opened 2017-05-04
Move tlogserver main file dir from cmd to tlog/tlogserver.

iwanbk:
move back the dir of main file to tlog/tlogserver.
reason:
- consistent with nbdserver
- moving it to cmd/tlogserver make the code browsing more difficult.


GlenDC:
Ah dunno, I was actually planning to work the other way, and also move nbdserver to a cmd folder. Feels nicer to me. But I'm not anal about this, and it's not important. So I'm fine with this :)

Issue #86, opened 2017-04-28
Tool to recreate a vdisk from the tlogs


robvanmieghem:

iwanbk:
depend on #40

iwanbk:
I can see two options to implement this:

(1) the tool work as nbd client
 it connects to nbdserver and replay the tlogs as nbd client
pros:
- simple, clean, and easy

cons:
- slower because of communication overhead between nbd server & client
- it seems we need to implement nbd client in Go first

(2) use `nbdserver.ardb.Backend.WriteAt` directly
as defined at https://github.com/g8os/blockstor/blob/3588ab4e349a5318e024b472362fd1e52b384b81/nbdserver/ardb/backend.go#L19

pros:
- faster

cons:
- this tool might need to know a lot about nbdserver internals especially the configurations

iwanbk:
just found an ndb client library https://github.com/akmistry/go-nbd

iwanbk:
```
just found an ndb client library https://github.com/akmistry/go-nbd
```
this lib doesn't support tcp & unix socket, i'm checking if we can improve it.

robvanmieghem:
If we go this way, it would make the tool complex to run since you would need to start up a seperate nbdserver instance with the right configuration and write everything through the nbdclient.
I understand when using backend.WriteAt directly, it requires some more knowledge about the nbdserver internals but maybe we can create a simpler factorymethod for it so that it becomes just as easy to create such a backend as starting up an nbdserver with the same parameters would be.

iwanbk:
```
If we go this way, it would make the tool complex to run since you would need to start up a seperate nbdserver instance with the right configuration and write everything through the nbdclient.
```
yes, i knew it.
and i thought that it makes things simpler because the user doesn't need to learn about this tool, he only need to knows how to configure nbd server.
But maybe i missed something here

```
I understand when using backend.WriteAt directly, it requires some more knowledge about the nbdserver internals but maybe we can create a simpler factorymethod for it so that it becomes just as easy to create such a backend as starting up an nbdserver with the same parameters would be.
```
ok, checking it.

iwanbk:
it seems we already have the factory

iwanbk:
I managed to create the backend, i'll see if we can make it simpler.

Still need to wait for #40.

From my quick check, the integration prototype at https://github.com/g8os/blockstor/tree/tlog-integration-into-nbdserver should work fine. will check it.

GlenDC:
[...]

PR #115, opened 2017-05-05
Tlogreplay : Tool to recreate a vdisk from the tlogs

iwanbk:
Implement #86.


iwanbk:
This PR modify nbdserver code in these area:
- add additional info to send to tlogserver
- make the tlogserver optional for ardb backend.

nbdserver/ardb/backend.go
```diff
@@ -116,7 +117,7 @@ func (ab *Backend) WriteZeroesAt(ctx context.Context, offset, length int64, fua
 
 	// send tlog async
 	if ab.tlogClient != nil {
-		ab.sendTransaction(uint64(blockIndex), content)
+		ab.sendTransaction(uint64(blockIndex*ab.blockSize), schema.OpWriteZeroesAt, nil, uint64(length))
```
GlenDC: @iwanbk right now the `offsetInsideBlock` is ignored when sending the transaction. So your replay might be incorrect, in case you are writing zeroes on a non-0 (local block) offset. Or am I missing something?

nbdserver/ardb/backend.go
```diff
@@ -76,7 +79,7 @@ func (ab *Backend) WriteAt(ctx context.Context, b []byte, offset int64, fua bool
 
 	// send tlog async
 	if ab.tlogClient != nil {
-		ab.sendTransaction(uint64(blockIndex), content)
+		ab.sendTransaction(uint64(blockIndex*ab.blockSize), schema.OpWrite, content, uint64(len(content)))
```
GlenDC: @iwanbk right now the `offsetInsideBlock` is ignored when sending the transaction. So your replay might be incorrect, in case you are writing content on a non-0 (local block) offset. Or am I missing something?

iwanbk: we don't need `offsetInsideBlock` because it is in the `WriteAt` internal.
What we need is all values needed to call `WriteAt` again.

nbdserver/ardb/backend.go
iwanbk: same as `WriteAt` case, but indeed the `length` there is incorrect.

I also just discovered that writing all zero doesn't make the `WriteZeroesAt` to be called.
So, this part of code was not tested.

nbdserver/ardb/backend.go
GlenDC: In that case you would still need to store the offset, would just be not a local offset.

iwanbk: Yes, i already store the offset.
With offsetInsideBlock==0

PR #116, opened 2017-05-05
Make tlog optional for nbdserver.

iwanbk:
nbdserver currently create tlogserver with embedded ledisdb if `tlogrpc` is left empty.

We don't always need the tlog, and we only need that kind of tlogserver for dev/test.

This PR change the behavior of `tlogrpc` nbdserver command option, we need to set it to `auto` to create that kind of tlogserver.

### DIFF
--- a/nbdserver/main.go
+++ b/nbdserver/main.go
@@ -45,7 +45,7 @@ func main() {
 	flag.StringVar(&protocol, "protocol", "unix", "Protocol to listen on, 'tcp' or 'unix'")
 	flag.StringVar(&address, "address", "/tmp/nbd-socket", "Address to listen on, unix socket or tcp address, ':6666' for example")
 	flag.StringVar(&gridapiaddress, "gridapi", "", "Address of the grid api REST API, leave empty to use the embedded stub")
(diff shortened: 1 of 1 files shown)


nbdserver/main.go
```diff
@@ -45,7 +45,7 @@ func main() {
 	flag.StringVar(&protocol, "protocol", "unix", "Protocol to listen on, 'tcp' or 'unix'")
 	flag.StringVar(&address, "address", "/tmp/nbd-socket", "Address to listen on, unix socket or tcp address, ':6666' for example")
 	flag.StringVar(&gridapiaddress, "gridapi", "", "Address of the grid api REST API, leave empty to use the embedded stub")
-	flag.StringVar(&tlogrpcaddress, "tlogrpc", "", "Address of the tlog RPC, leave empty to use the inmemory version (test/dev only)")
+	flag.StringVar(&tlogrpcaddress, "tlogrpc", "", "Address of the tlog RPC, set to 'auto' to use the inmemory version (test/dev only)")
```
GlenDC: I'm not too convinced by this value, I would definitely make it a constant, but even so, I'm not too convinced. However for now I'm OK with it.

iwanbk:
```
I would definitely make it a constant
```
good catch!
of course you can `ask for change` for something like this.

PR #127, opened 2017-05-08
Tlog async flush

### LINKED ISSUE #106: tlog : do the flush asynchronously
This way the client user need to call the `RecvOne` from different goroutine or using the `Recv`
https://github.com/g8os/blockstor/blob/25c2691721f007c81f11531870725f7ea1f0ace0/tlog/tlogclient/client.go#L52

iwanbk:
Closes #106

the idea is from https://github.com/g8os/blockstor/issues/106#issuecomment-299394069
```
The alternative is to do async by default and then go to sync if the buffers reach some value
```
It removes tlogtab.go and use Go channel for tlog block buffering. It has these advantages:
- no need for lock
- buffered channel (with limited size) keep our RAM from exploding.


tlog/tlogserver/server/flusher.go
```diff
@@ -266,3 +275,33 @@ func (f *flusher) encodeCapnp(vdiskID string, blocks []*schema.TlogBlock, lastHa
 
 	return buf.Bytes(), err
 }
+
+func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
+	rc := f.redisPool.Get(0)
+	defer rc.Close()
+
+	hash, err := redis.Bytes(rc.Do("GET", f.lastHashKey(vdiskID)))
+	if err != nil && err != redis.ErrNil {
```
GlenDC: I feel like it would be clearer if we would change these lines to:

```go
if err == redis.ErrNil {
    err = nil
    hash = f.privKey
}
if err == nil {
    return nil, err
}
```

In fact, in the ardb module we in general don't consider nil values to be an error, which seems to be the case here as well, so in there we have a private util function for that: https://github.com/g8os/blockstor/blob/master/nbdserver/ardb/storage.go#L21-L33

tlog/tlogserver/server/server.go
```diff
@@ -97,23 +97,44 @@ func (s *Server) handle(conn net.Conn) error {
...
 		}
 
+		// vdiskID
 		vdiskID, err := tlb.VdiskID()
 		if err != nil {
 			log.Debugf("failed to get vdisk ID: %v", err)
 			return err
 		}
 
+		if tlc == nil {
```
GlenDC: I might be missing something, but this logic seem to be flawed.

If my assumptions are right, this logic will break as soon as there are at least 2 vdisks in use with different identifiers. Currently you seem to cache a tlog chan, but I think you should always use `tlc, err = s.f.getTlogChan(vdiskID)`. In theory you could cache the last used channel I guess, so that consecutive server calls will be faster, but not sure if that really will be faster or worth it.

Either way, the way it is now, we'll be sending the content onto the wrong channel, as the currently used tlog channel isn't updated once it is created once.

tlog/tlogserver/server/server.go
```diff
@@ -80,7 +80,7 @@ func (s *Server) ListenAddr() string {
 }
 
 func (s *Server) handle(conn net.Conn) error {
```
GlenDC: I think this server handle func should get a context as its first argument, such that this for loop can quite in case the context is done.

tlog/tlogserver/server/server.go
```diff
@@ -97,23 +97,44 @@ func (s *Server) handle(conn net.Conn) error {
...
 			return err
 		}
 
+		if tlc == nil {
+			tlc, err = s.f.getTlogChan(vdiskID)
+			if err != nil {
+				log.Infof("failed to get tlog channel of vdisk:%v, err: %v", vdiskID, err)
+				return err
+			}
+			go s.sendResp(conn, vdiskID, tlc.respChan)
```
GlenDC: Nothing seems to wait for this function to be done, in case the function stops to exist. Perhaps that's fine, but perhaps not.

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
-// store a tlog message and check if we can flush.
-func (f *flusher) store(tlb *schema.TlogBlock, vdiskID string) *response {
-	// add blocks to tlog table
-	tab := f.getTlogTab(vdiskID)
-	tab.Add(tlb)
-
-	// check if we can do flush and do it
-	seqs, err := f.checkDoFlush(vdiskID, tab, false)
+	// get last hash from storage
+	lastHash, err := f.getLastHash(vdiskID)
```
iwanbk: sure

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
-		return &response{
-			Status: -1,
-		}
-	}
-
-	status := int8(1)
-	if err == nil && (seqs == nil || len(seqs) == 0) {
-		seqs = []uint64{tlb.Sequence()}
-		status = 0
+		return &tlc, err
```
iwanbk: yep

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
+	return &tlc, nil
 }
 
-// check if we can flush and do it
-func (f *flusher) checkDoFlush(vdiskID string, tab *tlogTab, periodic bool) ([]uint64, error) {
-	blocks, needFlush := tab.Pick(f.flushSize, f.flushTime, periodic)
-	if !needFlush {
-		return []uint64{}, nil
-	}
+// it is  flusher routine that do the flush asyncrhonously
```
iwanbk: thanks

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
+			}
+			toFlushLen = f.flushSize
 
-	seqs, err := f.flush(vdiskID, blocks[:], tab)
-	if err != nil {
-		log.Infof("flush failed vor vdisk ID:%v, err:%v", vdiskID, err)
-	}
-	return seqs, err
-}
+			pfTimer.Stop()
```
iwanbk: it is not stopped forever.

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
+			toFlushLen = len(tlogs)
+		}
 
-func (f *flusher) flush(vdiskID string, blocks []*schema.TlogBlock, tab *tlogTab) ([]uint64, error) {
-	numFlush++
-	log.Debugf("flush %v vdiskID=%v", numFlush, vdiskID)
-	// get last hash
+		// get the blocks
+		blocks := tlogs[:toFlushLen]
+		tlogs = tlogs[toFlushLen:]
```
iwanbk: nope, please see the code above it

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
+		var status int8
 
-		return tab.getLastHash(rc)
-	}()
-	if err != nil && err != redis.ErrNil {
-		return nil, err
-	}
+		lastHash, seqs, err = f.flush(vdiskID, blocks[:], lastHash)
+		if err != nil {
+			log.Infof("flush %v failed:%v", vdiskID, err)
```
iwanbk: sure :)

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
-	if err == redis.ErrNil {
-		lastHash = f.privKey
+		tlc.respChan <- &response{
+			Status:    status,
+			Sequences: seqs,
+		}
 	}
+}
 
+func (f *flusher) flush(vdiskID string, blocks []*schema.TlogBlock, lastHash []byte) ([]byte, []uint64, error) {
```
iwanbk: i don't think that named result parameter is a good idea for long func. See https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

```
Naked returns are okay if the function is a handful of lines. Once it's a medium sized function, be explicit with your return values.
```

tlog/tlogserver/server/flusher.go
```diff
@@ -55,100 +72,95 @@ func newFlusher(conf *Config) (*flusher, error) {
...
 	// capnp -> byte
 	data, err := f.encodeCapnp(vdiskID, blocks[:], lastHash)
 	if err != nil {
-		return nil, err
+		return lastHash, nil, err
```
iwanbk: same as previous

tlog/tlogserver/server/flusher.go
```diff
@@ -215,13 +228,9 @@ func (f *flusher) storeEncoded(vdiskID string, key []byte, encoded [][]byte, tab
...
 		defer wg.Done()
-
-		rc := f.redisPool.Get(0)
-		defer rc.Close()
-
-		err := tab.storeLastHash(rc, key)
+		err := f.storeLastHash(vdiskID, key)
 		if err != nil {
-			log.Debugf("error when setting last hash name: %v", err)
+			log.Infof("error when setting last hash name: %v", err)
```
iwanbk: I actually want to use 'Errorf', but it is not exist in the logger package.

GlenDC: Yeah because I don't believe in Error logging, neither warning logging. Warnings are demoted errors, and errors need to be handled by code. If an error can't be handled it is fatal. But if it is handled, it is no longer an error, and thus is simply to be logged for debug/info reasons.

tlog/tlogserver/server/flusher.go
```diff
@@ -215,13 +228,9 @@ func (f *flusher) storeEncoded(vdiskID string, key []byte, encoded [][]byte, tab
...
 			errGlob = err
```
iwanbk:
```
I know this is not a new change from this PR, but I noticed last week that we're simply overwriting the errGlob value in case of multiple values. Wouldn't it be a better idea to keep them all in a slice, and return that slice as an error, such that we have all errors?
```
Because what we do now with error is simply to Log it, and all errors already logged.
What we care now is simply knowing that it is failed/succeed.

Let's move to separate issue, i actually have created the issue for this in tlog repo https://github.com/g8os/tlog/issues/66

tlog/tlogserver/server/flusher.go
iwanbk: i think checking for error fist is a better practice. In fact,  your suggestion also missed the error checking :D.

```
so in there we have a private util function for that
```
feel free to add it to new tlog redis pool package if you think it is needed

tlog/tlogserver/server/flusher.go
```diff
@@ -266,3 +275,33 @@ func (f *flusher) encodeCapnp(vdiskID string, blocks []*schema.TlogBlock, lastHa
...
+
+func (f *flusher) storeLastHash(vdiskID string, lastHash []byte) error {
+	rc := f.redisPool.Get(0)
+	defer rc.Close()
+
+	_, err := rc.Do("SET", f.lastHashKey(vdiskID), lastHash)
+	return err
+}
+
+func (f *flusher) lastHashKey(vdiskID string) string {
```
iwanbk: Because i don't know what we need later.

nbdserver/ardb/backend.go
```diff
@@ -235,6 +235,22 @@ func (ab *Backend) GoBackground(ctx context.Context) {
 		return
 	}
 
+	type tlogResp struct {
+		resp *tlogclient.Response
+		err  error
+	}
+	respChan := make(chan tlogResp, 1)
+	go func() {
```
iwanbk: If we don't do this, then it is not async anymore :)

iwanbk: Oh sorry, i misread your question.
After i checked the `Recv` again, it seems not a good func

GlenDC: What? I don't understand, the `Recv` function is clearly also looping in a goroutine, so how is that different than what you are doing here?

tlog/tlogserver/server/flusher.go
GlenDC: Fair enough 👍

tlog/tlogserver/server/flusher.go
GlenDC: Yeah sorry my example is a bit wrong. But no it's ok, this was more of a nittypick comment, all good!

nbdserver/ardb/backend.go
GlenDC: What's wrong with the Recv function?

tlog/tlogserver/server/end_to_end_test.go
```diff
@@ -49,17 +50,35 @@ func TestEndToEnd(t *testing.T) {
...
-		tr, err := client.RecvOne()
-		assert.Nil(t, err)
-		assert.Equal(t, true, tr.Status >= 0)
-	}
+	wg.Add(2)
+
+	// send tlog
+	go func() {
+		defer wg.Done()
+		for i := 0; i < conf.FlushSize*numFlush; i++ {
```
iwanbk: done.
```
so if my assumptions are correct,
```
i actually assumed the other way.

I will check it for sure, i hope the compiler is as smart as i thought

tlog/tlogserver/server/server.go
iwanbk: feel free to submit a PR, as it is not part of this PR

tlog/tlogserver/server/server.go
iwanbk: Sorry, i don't understand your asssumption. Have you seen the full code of the func?

tlog/tlogserver/server/server.go
GlenDC: Fair enough!

tlog/tlogserver/server/server.go
iwanbk: In case that func error: that func will close the conn. The the parent func eventually exit.
and vice versa

nbdserver/ardb/backend.go
```diff
@@ -235,6 +235,22 @@ func (ab *Backend) GoBackground(ctx context.Context) {
 		return
 	}
 
+	type tlogResp struct {
```
iwanbk: oh, i missed this one.
Sure

tlog/tlogserver/server/server.go
GlenDC: Yes I have... OK so follow my my reading flow of this function:

1. start of function: `tlc == nil`
2. receive first request, `tlb.Vdisk() == "vdiskA"`, create/get tlc using `getTlogChan` and async `sendResp`;
3. receive second request, `tlb.Vdisk() == "vdiskB"`, tlc exists so correct channel isn't (created and) used...

See the issue that I'm seeing?

nbdserver/ardb/backend.go
iwanbk: `Recv` use two channels, one for the response, one for the error.
Using two channels make things complicated.

But maybe i should just simply improve the `Recv`

tlog/tlogserver/server/server.go
iwanbk: No.
```
 first request
second request
```
in case you missed it, it is persistent tcp connection.

1st & 2nd connection are going to server by different instance of `handle`, thus the are going to have different `tlc`.

`tlc` is local variable, not global

nbdserver/ardb/backend.go
GlenDC: Dunno, I quite like the 2 channel thing, means you have to send less data over the channels, and you have the for-loop-select already anyhow.

tlog/tlogserver/server/server.go
GlenDC: I think we are having a miss communication here.

I'm talking about the same connection. The Vdisk isn't bound to the tcp connection. Reading from the code, we are getting the vdiskID from the received value, each for loop iteration, meaning that there is no guarantee (as far that I know), that even within the same connection, we are sending always the same vdiskID.

Perhaps you never expect that a single connection returns different vdiskID, but it is certainly not validated here. And you know... always validate user input!

So either:

+ We need to make this code work with different vdisks on the same connection;
+ You need to validate that the vdiskID is always the same, and block the transaction in case it is meant for a different vdisk;

But the code as it right now is, would just silently accept this, and push the blocks to the wrong channel, as stated in my original comment...

nbdserver/ardb/backend.go
iwanbk: I think the better choice is to keep that `Recv` func.
And you can improve this part of code later.

Because i myself a bit confused to use 2 channels.

The main problem using two channels is: if you have two channels, the `select` doesn't guarantee that read order is the same as write order.
(this is the issue i got with decoder, it previously using two channels)

tlog/tlogserver/server/server.go
iwanbk:
```
 The Vdisk isn't bound to the tcp connection
```
Well, we need to fix many things if it is true. It is my assumption so far

with single connections for all vdisk, only one CPU core of tlog that will receive the packet. It is the case with dpdk, seastar, and Nim. And i think is the same for Go, at least in the in network stack level.

iwanbk:
```
always validate user input!
```
True!

nbdserver/ardb/backend.go
GlenDC: Ok that sounds fair enough, but then I would suggest to improve the Recv function, and use that instead. That seems like a better approach, compared to not using that Client function ever.

tlog/tlogserver/server/server.go
iwanbk:
```
You need to validate that the vdiskID is always the same, and block the transaction in case it is meant for a different vdisk;
```
Better choice for me.

GlenDC: Also, if this assumption isn't documented yet, we should really document that so that people using the tlog client know this kind of shit.

In which case, why not simply pass the vdiskID only at creation of the `tlog.Client`, and not require the user to give it every Send command, seems to be a bit useless, as that user will always give the same vdiskID anyhow to the same tlog.Client instance (or at least should do that, according to that assumption).

iwanbk: you're right

iwanbk:
```
Well, we need to fix many things if it is true. It is my assumption so far
```
Just realized that not so much to fix in Go tlog, only need to fix the part you addressed above and the receive channel

PR #139, opened 2017-05-12
add g8stor tool suite (replaces copyvdisk cli tool)

GlenDC:
Add initial g8stor tool suite, supporting following commands:

+ `g8stor version`: useful for when reporting bugs (especially when tool is built using the root Makefile);
+ `g8stor copy deduped ...`: copy metadata of deduped vdisk;
+ `g8stor copy nondeduped ...`: copy data (blocks) of nondeduped vdisk;
+ `g8stor delete deduped ...`: delete metadata of deduped vdisk;
+ `g8stor delete nondeduped ...`: delete data (blocks) of nondeduped vdisk;


nbdserver/ardb/nondeduped.go
```diff
@@ -122,13 +114,6 @@ func (ss *nonDedupedStorage) GoBackground(ctx context.Context) {
 	// no background thread needed
 }
 
-// get the unique key for a block,
-// based on its index and the shared vdiskID
-func (ss *nonDedupedStorage) getKey(blockIndex int64) string {
```
GlenDC: @robvanmieghem: nondeduped storage is now stored in a hashmap, rather than as individual keys of the form `vdiskid:index`. The reason being that it makes it much easier/performant to delete and copy nondeduped vdisks, while as far as I know, it doesn't have any negative performance impacts related to the nbdserver nondeduped vdisk operations.

I hope a breaking change like this is allowed, as I was hoping that none of the ardb content is persistent?! Please correct me if a breaking change like this is unpermitted :o

robvanmieghem: We don't have any production deployments for this yet so it's fine

Issue #131, opened 2017-05-10
tlog server : proper handle of failed flush


iwanbk:
moved from https://github.com/g8os/tlog/issues/66.

put back tlogs to memory and retry it?

- [x] retry 3 times, if failed to store to redis
- [x] last_hash of a vdisk must only be updated if all pieces successfully stored (or use another data structure, see : https://github.com/g8os/blockstor/issues/131#issuecomment-301045314) (#154)

GlenDC:
The question is of course how many retries do we want to allow? And are all failures equal, as in, should all types of failure trigger a retry, or are their fatal failures in which cases no retry would ever help.

Also, even though we keep them in memory, what if the service stops before those blocks in memory, were able to be sent to the tlogserver. Knowing this, should we also persist them on the client side (nbdserver) in some temporary cache file (or perhaps even better an ardb), or is this too much?

iwanbk:
```
Knowing this, should we also persist them on the client side (nbdserver) in some temporary cache file (or perhaps even better an ardb
```
We at least need to keep the block in nbdserver memory until that block received by the tlog server.

iwanbk:
updated first comment, write down the tasks list of this issue

iwanbk:
```
 last_hash of a vdisk must only be updated if all pieces successfully stored
```
In #144, flusher will rollback the last_hash of vdisk if pieces failed to be stored.
But if still has issue if the rollback itself failed.

It is a disaster because without knowing correct last hash, we couldn't restore the entire block chain.

As an alternative solution, instead of only store the last hash, we can store last X of last hash.
So, if the last one failed, we can use last - 1.

I think storing few last hash will be enough.

GlenDC:
> retry 3 times

This value looks pretty random and based on nothing. Also I stated before, I don't think it makes sense for all types of error to retry, as for some kind of failures you could retry all you want and it wouldn't make a difference.

TLDR: I think there should be a little more intelligence behind this, than simply retrying 3 times.

GlenDC:
> We at least need to keep the block in nbdserver memory until that block received by the tlog server.

Sure, but what if the nbdserver restarts/stops for whatever reason, while there are still some blocks in memory. Is it ok for these to be lost? Or do we want to store them in persistent memory or another type of storage?

iwanbk:
> This value looks pretty random and based on nothing. Also I stated before, I don't think it makes sense 
[...]

PR #144, opened 2017-05-12
tlog : retry flush if failed.

iwanbk:
Retry the flush if failed, part of #131.

### DIFF
--- a/tlog/tlogserver/server/flusher.go
+++ b/tlog/tlogserver/server/flusher.go
@@ -97,11 +97,13 @@ func (f *flusher) flush(blocks []*schema.TlogBlock, vd *vdisk) ([]uint64, error)
 
 	hash := blake2b.Sum256(encrypted)
 	lastHash := hash[:]
+
 	// store to ardb
-	if err := f.storeEncoded(vd.vdiskID, lastHash, erEncoded); err != nil {
+	if err := f.storeEncoded(vd, lastHash, erEncoded); err != nil {
 		return nil, err
 	}
-	vd.lastHash = lastHash
+
+	vd.lastHash = lastHash // only update last hash if everything is OK
 
 	seqs := make([]uint64, len(blocks))
 	for i := 0; i < len(blocks); i++ {
@@ -111,42 +113,59 @@ func (f *flusher) flush(blocks []*schema.TlogBlock, vd *vdisk) ([]uint64, error)
 	return seqs[:], nil
 }
 
-func (f *flusher) storeEncoded(vdiskID string, key []byte, encoded [][]byte) error {
+func (f *flusher) storeEncoded(vd *vdisk, key []byte, encoded [][]byte) error {
 	var wg sync.WaitGroup
 
 	length := f.k + f.m
 	wg.Add(length + 1)
 
-	var errGlob error
+	var allErr []error
+
 	// store encoded data
 	for i := 0; i < length; i++ {
 		go func(idx int) {
 			defer wg.Done()
 
 			blocks := encoded[idx]
-			rc := f.redisPool.Get(idx + 1)
-			defer rc.Close()
-
-			_, err := rc.Do("SET", key, blocks)
+			err := f.storeRedis(idx+1, key, blocks)
 			if err != nil {
-				log.Infof("error during flush idx %v: %v", idx, err)
-				errGlob = err
+				err = fmt.Errorf("error during flush idx %v: %v", idx, err)
+				allErr = append(allErr, err)
 			}
 		}(i)
 	}
 
 	// store last hash name
+	lastHashStored := false
 	go func() {
 		defer wg.Done()
-		err := f.storeLastHash(vdiskID, key)
+		err := f.storeLastHash(vd.vdiskID, key)
 		if err != nil {
-			log.Infof("error when setting last hash name: %v", err)
-			errGlob = err
+			err = fmt.Errorf("error when setting last hash name: %v", err)
+			allErr = append(allErr, err)
+		} else {
+			lastHashStored = true
 		}
 	}()
 
 	wg.Wait()
-	return errGlob
+
+	if lastHashStored && len(allErr) > 0 {
+		// if last hash successfully stored but we had error in data pieces,
+		// we need to roolback the last_hash
+		//
+		// FIXME : what to do if the rollback itself got error?
+		// - need to use other data structure?
+		// - only store last hash after we successfully store all pieces?
+		if err := f.storeLastHash(vd.vdiskID, vd.lastHash); err != nil {
+			log.Infof("WARNING: failed to rollback last hash of %v to %v, err: %v", vd.vdiskID, vd.lastHash, err)
+		}
+	}
+	if len(allErr) > 0 {
+		log.Infof("flush of vdiskID=%v failed: %v", vd.vdiskID, allErr)
+		return fmt.Errorf("flush vdiskID=%v failed", vd.vdiskID)
+	}
+	return nil
 }
 
 func (f *flusher) encodeCapnp(blocks []*schema.TlogBlock, vd *vdisk) ([]byte, error) {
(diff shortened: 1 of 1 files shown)


tlog/tlogserver/server/flusher.go
```diff
@@ -207,13 +226,40 @@ func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
...
+		rc := f.redisPool.Get(idx)
+		defer rc.Close()
+
+		_, err := rc.Do("SET", key, data)
+		return err
+	}
+
+	var err error
+
+	sleepMs := time.Duration(500) * time.Millisecond
```
GlenDC: should be made a constant (if we really want to go for this approach);

tlog/tlogserver/server/flusher.go
```diff
@@ -207,13 +226,40 @@ func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
...
+
+	for i := 0; i < 4; i++ {
```
GlenDC: As stated earlier in my comment on issue #131:

+ I think simply retrying N times, is flawed no matter how many times (as there is no intelligence behind it), and at best you will fix a few more cases, but at worst and more realistic you will make it all slower because of this;
+ We should indicate the categories of failures that are possible, and see for each of them how we should recover/resolve such failures automatically;

tlog/tlogserver/server/flusher.go
```diff
@@ -111,42 +113,59 @@ func (f *flusher) flush(blocks []*schema.TlogBlock, vd *vdisk) ([]uint64, error)
...
-				log.Infof("error during flush idx %v: %v", idx, err)
-				errGlob = err
+				err = fmt.Errorf("error during flush idx %v: %v", idx, err)
+				allErr = append(allErr, err)
 			}
 		}(i)
 	}
 
 	// store last hash name
+	lastHashStored := false
```
GlenDC: `var lastHashStored bool` is more golang-idiomatic I think.

tlog/tlogserver/server/flusher.go
```diff
@@ -111,42 +113,59 @@ func (f *flusher) flush(blocks []*schema.TlogBlock, vd *vdisk) ([]uint64, error)
...
 	go func() {
```
GlenDC: Doesn't this leave the storage in an unknown state as some hashes will be stored while others are not? Don't we want to clean those up and rollback more properly?

tlog/tlogserver/server/flusher.go
```diff
@@ -111,42 +113,59 @@ func (f *flusher) flush(blocks []*schema.TlogBlock, vd *vdisk) ([]uint64, error)
...
 
 	wg.Wait()
-	return errGlob
+
+	if lastHashStored && len(allErr) > 0 {
+		// if last hash successfully stored but we had error in data pieces,
+		// we need to roolback the last_hash
+		//
+		// FIXME : what to do if the rollback itself got error?
+		// - need to use other data structure?
```
GlenDC: What do you have in mind?

tlog/tlogserver/server/flusher.go
```diff
@@ -111,42 +113,59 @@ func (f *flusher) flush(blocks []*schema.TlogBlock, vd *vdisk) ([]uint64, error)
...
+		// - only store last hash after we successfully store all pieces?
```
GlenDC: Otherwise what's the point of storing last hash, when some hashes in between the previous last hash and the current last hash are missing, will make any attempt of replaying a tlog chain fail, no?

tlog/tlogserver/server/flusher.go
```diff
@@ -207,13 +226,40 @@ func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
...
+	for i := 0; i < 4; i++ {
+		err = store()
+		if err == nil {
+			return nil
+		}
+		log.Infof("error to store to idx:%v, err: %v", idx, err)
+
+		// sleep for a while
+		time.Sleep(sleepMs)
+		sleepMs *= 2 // double the sleep time for the next iteration
```
GlenDC: While it might seem like a good idea, I can really see this entire sleeping business make the tlog system even slower then it already is.

tlog/tlogserver/server/flusher.go
iwanbk: sorry, i don't get your question.

tlog/tlogserver/server/flusher.go
iwanbk:
```
 will make it all slower because of this;
```
i don't understand why it will make slower
```
We should indicate the categories of failures that are possible, and see for each of them how we should recover/resolve such failures automatically;
```
If you look at the code, it is only about SET to redis

tlog/tlogserver/server/flusher.go
iwanbk: i also don't understand why it will make it slower.

Sleeping/waiting is absolutely needed when we need for some condition to recover:
- ardb crashed
- OOM killer do some action

tlog/tlogserver/server/flusher.go
iwanbk:
```
 Don't we want to clean those up and rollback more properly
```
rollback (if needed, because we might use another data structure) is on another subtask, it is only  about retry flush

tlog/tlogserver/server/flusher.go
iwanbk: https://github.com/g8os/blockstor/issues/131#issuecomment-301045314

GlenDC: Sounds good!

tlog/tlogserver/server/flusher.go
GlenDC: You state in the fixme comment: `only store last hash after we successfully store all pieces?`

On which I respond that the fact that you only store the last hash _after_ you successfully store all pieces, should be a definitive yes.

tlog/tlogserver/server/flusher.go
GlenDC: > i don't understand why it will make slower

Because you have long and expensive sleep times, so in case of failure from which can't be recovered, the tlogserver will really slow down by a lot. Dunno maybe this doesn't matter, as a failure is bad anyhow, but not from all errors can be recovered you know, for example if it's about a broken TCP pipe, than you can retry as much as you want, but it will not work ever again, until you reestablish a new connection.

tlog/tlogserver/server/flusher.go
GlenDC: > ardb crashed

if the ardb crashed, I do think you need to recreate the TCP connection (I don't think the redis library does this for you, but I might be wrong), as the pipe will be broken. No?

tlog/tlogserver/server/flusher.go
iwanbk: i'm not sure if it will answer your question, we have two choices here:
- store all pieces and last hash in parallel -> faster, need rollback if piece(s) failed but last hash OK
- store all pieces then store last hash -> slower

tlog/tlogserver/server/flusher.go
iwanbk:
```
, for example if it's about a broken TCP pipe, than you can retry as much as you want, but it will not work ever again, until you reestablish a new connection.
```
when retrying, it get new conn from the pool.

tlog/tlogserver/server/flusher.go
GlenDC: Ah nevermind... Each retry for-loop-iteration we are getting another connection (and hopefully it recreates a connection and doesn't get a broken idle connection). So I guess simply retrying might work indeed. Sorry, forget what I said, hehe.

iwanbk: i get new connection from the pool. If the pool doesn't handle it, it should be fixed.

tlog/tlogserver/server/flusher.go
iwanbk:
```
Because you have long and expensive sleep times, so in case of failure from which can't be recovered, the tlogserver will really slow down by a lot
```
Please note that:
- it only slow that vdisk down, it won't affect other vdisk
- tlog need to store sequential history, if current aggregation is failed, there is no point to store next aggregation. we only need to do our best and then report to nbdserver that we are failed

tlog/tlogserver/server/flusher.go
GlenDC: Ok but this seems to get out of scope for this PR. Can you make a new issue for this or add the information to a relevant issue?

iwanbk: yep, i already make it a sub task of #131

PR #154, opened 2017-05-15
tlog : store last hash array instead of single last hash.

iwanbk:
Part of #131 

idea:

> 
> 
> Only storing one last hash will give us a problem if the last aggregation is failed to be stored to ardb but we already store the new last hash to ardb. We can of course try to roll it back, but the rollback itself might also failed.
> 
> To avoid the need to rollback, we can also only store the new last hash after all pieces successfully stored. But we are going to get performance hit.
> 
> So, storing X last hash is better choice here:
> 
> in normal condition, what we really need is the very last hash.
> if the very last value is not valid, we can use last -1
> if last -1 still not valid, we can use last-2...and so on.
> we need to keep X value to be small to avoid performance issue.
>

### DIFF
--- a/tlog/tlogclient/decoder/decoder.go
+++ b/tlog/tlogclient/decoder/decoder.go
@@ -3,6 +3,7 @@ package decoder
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 
 	"github.com/g8os/blockstor/log"
@@ -28,9 +29,6 @@ type Decoder struct {
 	decrypter tlog.AESDecrypter
 	privKey   []byte
 
-	// prefix of the last hash key
-	lastHashPrefix string
-
 	// pools of redis connection
 	redisPool *tlog.RedisPool
 }
@@ -45,7 +43,7 @@ type DecodedAggregation struct {
 // New creates a tlog decoder
 func New(objstorAddr []string, k, m int, vdiskID, privKey, nonce string) (*Decoder, error) {
 	if len(objstorAddr) != k+m+1 {
-		return nil, fmt.Errorf("invalid number of objstor")
+		return nil, errors.New("invalid number of objstor")
 	}
 
 	// create decrypter
@@ -61,14 +59,13 @@ func New(objstorAddr []string, k, m int, vdiskID, privKey, nonce string) (*Decod
 	}
 
 	return &Decoder{
-		objstorAddr:    objstorAddr,
-		vdiskID:        vdiskID,
-		k:              k,
-		m:              m,
-		decrypter:      decrypter,
-		privKey:        []byte(privKey),
-		lastHashPrefix: "last_hash_",
-		redisPool:      redisPool,
+		objstorAddr: objstorAddr,
+		vdiskID:     vdiskID,
+		k:           k,
+		m:           m,
+		decrypter:   decrypter,
+		privKey:     []byte(privKey),
+		redisPool:   redisPool,
 	}, nil
 }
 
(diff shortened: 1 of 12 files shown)


iwanbk:
It also updates the vendored redigo package as it is too old (from 2015) and doesn't have `redis.ByteSlices` func

tlog/tlogclient/decoder/decoder.go
```diff
@@ -227,16 +227,26 @@ func (d *Decoder) getAllPieces(key []byte) ([]byte, error) {
...
 
+// get last hash of a vdisk
 func (d *Decoder) getLastHash() ([]byte, error) {
 	rc := d.redisPool.Get(0)
 	defer rc.Close()
 
 	key := d.lastHashPrefix + d.vdiskID
-	b, err := redis.Bytes(rc.Do("get", key))
+
+	hashes, err := redis.ByteSlices(rc.Do("lrange", key, 0, -1))
```
GlenDC: be consistent with capitals in redis commands, either you use always full caps, or always lower caps, but mixing it triggers my OCD

tlog/tlogclient/decoder/decoder.go
```diff
@@ -227,16 +227,26 @@ func (d *Decoder) getAllPieces(key []byte) ([]byte, error) {
...
-	return b, nil
+
+	// check that the hash really valid
+	for _, hash := range hashes {
+		if _, err := rc.Do("get", hash); err == nil {
+			return hash, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no valid hash for vdiskID:%v", d.vdiskID)
```
GlenDC: you could just use `%s`, as you know it's a string

tlog/tlogserver/server/flusher.go
```diff
@@ -225,24 +226,37 @@ func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
...
-	return f.storeRedis(0, []byte(f.lastHashKey(vdiskID)), lastHash)
+	key := []byte(f.lastHashKey(vdiskID))
+
+	// we don't need to wait for the trim operation because it won't harm us
+	// if failed
+	go func() {
+		rc := f.redisPool.Get(0)
+		defer rc.Close()
+		if _, err := rc.Do("LTRIM", key, 0, lastHashNum); err != nil {
+			log.Errorf("failed to trim lasth hash of %v. err: %v", vdiskID, err)
```
GlenDC: -> `log.Errorf("failed to trim lasth hash of %s: %s", vdiskID, err.Error())`

tlog/tlogserver/server/flusher.go
```diff
@@ -225,24 +226,37 @@ func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
...
+	return f.storeRedis(0, "LPUSH", key, lastHash)
 }
 
 func (f *flusher) lastHashKey(vdiskID string) string {
 	return lastHashPrefix + vdiskID
 }
 
 // store data to redis and retry it if failed
-func (f *flusher) storeRedis(idx int, key, data []byte) error {
+func (f *flusher) storeRedis(idx int, command string, key, data []byte) error {
```
GlenDC: you could use the short `cmd`, it's a common enough abbreviation, thus acceptable

tlog/tlogserver/server/flusher.go
```diff
@@ -18,7 +18,8 @@ import (
 )
 
 const (
-	lastHashPrefix = "last_hash_"
+	lastHashPrefix = "last_hash_" // last hash ardb key prefix
```
GlenDC: this const should be shared between tlogclient and tlogserver, as they now both seem to define it

tlog/tlogserver/server/flusher.go
```diff
@@ -225,24 +226,37 @@ func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
...
 
 func (f *flusher) lastHashKey(vdiskID string) string {
 	return lastHashPrefix + vdiskID
 }
 
 // store data to redis and retry it if failed
-func (f *flusher) storeRedis(idx int, key, data []byte) error {
+func (f *flusher) storeRedis(idx int, command string, key, data []byte) error {
 
 	store := func() error {
```
GlenDC: should simply be a private method, as redeclaring it every time we use the `storeRedis` function is a inefficient

iwanbk: do you have docs to backup your statement?

GlenDC: No not really, and when looking back at it, I might have been wrong . What do you think @iwanbk?

iwanbk: IIRC i read or saw (video), said that closure is not defined every time the function called. It also make a bit optimization by calling the variable by reference/pointer instead of by value.

But unfortunately i can't find the docs anymore.

GlenDC: Please do look up that video, would love to learn about this.

In either case, if you're convinced that your approach was better, than please do change it back to that, and I'm sorry for the inconvenience I caused in that case.

PR #156, opened 2017-05-15
tlog client : recreate tcp connection if broken.

iwanbk:


tlog/tlogclient/client.go
```diff
@@ -81,33 +81,70 @@ func (c *Client) RecvOne() (*Response, error) {
 	}, nil
 }
 
-func createConn(addr string) (*net.TCPConn, error) {
-	tcpAddr, err := net.ResolveTCPAddr("tcp", addr)
+func (c *Client) createConn() error {
+	tcpAddr, err := net.ResolveTCPAddr("tcp", c.addr)
```
GlenDC: Question: Is there any benefit of first resolving the TCP address, and dialing TCP, rather then simply using the `net.Dial` func?

tlog/tlogclient/client.go
```diff
@@ -81,33 +81,70 @@ func (c *Client) RecvOne() (*Response, error) {
...
+			okToSend = true
+		}
 	}
-	return c.bw.Flush()
+	return err
 }
+
+const (
+	sendRetryNum = 3
+	sendSleepMs  = 500
```
GlenDC: should become `sendSleepMs = 500 * time.Millisecond`

tlog/tlogclient/client.go
```diff
@@ -81,33 +81,70 @@ func (c *Client) RecvOne() (*Response, error) {
...
 
 	conn, err := net.DialTCP("tcp", nil, tcpAddr)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
 	conn.SetKeepAlive(true)
-	return conn, nil
+	c.conn = conn
```
GlenDC: Doesn't this make the client thread-unsafe, as you're swapping out pointers here, which seems it might make things unstable if a single client is used in different routines at once. Unless of course we document that the client should only ever exist in 1 thread.

GlenDC:
Rebased master into this PR.

tlog/tlogclient/client.go
iwanbk: no, unless that net.DialTCP return *net.TCPConn explicitly

tlog/tlogclient/client.go
iwanbk: i added comment to `New`, saying that it is not goroutine safe.
Actually not only `createConn`, `Send` also not goroutine safe

tlog/tlogclient/client.go
iwanbk: Nice catch, fixed.

tlog/tlogclient/client.go
GlenDC: Fair enough, as long as it's clear from the documentation, that's fine for now. And for now it is not required to be multithreaded anyhow (I think).

PR #159, opened 2017-05-16
remove vdiskID from tlogblock

### LINKED ISSUE #141: tlog : remove vdisk ID from TlogBlock's field
We already have vdiskID in aggregation's field.
And all blocks in aggregation has same vdisk ID

GlenDC:
fixes #141


iwanbk:
Apologize to not share my idea about how it should be implemented.
I think we should do this:
- in initial connection, client send his vdiskID
- in `Send`, client do not need to send his vdiskID because it is a persistent connection. So we can avoid to send the same thing over and over.

GlenDC:
@iwanbk could you take over this PR then? As I'm about to stop working. Or if it's not too urgent, then I don't mind finishing this of with your idea, tomorrow.

iwanbk:
> Or if it's not too urgent, then I don't mind finishing this of with your idea

It is not too urgent.
Because we have variable-length vdiskID, we might need to add info that contains the length of the vdiskID. 

You can see https://github.com/g8os/blockstor/blob/52823a325b1ba4440703b92280b0b898b2ec91d8/tlog/capnp_prefix.go#L15-L43 for the example.

Another alternative is create new capnp struct and use same way as sending tlogBlock

GlenDC:
> Another alternative is create new capnp struct and use same way as sending tlogBlock

Yes, I think I would prefer to keep it as a capnp structure, as I think it's not good to start mixing protocols on the same connection. So I will do that.

iwanbk:
```
Yes, I think I would prefer to keep it as a capnp structure, as I think it's not good to start mixing protocols on the same connection. So I will do that.
```
Agree, we can also add more info in this initial message, like the expected sequence number.

tlog/tlogclient/client.go
```diff
@@ -36,12 +40,55 @@ func New(addr string, vdiskID string) (*Client, error) {
...
+	err = tlog.VerAckStatus(resp.Status()).Error()
+	if err != nil {
+		return err
+	}
+
+	// validate version
+	serverVersion := blockstor.Version(resp.Version())
+	if serverVersion.Compare(tlog.MinSupportedVersion) < 0 {
+		return errors.New("server version is not supported by this client")
+	}
```
iwanbk: I think it is enough for the server to check the compatibility.
Server knows better which client versions it supports.

Issue #128, opened 2017-05-08
tlog - nbd : multiple tlog client for each vdisk


iwanbk:
We currently only use one tlog client per vdisk.

Need to check if it is a good idea to make it multiple clients.

the server also need to be adapted to support this.

GlenDC:
Why would we need multiple clients per vdisk? Can't we use a single client for sending blocks in parallel?

iwanbk:
```
Can't we use a single client for sending blocks in parallel?
```
No, if we only have one TCP connection.

iwanbk:
I updated the issue's title and description, our main point is using multiple connections.

GlenDC:
Ah yeah, in that case, this does indeed sound like a good proposition 👍

iwanbk:
my tests using bigger image
- without tlog : 5.8 s

tlog doesn't store the messages -> no flush:
- current code : 11 s
- increase nbd `transactionChCapacity` to 100 (high watermark of tlog flusher) : 10 s
- use 2 tlog clients : 9 s
- use 2 tlog clients, set `transactionChCapacity` to 100: 9 s 

tlog store the messages
- current code : 18 s
- use 2 tlog clients, set `transactionChCapacity` to 100: 18 s -> yes, no improvement

iwanbk:
another interesting test data

with flush, but no capnp encoding. appended all block's data into one big `[]byte`
- current code : 12 seconds
- 2 tlog clients : 11 seconds

iwanbk:
but please note that when i said `2 tlog clients`, the tlog server hasn't been modified to handle the needed changes (i.e. : handle unordered data). This handling itself can slow down the server a bit.

iwanbk:
in latest develop:
- with single client : 12.7 s
- with 2 clients : 11.1 s

i'm still hesitant to use 2 clients now because:
- it added complexity to server (handling unordered data), it will slow the server down by some factor
- not sure about complexity in the client

iwanbk:
i'm just thinking to use two clients:
- one to build the capnp 
- one to send the data to server

lets see how it goes.

iwanbk:
```
i'm still hesitant to use 2 multiple clients now because:
```
but i guess it is something that unavoidable.

iwanbk:
```
i'm just thinking to  add another worker to build the capnp object
```
it only gave small (0.1 s from 12.4 s) improvement.

[...]

PR #168, opened 2017-05-17
support unordered tlog message from the start

### LINKED ISSUE #166: tlog - nbd : tlog might expect wrong first sequence
After #161 merged, tlog might expect wrong first sequence if these cases happen:
- first message **received succesfully**  is not the first message **sent** by nbdserver.
- nbdserver restarted and reset the sequence to zero.

We can add the valid sequence as part of initial connection message (see #159)

GlenDC:
fixes #166

What changed:

+ TlogClient now sends `firstSequence` as part of `ClientVerAck` object. This `firstSequence` gets used as the first sequence of the tlog sequence for that vdisk within the tlog server. This allows us to support unordered tlog messages right from the start;
+ TlogServer now only accepts 1 connection per active vdisk. Thus when the server receives a VerAck, it tries to create a vdisk (using the vdiskmanager) with the vdiskID given by the client. If that vdisk already exists, the connection will not be established and the returned status will be an error status. If it doesn't exist yet, it will be created, and the connection gets successfully established. When the connection drops, the vdisk will be deleted again from the vdiskManager;


iwanbk:
> TlogServer now only accepts 1 connection per active vdisk

How about the plan to support multiple connections?

tlog/tlogserver/server/vdisk.go
```diff
@@ -75,25 +76,34 @@ func newVdiskManager(blockSize, flushSize int) *vdiskManager {
...
+	vt.vdisks[vdiskID] = vd
 	return vd, nil
 }
 
+// delete a vdisk (if possible)
+func (vt *vdiskManager) delete(vdiskID string) {
+	vt.lock.Lock()
+	defer vt.lock.Unlock()
+
+	delete(vt.vdisks, vdiskID)
```
iwanbk: we also need to stop the goroutine if we want to delete it.(which i think is not necessary to fix this issue)

GlenDC:
> How about the plan to support multiple connections?

.... That's not possible with this PR in place...

But @iwanbk, I did specify this new behavior in our telegram chat. 

If we do want multiple connections, I'll need to adapt this PR.

Let me know, but this will be for tomorrow, as I don't have much more time left to work.

tlog/tlogserver/server/vdisk.go
GlenDC: Already resolved, refresh.

GlenDC:
If we do want to support multiple connections, than we do want to allow multiple connections on a single vdisk... however! In that case we do want to perhaps add some kind of reference counting within the vdisk manager, so that it can clean up automatically all goroutines once all connections for a vdisk are cleaned up. Because the old code simply leaves all goroutines open, which is not desired for long-running servers.

iwanbk:
> But @iwanbk, I did specify this new behavior in our telegram chat.

Well, only behavior to reject newly connected client 

> If we do want multiple connections, I'll need to adapt this PR.

Yep, we need to close #128 for this PR to be accepted. Let me know if #128 is a bad idea that we need to close it

> If we do want to support multiple connections, than we do want to allow multiple connections on a single vdisk... 

It already does. vdisk only accept new tlog message via a single `inputChan` channel.

> however! In that case we do want to perhaps add some kind of reference counting within the vdisk manager, so that it can clean up automatically all goroutines once all connections for a vdisk are cleaned up. Because the old code simply leaves all goroutines open, which is not desired for long-running servers..


While you're right about stopping the vdisk object, it is not needed for this issue.

GlenDC:
> While you're right about stopping the vdisk object, it is not needed for this issue.

Sure, but it also cannot wait for `1.1.0-alphe-4` I think...

>  Let me know if #128 is a bad idea that we need to close it

No, it's fine. Multiple connections does seem like a good enough idea, and your benchmarks do look good enough to support this more complex behavior.

iwanbk:
> Sure, but it also cannot wait for 1.1.0-alpha-4 I think...

Yep, just saying :). 

It is also better to confirm the goal of the milestone, because AFAIK the tlogserver in this milestone only supposed to handle a single VM.

Cleaning goroutines also not as simple as stopping it, we also need to flush the unflushed blocks. And of course handle the case where we have missing blocks sequence.

GlenDC:
> Cleaning goroutines also not as simple as stopping it, we also need to flush the unflushed blocks. And of course handle the case where we have missing blocks sequence.

That makes sense, can you add all this information to https://github.com/g8os/blockstor/issues/167 @iwanbk? Seems like we can fix that together with improving the general design of it all.

Will simply change the PR of this code then to not check for already connecting vdisks.

GlenDC:
@iwanbk:

+ I rebased master into this PR;
+ Modified this PR's commit to allow multiple connections per vdisk, thus not validating if a vdisk doesn't exist yet;

In other words. I fixed your only objection, thus I think this is ready to be merged.

Please review and merge.

iwanbk:
thanks @GlenDC. As a side note, can we rename `ClientVerAck` and 'ServerVerAck` to something else that more aligned with the content? Maybe handshakeReq and handshakeResp

I didn't ask for change in this PR because i already accept this naming in previous PR.

I already confused when it also contain vdiskID, it become even more confusing with `firstSequence` field addition.

GlenDC:
> As a side note, can we rename ClientVerAck and 'ServerVerAck` to something else that more aligned with the content? Maybe handshakeReq and handshakeResp

Ok! I'll think about it and open a PR aimed at you.

PR #179, opened 2017-05-18
suport yaml config in tlog

### LINKED ISSUE #170: tlogserver: get the ardb connectionstrings from a config.yml file
Similar to the nbdserver, don't remove the commandline possibility though. If the storage-addresses parameter is present, it overrules the configfile.

GlenDC:
+ fixes #170;
+ tlog.RedisPool is now an interface;
+ tlog.RedisPool API has changed, splitting Get(index) into a metadata and data method;
+ A tlog.RedisPool can now be created using MemoryRedis, Blockstor Config and simply giving the addresses (as was already supported);
+ flusher is now created during handshake phase;
+ tlog.RedisPool is now created during handshake phase (before flusher is created), using the new tlog.RedisPoolFactory interface;
+ all code interacting with tlog and tlog testcode has been modified to adapt to this change;

### DIFF
--- a/tlog/tlogclient/examples/walk_history/main.go
+++ b/tlog/tlogclient/examples/walk_history/main.go
@@ -2,21 +2,22 @@ package main
 
 import (
 	"flag"
-	"fmt"
 
+	blockstorcfg "github.com/g8os/blockstor/config"
 	"github.com/g8os/blockstor/log"
+	"github.com/g8os/blockstor/tlog"
 
 	"github.com/g8os/blockstor/tlog/tlogclient/decoder"
 )
 
 type config struct {
-	K                int
-	M                int
-	firstObjStorPort int
-	firstObjStorAddr string
-	privKey          string
-	vdiskID          string
-	nonce            string // encryption nonce
+	K                    int
+	M                    int
+	privKey              string
+	vdiskID              string
+	nonce                string // encryption nonce
+	TlogObjStorAddresses string
+	ConfigPath           string
 }
 
 func main() {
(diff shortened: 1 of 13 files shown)


GlenDC:
Rebased master into this PR's branch. It's ready for review @iwanbk.

tlog/tlogclient/examples/walk_history/main.go
```diff
@@ -73,6 +87,36 @@ func main() {
 	}
 }
 
+func parseConnectionStrings(storageStrings string) (configs []blockstorcfg.StorageServerConfig, err error) {
```
iwanbk: can't we make it a public function that can be shared? 
I see this func defined 3 times in this PR

iwanbk:
Thanks @GlenDC, only one comment in the PR.

Don't you think we need to update the docs/readme as well?

tlog/tlogclient/examples/walk_history/main.go
GlenDC: Sure, do you know any good place where we could put the function?

GlenDC:
> Don't you think we need to update the docs/readme as well?

Which docs/readmes are you talking about?

tlog/tlogclient/examples/walk_history/main.go
iwanbk: no

iwanbk:
> Which docs/readmes are you talking about?

https://github.com/g8os/blockstor/blob/master/tlog/tlogserver/README.md

Just realized that it is still very minimal docs. It is up to you then to update it or not.

GlenDC:
Going to make a separate issue for it, which would be part of milestone alpha-3, as better documentation is probably required, so that the Cairo knows how to use everything.

tlog/tlogclient/examples/walk_history/main.go
GlenDC: Going to add it in the root blockstor package in an `ardb.go` file, where we can put general utility functions like this. Will also provide some quick unit tests for it.

GlenDC: Will merge it all once I did that, is that ok with you @iwanbk?

iwanbk: sure

Issue #162, opened 2017-05-16
re-send the tlog if doesn't get response or response says error


iwanbk:
Part of #126

iwanbk:
implementation design:
- keep the transaction log in buffer/cache until it received by tlog server
- capnp encoding wont be retried except that it is caused by system (e.g. memory)  error
- if timeout passed, we re send the tlog message.
- the number of retry we will do should be configurable

iwanbk:
A work in progress can be found on `tlogclient-retry-timeout` branch.

I added a container to manage the block that already sent but not get OK response yet.
I'm happy with the API but i think it still need to be improved:
- unit test (still no idea how to simulate this case)
- it still use Go map, which is work but might be slow for this case. i think i can repace it with https://godoc.org/github.com/emirpasic/gods/maps/treemap to improve the performance

iwanbk:
> it still use Go map, which is work but might be slow for this case. i think i can repace it with https://godoc.org/github.com/emirpasic/gods/maps/treemap to improve the performance

treemap is not good choice, because it order the data by it's key.
What we need is something that other the data by it's content and we can still access it by key.

If we can't find it, we can use two data structure to avoid iterating over unordered map.
But using two data structures itself add some kind of complexity to the code and might create maintenance issue.

Fortunately, we don't need it now because from my benchmark, it doesn't give perfomance reduction to the client.

iwanbk:
> unit test (still no idea how to simulate this case)

Implementation at PR #181 still doesn't have unit test, i only tested it manually there.

We might need fake/dummy tlog server to properly test it automatically.

GlenDC:
I think it should be pretty do-able to write a test for this. Could you describe to me:

+ what are the scenarios that you want to test;
+ what are you exactly doing to manually test it for now;

Knowing this I might be able to help you with this.

iwanbk:
Considered this issue and the test as low priority, was planning to do it after i feel safe we can finish other issue in this milestone ( which i don't think so).
I'm going to check all other issues after finishing #183

> what are you exactly doing to manually test it for now;

modified the 'Recv':
- for some sequences, i ignored the response (it simulates timeout)
- for some sequences, i force it to enter `if tlog.BlockStatusRecvFailed` condition

what is your plan to simulate the server btw?

GlenDC:
Not sure, all depends on what we want to tests etc.

iwanbk:
[...]

PR #181, opened 2017-05-19
Tlogclient retry on timeout/error

iwanbk:
Implement part of #162

### DIFF
--- a/tlog/tlogclient/client.go
+++ b/tlog/tlogclient/client.go
@@ -117,25 +117,27 @@ func (c *Client) Recv(chanSize int) <-chan *Result {
 	go func() {
 		for {
 			tr, err := c.RecvOne()
-			status := tlog.BlockStatus(tr.Status)
-			seq := tr.Sequences[0]
+			if tr != nil {
+				status := tlog.BlockStatus(tr.Status)
+				seq := tr.Sequences[0]
 
-			// if it failed to be received, promote it to be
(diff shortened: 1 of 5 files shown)


tlog/tlogclient/client.go
```diff
@@ -90,6 +127,29 @@ func (c *Client) Recv(chanSize int) <-chan *Result {
 	go func() {
 		for {
 			tr, err := c.RecvOne()
+			if tr != nil {
+				status := tlog.BlockStatus(tr.Status)
+				seq := tr.Sequences[0]
+
+				// if it failed to be received, promote it to be
+				// timed out as soon as possible.
+				if status == tlog.BlockStatusRecvFailed {
```
GlenDC: `tlog.BlockStatusRecvFailed` never gets sent by the server? Also you don't seem to check the other possible errors (`tlog.BlockStatusFlushFailed`) here?

iwanbk: > tlog.BlockStatusRecvFailed never gets sent by the server? 


Yes, for now, was planning to sent it in case of server failed to decode or check the hash.
But in the end i kept the current code, to close the connection because it indicate a bad network connection or faulty client.
Maybe i should delete this code

> Also you don't seem to check the other possible errors (tlog.BlockStatusFlushFailed) here?

Because we don't need to

PR #182, opened 2017-05-19
Fix incorect get last hash in tlog server during vdisk startup.

### LINKED ISSUE #176: tlog : vdisk failed to be restarted
Must be the culprit of #174 for this error

> couldn't get or create vdisk 1234567890, err: WRONGTYPE Operation against a key holding the wrong kind of value module=global caller=server.go:170

But not for the handshake error.

I believe it caused by regression of #154.

This https://github.com/g8os/blockstor/blob/c56caccf3e5ce418a0f807dee61537e55785fed9/tlog/tlogserver/server/flusher.go#L208 should be moved to `vdisk` instead of `flusher` (vdisk was not exist in initial code) and use same func used by decoder.

iwanbk:
Fixes #176

### DIFF
--- a/tlog/tlogclient/decoder/last_hash.go
+++ b/tlog/tlogclient/decoder/last_hash.go
@@ -1,0 +1,43 @@
+package decoder
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/garyburd/redigo/redis"
+
+	"github.com/g8os/blockstor/tlog"
+)
+
+var (
+	// ErrNilLastHash indicates that there is no last hash entry
+	// in the metadata storage.
+	ErrNilLastHash = errors.New("nil last hash")
+)
+
+// GetLastHash returns valid last hash of a vdisk.
+func GetLastHash(rc redis.Conn, vdiskID string) ([]byte, error) {
+	key := tlog.LastHashPrefix + vdiskID
+
+	hashes, err := redis.ByteSlices(rc.Do("LRANGE", key, 0, -1))
+	if err == redis.ErrNil {
+		return nil, ErrNilLastHash
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(hashes) == 0 {
+		return nil, ErrNilLastHash
+	}
+
+	// check that the hash really valid
+	for _, hash := range hashes {
+		if _, err := rc.Do("GET", hash); err == nil {
+			return hash, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no valid hash for vdiskID: %s", vdiskID)
+}
--- a/tlog/tlogclient/decoder/decoder.go
+++ b/tlog/tlogclient/decoder/decoder.go
@@ -222,21 +222,7 @@ func (d *Decoder) getLastHash() ([]byte, error) {
 	rc := d.pool.MetadataConnection()
 	defer rc.Close()
 
-	key := tlog.LastHashPrefix + d.vdiskID
-
-	hashes, err := redis.ByteSlices(rc.Do("LRANGE", key, 0, -1))
-	if err != nil {
-		return nil, fmt.Errorf("failed to get key:%v, err:%v", key, err)
-	}
-
-	// check that the hash really valid
-	for _, hash := range hashes {
-		if _, err := rc.Do("GET", hash); err == nil {
-			return hash, nil
-		}
-	}
-
-	return nil, fmt.Errorf("no valid hash for vdiskID: %s", d.vdiskID)
+	return GetLastHash(rc, d.vdiskID)
 }
 
 func (d *Decoder) decodeCapnp(data []byte) (*schema.TlogAggregation, error) {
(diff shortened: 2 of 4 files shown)


tlog/tlogclient/decoder/last_hash.go
```diff
@@ -0,0 +1,43 @@
...
+
+var (
+	// ErrNilLastHash indicates that there is no last hash entry
+	// in the metadata storage.
+	ErrNilLastHash = errors.New("nil last hash")
+)
+
+// GetLastHash returns valid last hash of a vdisk.
+func GetLastHash(rc redis.Conn, vdiskID string) ([]byte, error) {
+	key := tlog.LastHashPrefix + vdiskID
```
GlenDC: Perhaps move the

```
func (f *flusher) lastHashKey(vdiskID string) string {
	return tlog.LastHashPrefix + vdiskID
}
```
method to a public method in this file, so you can also use that same logic here, makes it easier to change the exact behavior later on

tlog/tlogclient/decoder/last_hash.go
```diff
@@ -0,0 +1,43 @@
...
+	hashes, err := redis.ByteSlices(rc.Do("LRANGE", key, 0, -1))
+	if err == redis.ErrNil {
+		return nil, ErrNilLastHash
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(hashes) == 0 {
```
GlenDC: I don't think this can ever occur, I'm fairly sure that `redis.ByteSlices` returns `redis.ErrNil` in this case, which is something you already check.

tlog/tlogclient/decoder/last_hash.go
```diff
@@ -0,0 +1,43 @@
...
+		return nil, ErrNilLastHash
+	}
+
+	// check that the hash really valid
+	for _, hash := range hashes {
```
GlenDC: Shouldn't we also perhaps delete (asynchronously) the invalid hashes from the metadata, as to not keep these invalid hashes in memory, or is there a good reason why you keep them in memory?

tlog/tlogclient/decoder/last_hash.go
iwanbk: agree

tlog/tlogclient/decoder/last_hash.go
iwanbk: I got this case :), maybe will check it again.

tlog/tlogclient/decoder/last_hash.go
iwanbk: > we also perhaps delete (asynchronously

no need to be async, we only need it at:
- vdisk creation
- once at tlog replay

> as to not keep these invalid hashes in memory

it is in ardb, not in memory, in case you missed it

> is there a good reason why you keep them in memory?

There is no good reason, i just didn't think that we need to do it.
But perhaps deleting it from ardb is better

tlog/tlogclient/decoder/last_hash.go
GlenDC: @iwanbk you are right. I just checked through the code of `redis.BytesSlice` and it is indeed possible, as they are simply doing a type check on `[]interface{}`, which is also possible to match for an empty slice.

In that case I would change the code to:

```go
if err == redis.ErrNil || len(hashes) == 0 {
    return nil, ErrNilLastHash
}

if err != nil {
    return nil, err
}
```

Or perhaps you have a good reason to not group it together. Either way I don't think it _really_ matters.

tlog/tlogclient/decoder/last_hash.go
GlenDC: Ah yeah sorry for the confusion, I know it is in the ardb, but in the end redis is also by default just in memory, which is why I was talking about "in memory", but I guess that's not completely correct, so yeah I'll try to talk about ardb next time.

I don't think you _need_ to do it. But if those invalid hashes will stay invalid forever, it might increase performance perhaps a bit if you remove them as soon as you detect it? Or what do you think?

tlog/tlogclient/decoder/last_hash.go
iwanbk: i think that won't work because `if err != nil ` is basically the same as `if len(hashes)==0`

tlog/tlogclient/decoder/last_hash.go
iwanbk: we currently do LTRIM when we LPUSH/RPUSH it, so the invalid hashes will be eventually gone

iwanbk: i might refactor it at #183,  perhaps put this last_hast things in separate package, although not sure how to do it because we will set last hash in the same time we set the erasure encoded data pieces

tlog/tlogclient/decoder/last_hash.go
GlenDC: That's not true, because you check now after `if err != nil { return nil, err }`, the `len(hashes) == 0` case, as we can have a zero-slice without it being an error apparently. So I don't understand why this wouldn't work.

iwanbk: with this code
```
if err == redis.ErrNil || len(hashes) == 0 {
```
what kind of error that is meet those condition?
Because all error will give us `len(hashes)==0`

GlenDC: OOoooh. I understand you. Ok yeah, that's correct, then the code is good as it is! 👍

PR #184, opened 2017-05-19
tlog : make RecvOne to private func and improve the example.

### LINKED ISSUE #165: Remove RecvOne method from the TlogClient public API
As it is hard to use it properly anyway, and it's good for the api to have its surface as small as possible.

Originally mentioned as a comment in #126.

iwanbk:
fix #165

### DIFF
--- a/tlog/tlogclient/examples/send_tlog/client.go
+++ b/tlog/tlogclient/examples/send_tlog/client.go
@@ -2,6 +2,8 @@ package main
 
 import (
 	"flag"
+	"sync"
+	"time"
 
 	"github.com/g8os/blockstor/log"
 	"github.com/g8os/blockstor/tlog/schema"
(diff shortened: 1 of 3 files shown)


tlog/tlogclient/examples/send_tlog/client.go
```diff
@@ -22,47 +24,53 @@ func main() {
...
-			log.Infof("j=%v\n", j)
-			err := client.Send(schema.OpWrite, j, j, j, data, uint64(len(data)))
-			if err != nil {
-				log.Info("client died\n")
-				return
-			}
-			tr, err := client.RecvOne()
-			if err != nil {
-				log.Fatalf("client failed to recv:%v\n", err)
+		respChan := client.Recv(1)
```
GlenDC: better to make this a local constant as in:

```go
const chanSize = 1
respChan := client.Recv(chanSize)
```

makes the example code more clear, as the first-time reader of this code, might not know what that 1 represents without looking it up, this small change makes it immediately clear.

iwanbk:
```
const chanSize = 1
```
Although you are 100% right about immutability things, i'm not sure if it is an idiomatic code.

GlenDC: In this case the const is just logical, as code should be immutable by default. But more important in this case is that it is example code, and thus I think it should be very clear what parameters are. And if you see `client.Recv(1)`, then I really wouldn't know what this `1` is without having to look it up. So in this case it's more about making the example code very clear, and naming the parameter, so it is very very clear to the reader of the example code that this `1` is the chanSize.

iwanbk: well, i think if we need to improve something here, then is the docs of `client.Recv`.

But i will improve it anyway :) .

And to be honest, i don't understand why you only need that part.
This line
```
client.Send(schema.OpWrite, seq, seq*dataLen, uint64(time.Now().Unix()), data, uint64(len(data)))
```
is more cryptic for me :D

GlenDC: That is very cryptic indeed.

PR #185, opened 2017-05-19
tlog client : fix reconnect in regard to handshaking

### LINKED ISSUE #174: tlog: handshake failed: client version is not supported by this server
Log in tlog server with nbdserver as client
```
handshake failed: client version is not supported by this server module=global caller=server.go:163
```

iwanbk:
Fixes #174

- also send handshake when reconnecting. This why in #174 tlog server complained about unsupported version. Because the client indeed didn't send handshake message at all
- Fix race condition between handshake and receiver thread. 'Recv' is reading from socket continuously, while the handshaking also need to read from it.

`send_tlog` example also improved to wait for tlogs to be flushed before exiting.

### DIFF
--- a/tlog/tlogclient/blockbuffer/buffer.go
+++ b/tlog/tlogclient/blockbuffer/buffer.go
@@ -3,6 +3,7 @@ package blockbuffer
 import (
 	"errors"
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
@@ -16,11 +17,12 @@ var (
 // Buffer defines buffer of tlog blocks that already sent
 // but still waiting to be succesfully received by the server
 type Buffer struct {
-	lock      sync.Mutex
-	entries   map[uint64]*entry
-	readyChan chan *schema.TlogBlock
-	timeout   time.Duration
-	maxRetry  int
+	lock            sync.Mutex
+	entries         map[uint64]*entry
+	readyChan       chan *schema.TlogBlock
+	timeout         time.Duration
+	maxRetry        int
+	highestSequence uint64 // the highest sequence we received
 }
 
 // NewBuffer creates a new tlog blocks buffer
@@ -65,6 +67,12 @@ func (b *Buffer) Add(block *schema.TlogBlock) {
 		ent.update(b.timeout)
 		return
 	}
+
+	seq := block.Sequence()
+	if seq > b.highestSequence {
+		b.highestSequence = seq
+	}
+
 	ent = newEntry(block, b.timeout)
 
 	b.entries[key] = ent
@@ -107,4 +115,24 @@ func (b *Buffer) TimedOut() <-chan *schema.TlogBlock {
 	}()
 
 	return b.readyChan
+}
+
+// MinSequence returns min sequence that this
+// buffer will/currently has.
+func (b *Buffer) MinSequence() uint64 {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	if len(b.entries) == 0 {
+		return b.highestSequence + 1
+	}
+
+	var seq uint64 = math.MaxUint64
+	for _, ent := range b.entries {
+		blockSeq := ent.block.Sequence()
+		if blockSeq < seq {
+			seq = blockSeq
+		}
+	}
+	return seq
 }
--- a/tlog/tlogclient/client.go
+++ b/tlog/tlogclient/client.go
@@ -57,20 +57,28 @@ func New(addr, vdiskID string, firstSequence uint64) (*Client, error) {
 		ctx:         ctx,
 		cancelFunc:  cancelFunc,
 	}
-	err := client.createConn()
-	if err != nil {
-		return nil, fmt.Errorf("client couldn't be created: %s", err.Error())
+	if err := client.connect(firstSequence); err != nil {
+		return nil, err
 	}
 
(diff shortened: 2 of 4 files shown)


tlog/tlogclient/client.go
```diff
@@ -15,8 +15,11 @@ import (
 	"github.com/g8os/blockstor/tlog/tlogclient/blockbuffer"
 )
 
-var (
-	resendTimeoutDur = 5 * time.Second // duration to wait before re-send the tlog.
+const (
```
GlenDC: I wonder if we can make these constants a bit more dynamic, and define them based on the CPU power etc. As it would make these timings a bit more constant. Constant values mean that on a powerful machine these values are very high, while on a very crippled VM these values might be a lot more lower.

Not sure if it's important at this time though, I think it's definitely outside the scope of this PR, but throwing it out here, to hear your thoughts on this.

tlog/tlogclient/client.go
```diff
@@ -41,9 +44,16 @@ type Client struct {
 	bw              *bufio.Writer
 	blockBuffer     *blockbuffer.Buffer
 	capnpSegmentBuf []byte
-	lock            sync.RWMutex
-	ctx             context.Context
-	cancelFunc      context.CancelFunc
+
+	// write lock, to protect against parallel Send
+	// which is not goroutine safe yet
+	wLock sync.RWMutex
```
GlenDC: The way you are using these mutexes, I think you can simply use `sync.Mutex`.

tlog/tlogclient/examples/send_tlog/client.go
```diff
@@ -54,10 +55,13 @@ func main() {
 	go func() {
 		defer wg.Done()
 
-		respChan := client.Recv(1)
+		respChan := client.Recv(100)
```
GlenDC: Probably best to move this to a constant value, locally or globally defined, up to you.

tlog/tlogclient/examples/send_tlog/client.go
```diff
@@ -77,7 +81,8 @@ func main() {
 	}()
 
 	// send the data
-	for seq, _ := range logsToSend {
+	for i := 0; i < numLogsToSend; i++ {
+		seq := uint64(i)
```
GlenDC: good fix 🗡

tlog/tlogclient/blockbuffer/buffer.go
```diff
@@ -65,6 +67,12 @@ func (b *Buffer) Add(block *schema.TlogBlock) {
 		ent.update(b.timeout)
 		return
 	}
+
+	seq := block.Sequence()
```
iwanbk: my mistake

tlog/tlogclient/blockbuffer/buffer.go
```diff
@@ -108,3 +116,23 @@ func (b *Buffer) TimedOut() <-chan *schema.TlogBlock {
...
+// buffer will/currently has.
+func (b *Buffer) MinSequence() uint64 {
+	b.lock.Lock()
+	defer b.lock.Unlock()
+
+	if len(b.entries) == 0 {
+		return b.highestSequence + 1
+	}
+
+	var seq uint64 = math.MaxUint64
```
iwanbk:
```
Nitpick: it's idiomatic to use seq := math.MaxUint64
```
Are you sure that it is working?

tlog/tlogclient/blockbuffer/buffer.go
```diff
@@ -108,3 +116,23 @@ func (b *Buffer) TimedOut() <-chan *schema.TlogBlock {
...
+	for _, ent := range b.entries {
+		blockSeq := ent.block.Sequence()
```
iwanbk: yep

tlog/tlogclient/client.go
iwanbk: if we also want to consider CPU power, then it is CPU power of the server, not the client that really matter

tlog/tlogclient/client.go
iwanbk: true

iwanbk:
```
 but resolving these comments and questions won't change the logic of this PR, so I'm going to merge these changes already as they look useful!
```
Will improve the code based on your comment later, thanks for the review.
Of course it is useful, without this PR the reconnect won't working at all :)

tlog/tlogclient/blockbuffer/buffer.go
GlenDC: > Are you sure that it is working?

You are right, that doesn't work. Man, I don't know why they didn't typedef that const as a uint64, a bit weird, but OK. You would have to do `seq := uint64(math.MaxUint64)`, which is worse than `var seq uint64 = math.MaxUint64`, so just ignore my original comment, haha.

tlog/tlogclient/client.go
GlenDC: You are right. Although the CPU power of client also has some influence. But OK, yeah this might start to be a bit much (overcomplicated), I guess as long as we keep these values big enough, it should be OK.

iwanbk:
> Will improve the code based on your comment later, thanks for the review.

done at c3cae334e2ecd865a2a8baec18dbbc3432b38750

PR #189, opened 2017-05-22
Tlog metadata redundancy

### LINKED ISSUE #183: tlog : add redundancy to the metadata
We already have redundancy mechanism for the data pieces using erasure encoding.

But we don't have it for the metadata (it currently only last hash).

My idea right now is using redis stored procedure to replicate the metadata to all storages we have.
So, we don't need the dedicated storage for metadata now, we can put it in the data&encoding storage.

What do you think @GlenDC @maxux @nbabenko?

iwanbk:
Fixes #183:
- use redis pipelining to write data and update last_hash
- update tlog decoder to check&validate last hash from ALL storages.
- remove Metadata from tlog.RedisPool

### DIFF
--- a/tlog/redispool.go
+++ b/tlog/redispool.go
@@ -24,16 +24,14 @@ import (
 //   an inmemory redis pool will be created instead;
 func AnyRedisPoolFactory(cfg RedisPoolFactoryConfig) (RedisPoolFactory, error) {
 	if length := len(cfg.ServerConfigs); length > 0 {
-		requiredServers := cfg.RequiredDataServerCount
-
 		serverConfigs := cfg.ServerConfigs
-		if length < requiredServers {
+		if length < cfg.RequiredDataServerCount {
 			if !cfg.AutoFill {
 				return nil, errors.New("not enough storage server configs are given")
 			}
 
 			var err error
-			serverConfigs, err = autoFillStorageServerConfigs(requiredServers, serverConfigs)
+			serverConfigs, err = autoFillStorageServerConfigs(cfg.RequiredDataServerCount, serverConfigs)
 			if err != nil {
 				return nil, err
 			}
@@ -101,16 +99,14 @@ func ConfigRedisPoolFactory(requiredDataServerCount int, configPath string) (Red
 //   and allowInMemory is true, an inmemory redis pool will be created instead;
 func AnyRedisPool(cfg RedisPoolConfig) (RedisPool, error) {
 	if length := len(cfg.ServerConfigs); length > 0 {
-		requiredServers := cfg.RequiredDataServerCount
-
 		serverConfigs := cfg.ServerConfigs
-		if length < requiredServers {
+		if length < cfg.RequiredDataServerCount {
 			if !cfg.AutoFill {
 				return nil, errors.New("not enough storage server configs are given")
 			}
 
 			var err error
-			serverConfigs, err = autoFillStorageServerConfigs(requiredServers, serverConfigs)
+			serverConfigs, err = autoFillStorageServerConfigs(cfg.RequiredDataServerCount, serverConfigs)
 			if err != nil {
 				return nil, err
 			}
@@ -147,9 +143,7 @@ func NewRedisPool(requiredDataServerCount int, storageServers []config.StorageSe
 func InMemoryRedisPool(requiredDataServerCount int) RedisPool {
 	var serverConfigs []config.StorageServerConfig
 
-	requiredServers := requiredDataServerCount
-
-	for i := 0; i < requiredServers; i++ {
+	for i := 0; i < requiredDataServerCount; i++ {
 		memoryRedis := redisstub.NewMemoryRedis()
 		memoryAddress := memoryRedis.Address()
 
--- a/tlog/redispool.go
+++ b/tlog/redispool.go
@@ -24,8 +24,7 @@ import (
 //   an inmemory redis pool will be created instead;
 func AnyRedisPoolFactory(cfg RedisPoolFactoryConfig) (RedisPoolFactory, error) {
 	if length := len(cfg.ServerConfigs); length > 0 {
-		// one extra (as we require also a metaDataServer)
-		requiredServers := cfg.RequiredDataServerCount + 1
+		requiredServers := cfg.RequiredDataServerCount
 
 		serverConfigs := cfg.ServerConfigs
 		if length < requiredServers {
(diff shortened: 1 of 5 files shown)


tlog/tlogclient/decoder/last_hash.go
```diff
@@ -22,33 +21,86 @@ func GetLashHashKey(vdiskID string) []byte {
...
 
 // GetLastHash returns valid last hash of a vdisk.
-func GetLastHash(pool tlog.RedisPool, vdiskID string) ([]byte, error) {
-	return getLastHashFromShard(pool, 1, vdiskID)
+// It checks all data shards to get latest valid hash
+func (d *Decoder) GetLastHash() ([]byte, error) {
+	latests := map[uint64][]byte{}
+
+	// get all valid hashes from all data storages
+	for i := 0; i < d.k+d.m; i++ {
```
GlenDC: should probably cache this value

tlog/tlogclient/decoder/last_hash.go
```diff
@@ -22,33 +21,86 @@ func GetLashHashKey(vdiskID string) []byte {
...
+		}
+	}
+
+	// no hash found, return Nil
+	if len(latests) == 0 {
+		return nil, ErrNilLastHash
+	}
+
+	// get the latest valid hash
+	var maxSeq uint64
```
GlenDC: If we only care about the maxSeq, why store all of them in a map first? We could simply do this right from the beginning, no?

tlog/tlogserver/server/flusher.go
```diff
@@ -275,7 +279,12 @@ func (f *flusher) store(idx int, hash, lastHashKey, data []byte) errStore {
 }
 
 func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
-	hash, err := decoder.GetLastHash(f.pool, vdiskID)
+	dec, err := decoder.New(f.pool, f.k, f.m, vdiskID, f.privKey, f.hexNonce)
```
GlenDC: Why create this decoder every time you getLastHash, seems pretty expensive, can we perhaps keep it as a private member of the fluser, and create the decoder at creation time of the flusher?

tlog/tlogclient/decoder/last_hash.go
```diff
@@ -22,33 +21,86 @@ func GetLashHashKey(vdiskID string) []byte {
...
+
+	// check latest sequence
+	blocks, err := agg.Blocks()
+	if blocks.Len() == 0 {
+		return 0, fmt.Errorf("empty block")
+	}
+
+	for i := 0; i < blocks.Len(); i++ {
+		block := blocks.At(i)
+		if block.Sequence() > lastSeq {
```
GlenDC: `block.Sequence()` can be cached instead of calling it 2 times.

tlog/tlogclient/decoder/last_hash.go
```diff
@@ -22,33 +21,86 @@ func GetLashHashKey(vdiskID string) []byte {
...
+	agg, err := d.get(hash)
+	if err != nil {
+		return 0, err
+	}
+
+	var lastSeq uint64
+
+	// check latest sequence
+	blocks, err := agg.Blocks()
+	if blocks.Len() == 0 {
```
GlenDC: `blocks.Len()` can be cached as well.

tlog/redispool.go
```diff
@@ -104,8 +101,7 @@ func ConfigRedisPoolFactory(requiredDataServerCount int, configPath string) (Red
 //   and allowInMemory is true, an inmemory redis pool will be created instead;
 func AnyRedisPool(cfg RedisPoolConfig) (RedisPool, error) {
 	if length := len(cfg.ServerConfigs); length > 0 {
-		// one extra (as we require also a metaDataServer)
-		requiredServers := cfg.RequiredDataServerCount + 1
+		requiredServers := cfg.RequiredDataServerCount
```
GlenDC: I don't think there is a need to cache this value if we don't modify it.

tlog/tlogserver/server/flusher.go
```diff
@@ -195,101 +144,154 @@ func (f *flusher) encodeCapnp(blocks []*schema.TlogBlock, vd *vdisk) ([]byte, er
...
-func (f *flusher) storeMetadataAndRetry(cmd string, key, data []byte) error {
-	var err error
+	// send command to store the last hash
+	if err := rc.Send("LPUSH", lastHashKey, hash); err != nil {
+		es.errSend = err
+		return es
+	}
 
-	sleepMs := time.Duration(500) * time.Millisecond
+	// flush the command
```
GlenDC: s/command/commands

tlog/tlogserver/server/flusher.go
```diff
@@ -195,101 +144,154 @@ func (f *flusher) encodeCapnp(blocks []*schema.TlogBlock, vd *vdisk) ([]byte, er
...
 }
 
-// store metadata to redis and retry it if failed
-func (f *flusher) storeMetadata(cmd string, key, data []byte) error {
-	// get new connection inside this function.
-	// so in case of error, we got new connection that hopefully better
-	rc := f.pool.MetadataConnection()
-	defer rc.Close()
+func (f *flusher) getLastHash(vdiskID string) ([]byte, error) {
+	dec, err := decoder.New(f.pool, f.k, f.m, vdiskID, f.privKey, f.hexNonce)
```
GlenDC: Why create this decoder every time you getLastHash, seems pretty expensive, can we perhaps keep it as a private member of the fluser, and create the decoder at creation time of the flusher?

tlog/tlogclient/decoder/last_hash.go
iwanbk: i knew you'll comment on this :D.
it is not in hot path, so keep it there make the code shorter and cleaner for me.

tlog/tlogclient/decoder/last_hash.go
iwanbk: my bad, fixed

tlog/tlogclient/decoder/last_hash.go
iwanbk: the reason is the same as previous comment about cached value.

tlog/tlogclient/decoder/last_hash.go
iwanbk: the reason is the same as previous comment about cached value. And more, it only called twice once.
but i have improved the code, so this part is entirely gone :)

tlog/tlogserver/server/flusher.go
iwanbk: fixed

tlog/redispool.go
iwanbk: ah, it was quick search and modify, fixed

tlog/tlogserver/server/flusher.go
iwanbk: no need for it, we only call it once, at vdisk creation

GlenDC: Then why store `privKey` and `hexNonce` as members of flusher?

tlog/tlogserver/server/flusher.go
iwanbk: no need for it, we only call it once, at vdisk creation

tlog/tlogserver/server/flusher.go
iwanbk: short reason : it is not called in flusher creation
long reason : with current code, from quick my view,  i can't find a quick way to to call it on flusher creation. (oh it is still short reason :D)

PR #194, opened 2017-05-22
only create flusher when vdisk is to be created

### LINKED ISSUE #193: tlog : newFlusher is called at each handshake
not efficient, because we only need it once on vdisk creation.

GlenDC:
fixes #193

### DIFF
--- a/tlog/tlogserver/server/server.go
+++ b/tlog/tlogserver/server/server.go
@@ -110,14 +110,14 @@ func (s *Server) ListenAddr() string {
 }
 
 // handshake stage, required prior to receiving blocks
-func (s *Server) handshake(r io.Reader, w io.Writer) (cfg *connectionConfig, err error) {
+func (s *Server) handshake(r io.Reader, conn net.Conn) (vd *vdisk, err error) {
 	status := tlog.HandshakeStatusInternalServerError
 
 	// always return response, even in case of a panic,
 	// but normally this is triggered because of a(n early) return
 	defer func() {
 		segmentBuf := make([]byte, 0, schema.RawServerHandshakeLen())
-		err := s.writeHandshakeResponse(w, segmentBuf, status)
+		err := s.writeHandshakeResponse(conn, segmentBuf, status)
 		if err != nil {
 			log.Infof("couldn't write server %s response: %s",
 				status.String(), err.Error())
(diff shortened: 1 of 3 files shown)


tlog/tlogserver/server/server.go
```diff
@@ -110,14 +110,14 @@ func (s *Server) ListenAddr() string {
 }
 
 // handshake stage, required prior to receiving blocks
-func (s *Server) handshake(r io.Reader, w io.Writer) (cfg *connectionConfig, err error) {
+func (s *Server) handshake(r io.Reader, conn *net.TCPConn) (vd *vdisk, err error) {
```
iwanbk: why changing from `io.Writer` to `*net.TCPConn`?

tlog/tlogserver/server/vdisk.go
```diff
@@ -78,25 +78,66 @@ func newVdiskManager(blockSize, flushSize int) *vdiskManager {
...
+
+	log.Debugf("create vdisk with expectedSequence:%v", vd.expectedSequence)
+
+	go vd.runFlusher()
+	go vd.runReceiver()
+
+	return
+}
+
+func (vt *vdiskManager) createOrGet(vdiskID string, f *flusher, firstSequence uint64) (vd *vdisk, created bool, err error) {
```
iwanbk: it seems this new func is not used at all?

GlenDC: oops, sorry, was from my first idea, this can be removed now.

tlog/tlogserver/server/server.go
GlenDC: because `func (s *Server) sendResp` requires `TCPConn`, but In theory that function could also take an `io.Writer`, so perhaps I should just do that instead :)

tlog/tlogserver/server/vdisk.go
GlenDC: Has been removed!

tlog/tlogserver/server/server.go
GlenDC: the handshake and sendResp function now use both the interface type `net.Conn`. This type is needed as you want to close the connection in the sendResp goroutine, which is spawned in the handshake function

iwanbk: > sendResp goroutine, which is spawned in the handshake function

Oh, just noticed this.
we can't do this. Because like the `vdisk`, the `sendResp` is going to be spawned only once. Then is going to use closed socket once the first connection closed.

GlenDC: Oh I see. This seems to be an issue indeed. Let's say a second connection gets opened. The vdisk's goroutines will still be writing to the first connection. This seems to be a flaw in the system, no @iwanbk? But that does seem outside the scope of this PR, as that's basically an issue separate from this. So perhaps you could merge this one, open a new issue where this issue is described and work on that today?

iwanbk: No, it is not a flaw.
vdisk only write to `respChan` channel, it doesn't know who will read from it.
when 2nd connection come, it spawns `sendResp` goroutine which reads from that respChan channel.

iwanbk: i've added a client reconnect test to master branch. It will help you to detect this error.

iwanbk: > No, it is not a flaw.

To make it clear, it is not a flaw in master branch code. But indeed a flaw of this PR.

GlenDC: Oh it is _definitely_ also a flaw in the master branch code. The difference is that you have a bug that hides this mistake. In the current master branch you _always_ create a new vdisk, because you don't store the newly created vdisk in the vdiskManager's map.

If you check https://github.com/g8os/blockstor/blob/2ce5d77be9e5a81682c1d1b8361bfede9ff56e57/tlog/tlogserver/server/vdisk.go#L82-L100, it clearly doesn't store the newly created `vd`, as you miss a line such as `vt.vdisks[vdiskID] = vd` at or right after https://github.com/g8os/blockstor/blob/2ce5d77be9e5a81682c1d1b8361bfede9ff56e57/tlog/tlogserver/server/vdisk.go#L94

Like I said before, this really is an issue that has nothing to do with this PR. I can and will of course still look into the issue and see if I can come with a solution. But out of scope it is never the less.

GlenDC: Pushed a new commit (https://github.com/g8os/blockstor/pull/194/commits/b4733905ca6d564b1d25715a3cea3dfbcfda1914) to resolve this issue. `(*server).sendResp` now works with multiple connections, doesn't close the connection anymore either. And `(*server).handle` and `(*server).sendResp` now both take an `io.Writer`, rather then a `net.Conn`. All your new tests pass as well.

iwanbk: > In the current master branch you always create a new vdisk, because you don't store the newly created vdisk in the vdiskManager's internal map.

Oh, my bad, you're right in this case, bad bug.
But still, your description of flaw in previous comment is not correct.

```
 resolve this issue. (*server).sendResp now works with multiple connections,
```
as far as i can see, it only works on single connection, which is good for me.

iwanbk: > Let's say a second connection gets opened. The vdisk's goroutines will still be writing to the first connection. This seems to be a flaw in the system, no @iwanbk?

Oh, i know what you meant now.

Yes, it is correct if it is the case for multiple connections (1st connection still open) :).  I was thinking about single connection (1 connection has been closed) because we indeed only support single connection now.

But to make it clear, it is not always to 1st connection, but both.
It also might not be a problem if we only use single client with multiple connection.

But still, you're correct :)

iwanbk: > resolve this issue. (*server).sendResp now works with multiple connections,
> as far as i can see, it only works on single connection, which is good for me.

I see what your patch fixes now.
In previous code, the `sendResp` will eventually be exited, because writing to closed connection will result in error and then the goroutine exit.

But it still consume one response from the channel.

Thanks for this!

GlenDC: So with this can you approve my PR and merge it? or are there still things to do?

PR #195, opened 2017-05-23
add g8stor list vdisks command

### LINKED ISSUE #173: Add g8stor list
It'd be nice if `g8stor`  has a list feature to list vdisks

GlenDC:
+ fixes #173

### DIFF
--- a/docs/g8stor/usage.md
+++ b/docs/g8stor/usage.md
@@ -13,13 +13,15 @@ Available Commands:
   copy        Copy a g8os resource
   delete      Delete a g8os resource
   help        Help about any command
+  list        List g8os resources
   restore     Restore a g8os resource
   version     Output the version information
 
 Flags:
+  -h, --help      help for g8stor
   -v, --verbose   log available information
 
-Use "g8stor [command] --help" for more information about a command
+Use "g8stor [command] --help" for more information about a command.
 ```
 
 ## Examples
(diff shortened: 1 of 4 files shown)


GlenDC:
Rebased master.

GlenDC:
@robvanmieghem would you prefer it to be a smaller Lua script, which does only 1 `SCAN` call, and than X `TYPE` calls, where `x` is the number of keys the `SCAN` has returned. So that we return the unique hash keys and the cursor. And do the `repeat until` loop on the Golang side. This would keep the time that the server is busy hopefully to a minimum.

So something like:

```lua
local cursor = ARGV[1]

local result = redis.call("SCAN", cursor)
local batch = result[2]

local key
local type
for i = 1, #batch do
	 key = batch[i]

	 -- only add hashmaps
	  type = redis.call("TYPE", key)
	  type = type.ok or type
	  if type == "hash" then
		 table.insert(vdiskids, key)
	   end
end

cursor = tonumber(result[1])
return {cursor, vdiskids}
```

I do agree that we should add a big warning in the help text and documentation. But this proposed version of the script, should not keep the server busy for just some milliseconds, as the `SCAN` key never returns only a small group of keys at once, so should be quite fast.

GlenDC:
The proposed script above would make the `g8stor list vdisks` command slower (as we need multiple TCP calls), but at least it would never lock the server for long. Which I think is what you prefer?

GlenDC:
> Cool but a bit afraid since a lua scripts locks the entire ardb during executing and this can take a while. At least I would add a big fat warning in the help texts so people know this.

@robvanmieghem I added the warning text, and I also now use the new updated Lua script, as discussed in my previous comments.

I also rebased master into this PR's branch.

GlenDC:
Rebased master.

PR #196, opened 2017-05-23
Add simple automatic client reconnect test.

iwanbk:


tlog/tlogclient/reconnect_test.go
```diff
@@ -0,0 +1,70 @@
...
+		assert.Equal(t, tr.Status, tlog.BlockStatusRecvOK)
+		assert.Equal(t, 1, len(tr.Sequences))
+		seqResp := tr.Sequences[0]
+		assert.Equal(t, x, seqResp)
+	}
+}
+
+func createTestServer() (*server.Server, *server.Config, error) {
+	conf := server.DefaultConfig()
+	conf.ListenAddr = "127.0.0.1:0"
```
GlenDC: No need to assign `"127.0.0.1:0"`, you can simply assign `""`, which will do the same thing, as auto-assigned `ListenAddr` is supported by the server.

tlog/tlogclient/reconnect_test.go
```diff
@@ -0,0 +1,70 @@
...
+	oneTenth := numLogs / 10
+
+	// send tlog, and disconnect it few times in the middle to test re-connect ability.
+	// to not create unnecessary complication, we do these:
+	// - we use sync funcs
+	// - send below flushSize
+	for i := 0; i < numLogs; i++ {
+		x := uint64(i)
+
+		if numLogs%oneTenth == 0 { // close the connection every one-tenth of numLogs
```
GlenDC: should be `i%oneTenth == 0` I think, as now you will close the connection _every_ log

iwanbk: wow, you're right, that's why there are so many logs about closed connection

Issue #188, opened 2017-05-22
add command to force tlog flush


iwanbk:

iwanbk:
**Implementation plan**
There are 2 things here:
(1) the protocol to send the command: 

- need to add another capnp struct, i think `TlogCommand` or `Command`  is good name. It only has one field: `commandType`
- before sending the `command` or `tlogBlock`, client needs to send one byte of message, containing the type of capnp struct it is going to send : `command` or `tlog block`
- server reads the command and forward it to `vdisk` via a channel

(2) the force flush itself

'runFlusher` need to `select` the command channel, and do force flush if it is a force flush command.
If we have more command in the future, we can start another goroutine which then forward it to `runFlusher`

iwanbk:
A WIP can be found on `force-flush` branch.
The test still failed  in travis, i can't reproduce it at all in my local PC.

iwanbk:
found the issue, it was race condition.

It happened when the client call `ForceFlush` but the blocks it already sent haven't reached the flush buffer yet. So, the client waiting forever.
It's been fixed by set a timer to the waiting process, retry the `ForceFlush` if it got timeout.

iwanbk:
The command currently doesn't have any response, i think it might be useful to give last sequence flushed to the response. It helps when the previous flush response was  missed by the client.

iwanbk:
Was thinking about adding optional sequence param to the force flush command. Was not sure that it is useful, but i'm sure that it is useful now.

If specified, the tlog server needs to wait until it have the given sequence in (ready to) flush buffer before doing the force flush. 
(in current implementation, force flush means doing the flush right now).
With this, this loop https://github.com/g8os/blockstor/pull/213/files#diff-40a85ef7b570df44705a555a2c2bb353R99 will become simpler.


Will implement it tomorrow as separate PR.

PR #209, opened 2017-05-28
add tlogstorage implementation and tests

### LINKED ISSUE #153: nbdserver with tlog : only Flush the operation after it get Flushed by tlog
part of #126

### LINKED ISSUE #178: tlogclient can't finish work at end of nbdserver's backend life cycle
At the moment, the nbdserver does not wait for the Backend's background routine (where we allow the tlogclient to do its work) to be finished. This is a problem, as it means that the tlog sent is corrupt.

On the tlogserver this event results in a log similar to the following:

```
EROR[05-18|16:26:00] failed to decode tlog: EOF               module=global caller=server.go:211
INFO[05-18|16:26:21] failed to send resp to :myvdisk, err:writev tcp 127.0.0.1:11211->127.0.0.1:62669: use of closed network connection module=global caller=server.go:235
^C
```

The solution will be to make the nbdserver _safely_ wait for the goroutine to be finished (a timeout value is to be used though, to prevent getting stuck due to a broken backend).

GlenDC:
+ fixes #153;
+ fixes #178;
+ add tlogStorage, which is a special kind of storage that wraps around any other storage;
+ this removes all tlog code from the backend, as that one know only has to deal with storage code once again;
+ no longer return merged content in storage merge function (as this is outdated code from an old refactor);
+ add unit tests for the new tlogStorage;
+ add extra unit tests for all storages;
+ improve some existing code;
+ add tlogStorage documentation;


GlenDC:
The current tlog storage implementation can be represented by following flow graph:

tlog_storage

(which is only a slighly updated version of the flow graph I posted in my proposal of this implementation in issue #153, which means that the proposal was pretty much on point).

There are still some issues with the current implementation though. The biggest one for now seems to be that the **flushing** doesn't happen correctly. Perhaps that's the only serious issue, it's certainly the only thing that's failing it all for now (AFAIK).

GlenDC:
@iwanbk I more or less finished my implementation of the tlogStorage, however...

I tried to simply do what you asked, and that is, simply waiting for a certain time (see: https://github.com/g8os/blockstor/blob/5ac18a9e09acc7b8bad7f6dfbc3ec325318ec2cc/nbdserver/ardb/tlog.go#L154-L166), however that really isn't reliable at all, more or less like I predicted. Whenever I try to do a manual test (writing an image), it fails because  the server didn't flush on time.

I _think_ we'll need to have https://github.com/g8os/blockstor/issues/188 working and implemented if we want this implementation ever work reliable. Because simply waiting for the tlog-server to be done flushing and sending us back the responses we desire, clearly doesn't seem to work... What do you think about that @iwanbk?

There might still be some issues with the tlogStorage implementation though, but nothing that the unit tests as they are seem to show. As those all pass fine on my computer. It seems to be that my DummyTlogServer is unpredictable, making unit tests fail now and then on Travis. I need to figure out if that's because there is an actual issue with the tlogStorage code, or simply that my DummyTlogServer (which actually is a client and a server) is flawed. Perhaps it was a dumb idea to add a dummy tlogServer, but the goal was so that I could be sure that the tlogStorage works, even if there is a flaw in the actual tlog Server+client. But it seems that the tests with the actual tlogserver+client do always pass, while the dummy one not :(

GlenDC:
While the `tlogStorage` as it is, does leave its impact on the heap profiling (due to the in-memory sequence caching), its activity is neglectable on the cpu profiling.

If you look at the following profile ([profile_tlogstorage_cpu.pdf](https://github.com/g8os/blockstor/files/1036938/profile_tlogstorage_cpu.pdf)) (taking while simply converting a 258MB image), you'll see that almost all of the heaping power goes to the `tlogClient`. So perhaps the bigger issue is the current implementation of the `tlogClient`.

iwanbk:
> I think we'll need to have #188 working and implemented if we want this implementation ever work reliable. Because simply waiting for the tlog-server to be done flushing and sending us back the responses we desire, clearly doesn't seem to work... What do you think about that @iwanbk?

Done at #213

>  It seems to be that my DummyTlogServer is unpredictable, making unit tests fail now and then on Travis

I added message type at #213, you might need to update your dummy server a little bit.

iwanbk:
> If you look at the following cpu profile (profile_tlogstorage_cpu.pdf) (taking while simply converting a 258MB image), you'll see that almost all of the cpu power goes to the tlogClient

Thanks for the graph, will work on it at #214

iwanbk:
> Whenever I try to do a manual test (writing an image), it fails because the server didn't flush on time.

By `on time`, did you mean `too slow`? Because if nbdserver has strict timing requirement, then i might missed something.

GlenDC:
@iwanbk thank you super much for 38e31931090986de17e6c03ba1475c0431053224! That seems to do the trick, the unit tests run very stable now! Also they run twice or trice as fast now, so good job 👍 👍 👍 

---

> By on time, did you mean too slow?

Yeah I meant too slow. Before commit 38e31931090986de17e6c03ba1475c0431053224, it took too long for a server to catch up with all the recent writes. So we could wait for a long time, but if you push it enough, it was still easy enough to break it. But since the force flush commit has been added, that problem seems to have been gone.

I'm sure there is still room for improvement, but this is a great beginning :)

GlenDC:
@iwanbk we do still get the log message from the tlog server at the end of a vdisk session (which is what #178 is about):

```
lvl=info msg="connection from 127.0.0.1:61523 dropped: EOF" module=global caller=server.go:103
```

At the moment I am not sure whether or not the case that triggers that message is harmful or not. Will need to investigate.

iwanbk:
> At the moment I am not sure whether or not the case that triggers that message is harmful or not. Will need to investigate.

i don't remember to added that logging.

If it is because the https://github.com/g8os/blockstor/pull/213#pullrequestreview-41020565, then it is not harmful at all.

GlenDC:
> i don't remember to added that logging.
> If it is because the #213 (review), then it is not harmful at all.

Oh you're right. I added it (https://github.com/g8os/blockstor/blob/38e31931090986de17e6c03ba1475c0431053224/tlog/tlogserver/server/server.go#L103), now I remember. It's not harmful no in that case, but in that case we should return `nil` there, so that it is nog logged. More information on that review comment :)

GlenDC:
`TestTlogStorageWithDeduped` fails right now once every 20 runs:

```
Running tool: /usr/local/go/bin/go test -timeout 30s -tags  -run ^TestTlogStorageWithDeduped$

t=2017-05-30T20:39:40-0500 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:54526" module=global caller=redis.go:47
t=2017-05-30T20:39:40-0500 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:54527" module=global caller=redis.go:47
t=2017-05-30T20:39:40-0500 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:54528" module=global caller=redis.go:47
t=2017-05-30T20:39:40-0500 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:54529" module=global caller=redis.go:47
t=2017-05-30T20:39:40-0500 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:54530" module=global caller=redis.go:47
t=2017-05-30T20:39:40-0500 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:54531" module=global caller=redis.go:47
t=2017-05-30T20:39:40-0500 lvl=info msg="Started listening on local address 127.0.0.1:54533" module=global caller=server.go:55
t=2017-05-30T20:39:40-0500 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:54532" module=global caller=redis.go:47
t=2017-05-30T20:39:40-0500 lvl=dbug msg="received handshake request from incoming connection" module=global caller=server.go:140
t=2017-05-30T20:39:40-0500 lvl=dbug msg="incoming connection checks out with version: 1.1.0-alpha" module=global caller=server.go:150
t=2017-05-30T20:39:40-0500 lvl=dbug msg="create vdisk with expectedSequence:0" module=global caller=vdisk.go:116
t=2017-05-30T20:39:40-0500 lvl=dbug msg="handshake phase successfully completed" module=global caller=server.go:167
t=2017-05-30T20:39:40-0500 lvl=dbug msg="replying handshake with status: OK" module=global caller=server.go:194
t=2017-05-30T20:39:40-0500 lvl=crit msg="tlog server resulted in error for vdisk a: read tcp 127.0.0.1:54534->127.0.0.1:54533: use of closed network connection" module=global caller=tlog.go:227
exit status 1
FAIL	github.com/g8os/blockstor/nbdserver/ardb	0.036s
Error: Tests failed.
```

For some reason the network is closed when sending the first block, even though the handshake was successful. Might be something vague triggered by LedisDB though. But I think I'll have to track it down, as this might give us false failures (if it is indeed a false failure and not an actual bug), down the line.

Have you ever seen this @iwanbk?

iwanbk:
> . Might be something vague triggered by LedisDB though.

Why you thought that it related to ledisdb?

> Have you ever seen this @iwanbk?
no

GlenDC:
> Why you thought that it related to ledisdb?

Ah no I don't think it's in ledisdb code itself, but perhaps the way we use it. Dunno, probably it's more somewhere in our codebase. Just saying that I only get that error during unit tests (which use ledisdb). I haven't seen it in manual tests (using redis).

GlenDC:
Found a bigger issue during manual testing @iwanbk.

my config file:

```yaml
storageClusters:
  mycluster:
    dataStorage:
      - address: localhost:6379
    metadataStorage:
      address: localhost:6379
  tlogcluster:
    dataStorage:
      - address: localhost:16379
      - address: localhost:16380
      - address: localhost:16381
      - address: localhost:16382
      - address: localhost:16383
      - address: localhost:16384
    metadataStorage:
      address: localhost:16379
vdisks:
  myvdisk:
    size: 10
    blockSize: 4096
    readOnly: false
    storageCluster: mycluster
    tlogStorageCluster: tlogcluster
    type: boot
```

What I do:

1. start tlogserver: `bin/tlogserver -storage-addresses localhost:16379 -flush-size=25 -flush-time=25 -v`
2. start nbdserver : `bin/nbdserver -protocol tcp -address localhost:6666 -tlogrpc localhost:11211 -v`
3. write image using qemu: `qemu-img convert -p -n myimg.img -O nbd nbd+tcp://localhost:6666/myvdisk`
4. boot up image using qemu: `qemu-system-x86_64 -machine pc-1.3,usb=off -boot order=c -m 2048 -smp 2 -net nic -net user -drive file=nbd+tcp://localhost:6666/myvdisk,index=0,format=raw`

What do I see in the logs (among other things):

```
DBUG[05-30|21:06:49] Loaded storagecluster 'mycluster' configuration (1 servers) module=global caller=cluster.go:335
INFO[05-30|21:06:49] ready to reload StorageClusterConfig upon SIGHUP receival for:myvdisk module=storagecluster caller=cluster.go:230
DBUG[05-30|21:06:49] creating tlogclient for backend myvdisk (boot) module=global caller=ardb.go:152
DBUG[05-30|21:06:49] generating backend ardb, using 333 flags, for 127.0.0.1:56321 module=nbdserver caller=connection.go:1120
INFO[05-30|21:06:49] Negotiation succeeded with 127.0.0.1:56321/myvdisk module=nbdserver caller=connection.go:609
DBUG[05-30|21:06:49] Start transmition phase with127.0.0.1:56321/myvdisk module=nbdserver caller=connection.go:624
DBUG[05-30|21:07:00] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:00] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:00] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:00] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:00] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:00] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:00] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:00] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:00] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:00] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:00] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:00] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:01] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:01] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:01] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:01] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:01] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:01] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:02] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:02] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:02] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:03] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:03] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:03] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:03] sequence cache for tlog storage myvdisk is not yet empty map[0:4224] module=global caller=tlog.go:178
INFO[05-30|21:07:03] Client 127.0.0.1:56321/myvdisk got flush I/O error: sequence cache for tlog storage myvdisk is not yet empty module=nbdserver caller=connection.go:477
DBUG[05-30|21:07:03] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:03] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:03] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:03] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:03] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:03] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:04] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:04] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:04] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
DBUG[05-30|21:07:04] vdisk myvdisk's tlog server has received force flush message module=global caller=tlog.go:246
INFO[05-30|21:07:04] tried to evict invalid blockIndex: 69770 module=global caller=tlog.go:442
DBUG[05-30|21:07:04] vdisk myvdisk: storing 0 block(s) for tlog's flushed sequences [69770] module=global caller=tlog.go:283
```

As you can see, for some reason does the `tlogserver` think that the sequence number is differently, then the one stored in the `tlogStorage`. Because we have a new session, we also have a fresh vdisk and thus fresh `tlogStorage`, meaning that our sequenceIndex is now starting from `0` again. however the `tlogserver` never resets.

This can easily been seen, as my manual test does work, when you restart the `tlogserver` manually between step 3 and 4 of the steps described above.

I hacked a way around this issue by doing:

```diff
diff --git a/tlog/tlogserver/server/vdisk.go b/tlog/tlogserver/server/vdisk.go
index 03777ad..445bb73 100644
--- a/tlog/tlogserver/server/vdisk.go
+++ b/tlog/tlogserver/server/vdisk.go
@@ -99,6 +99,8 @@ func (vt *vdiskManager) Get(vdiskID string, firstSequence uint64, ff flusherFact
 
 	vd, ok := vt.vdisks[vdiskID]
 	if ok {
+		// HACK: reset expected sequence
+		vd.expectedSequence = firstSequence
 		return
 	}
 ```

But this is obviously a hack, as this will break as soon as we start using multiple connections, which is something you want to support @iwanbk. So because of that I don't think we can simply reset this. One fake solution I see is:

+ during the handshake phase, the server gives a firstSequence back, which will be either the same as the one given by the client (which is the case if it's a new vdisk), or it will be a different one (which is the case if it's an existing vdisk);
  + this would allow for multiple connections;
  + however this would still be prone to error, as other connections might still be active, and changing the expectedSequence value right after that value was given to the new sequence, thus still breaking that new connection as we see now as well;

We can also not use a new vdisk for each connection, as this would also not give the desired result for multiple connections. So... yeah. I'm not sure what to do. But perhaps this should become a separate issue, and I can simply have this PR with this hack (as my manual test does work with this hack in place, without having to restart the tlogserver between step 3 and 4). And we would have a new milestone-3 issue to resolve this issue with the `expectedSequence` tlogserver's vdisk state...

**Comment by iwanbk** [comment]
> Because we have a new session, we also have a fresh vdisk and thus fresh tlogStorage, meaning that our sequenceIndex is now starting from 0 again. however the tlogserver never resets.

Yep, i actually already addressed this issue in previous issue and was thinking to create new issue for this today.
See my comments in previous issue 
https://github.com/g8os/blockstor/issues/166#issue-229219759
```
nbdserver restarted and reset the sequence to zero.
```

https://github.com/g8os/blockstor/issues/166#issuecomment-301980071
```
nbdserver can also force vdisk object to change it's expected sequence. It is needed when vdisk object already exist in tlog and nbdserver has no idea about the correct expected sequence and actually nbdserver doesn't really care about it.
```
Your PR which solve that issue didn't handle it, so i thought that ndbserver has no problem with it.

But last night i thought that it must be difficult for nbdserver to handle it.

My suggestion still the same, provide the ability for the ndbserver to force change the expected first sequence.
If you're OK, i can provide this feature today.

**Comment by GlenDC** [comment]
> My suggestion still the same, provide the ability for the ndbserver to force change the expected first sequence. If you're OK, i can provide this feature on tlog server today

I'm 100% OK with you providing this feature on the tlog server today. As that would be that last issue on your side, standing in the way of this PR being able to be merged in.

Than there is still some weird race condition issue in my test code, that makes the Dummy `TestDummyTlogStorageDeadlock` test fail most of the time. But that doesn't affect product code, as this is clearly related to some flaw in my `dummyTlogClient`. So will focus on that now, as that's the last thing I have to do, before this PR is ready I think.

So yeah @iwanbk. Please do work on making that feature into the tlog server, as that would make this PR ready. I'll also try to do the thing I just said. If we can both finish than this PR would be ready for review+merge tomorrow.

Please do enlighten me on how you would be able to give the nbdserver full control over the firstSequence (and thus expected sequence), WHILE still supporting multiple connections o.o As I really don't see how you would do that.

**Comment by iwanbk** [comment]
```
So yeah @iwanbk. Please do work on making that feature into the tlog server,
```
OK, will start in next few hours

```
Please do enlighten me on how you would be able to give the nbdserver full control over the firstSequence (and thus expected sequence), WHILE still supporting multiple connections o.o As I really don't see how you would do that.
```
This sequence reset only happen on fresh vdisk, also means fresh new tlog client.
(Please note that what we want to support is multiple connections, not multiple clients.)
I think is simple to limit the very first handshake to only one connection, no?

So, `full control` might not appropriate word because it only happens on initialization.

If we also need to reset the sequence in the middle of nbdserver operation (means : not new tlogclient), we can stop all connections and recreate it. It might not simple, but still doable.
But i don't think we need this.

**Comment by GlenDC** [comment]
> Please note that what we want to support is multiple connections, not multiple clients.

Oh that makes sense. So you mean you want to support multiple connections, as long as the connections are not occurring at the same moment.

In that case you will need to add some code to guarantee this though. As for now you could easily connect multiple connections AT THE SAME TIME to the same vdisk, which would screw things up big time, with your plan.

So you will need some state (could be as simple as a bool) that states that the (already existing vdisk) is busy at the moment, and thus that client can't connect and its handshake will fail in that case.

Does that make sense?

**Comment by iwanbk** [comment]
> Oh that makes sense. So you mean you want to support multiple connections, as long as the connections are not occurring at the same moment.

No, it is single client with multiple TCP connections

**Comment by GlenDC** [comment]
> No, it is single client with multiple TCP connections

Oh... Well if you can make it work, I'm all happy with that.

How will that work with the handshake and stuff?

Can you describe some flows, as I really don't se how that will work.

**Comment by GlenDC** [comment]
> Than there is still some weird race condition issue in my test code, that makes the Dummy TestDummyTlogStorageDeadlock test fail most of the time. But that doesn't affect product code, as this is clearly related to some flaw in my dummyTlogClient. So will focus on that now, as that's the last thing I have to do, before this PR is ready I think.

I'm removing the `testDummyTlogClient` as it is extra code to maintain, which is never used in production, and so far it has never helped me catch any extra bugs. On top of that, its implementation is flawed. So I think it's better to simply have the code as it is, but without the dummy unit tests.

Pushed the new code to the branch of this PR. With that done, I am finished. So @iwanbk feel free to review and merge the code as it is. If you're happy with it, it could go in master, and this way you can also test your work of today directly against this code.

File: nbdserver/ardb/tlog.go [hunk 1]
```diff
@@ -0,0 +1,578 @@
...
+				continue
+			}
+
+			switch res.Resp.Status {
+			case tlog.BlockStatusRecvOK:
+				// nothing to do, logging would be way too verbose
+
+			case tlog.BlockStatusForceFlushReceived:
+				log.Debugf("vdisk %s's tlog server has received force flush message", tls.vdiskID)
+				err := tls.flushCachedContent(res.Resp.Sequences)
```
iwanbk: I'm afraid you use response message of `ForceFlush` incorrectly.  (Sorry if still lack of docs). 
From my PR https://github.com/g8os/blockstor/pull/213#issuecomment-304815689

> force flush response is the last sequence flushed by vdisk

While in this line, it seems you use it as `array of flushed sequences`, please note that `tlog.BlockStatusForceFlushReceived` only indicates that the `ForceFlush` has been **received** by the server

nbdserver/ardb/tlog.go
```diff
@@ -0,0 +1,578 @@
...
+				if err != nil {
+					log.Fatalf(
+						"failed to write cached content into storage for vdisk %s: %s",
+						tls.vdiskID, err)
+				}
+
+			case tlog.BlockStatusFlushOK:
+				err := tls.flushCachedContent(res.Resp.Sequences)
```
iwanbk: In this line, you use the response message correctly

nbdserver/ardb/tlog.go
```diff
@@ -0,0 +1,578 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// get the next sequence index,
+// and increment if such that the next time another one can be retrieved
+func (tls *tlogStorage) getNextSequenceIndex() (sequence uint64) {
```
iwanbk: If we use same architecture as master code, we won't need this log anymore

nbdserver/ardb/tlog.go
```diff
@@ -0,0 +1,578 @@
...
+	tls.sequence++
+	return
+}
+
+// tlogClient represents the tlog client interface used
+// by the tlogStorage, usually filled by (*tlogclient.Client)
+//
+// using an interface makes it possible to use a dummy version
+// for testing purposes
+type tlogClient interface {
```
iwanbk: IIRC you already deleted the dummy tlog client, so this code seems to be not used anymore

nbdserver/ardb/tlog.go
```diff
@@ -0,0 +1,578 @@
...
+// Delete implements backendStorage.Delete
+func (tls *tlogStorage) Delete(blockIndex int64) (err error) {
+	err = tls.Set(blockIndex, nil)
+	return
+}
+
+// Flush implements backendStorage.Flush
+func (tls *tlogStorage) Flush() (err error) {
+	const sleepTime = 128 * time.Millisecond
+	for i := 0; i < 16; i++ {
```
iwanbk: This kind of semi busy waiting also not really good.
It at least abuses the `ForceFlush` facility.
Just realise that my unit test might give bad example.

I suggest two alternatives, all of them need the ndbserver to knows which sequence number that MUST be flushed, lets call this sequence as `lastSeq`:
- wait until tlog client recv `BlockStatusRecvOK` for `lastSeq` and then send the `ForceFlush`
- more performant and simple alternative is to implement my idea mentioned at https://github.com/g8os/blockstor/issues/188#issuecomment-304852303 and using it. What do you think?

nbdserver/ardb/tlog.go
```diff
@@ -0,0 +1,578 @@
...
+	// even though it is stored in the replay server
+	err := tls.cache.Add(sequence, blockIndex, content)
+	if err != nil {
+		return fmt.Errorf(
+			"tlogStorage couldn't set content at block %d (seq: %d): %s",
+			blockIndex, sequence, err)
+	}
+
+	// send transaction to tlog server
+	go func() {
```
GlenDC: Yeah you're right, I'm going to try that, and let's see what happens.

nbdserver/ardb/tlog.go
GlenDC: I think it would still be good to have this as an interface (but perhaps it should move to the tlogclient pacakge), this way the tlogclient package could also provide alternative/stub/test implementations in the future. Always better to expose interfaces over direct structs if possible.

nbdserver/ardb/tlog.go
GlenDC: Could be, but we still have to wait for a magical amount of time I guess. But yeah, doing so with more knowledge and actually knowing when to force flush is probably a good idea yes.

nbdserver/ardb/tlog.go
iwanbk: s/log/lock

nbdserver/ardb/tlog.go
iwanbk: agree

nbdserver/ardb/tlog.go
GlenDC: We still require this lock, as we require the sequence index upfront, so we can immediately add the data in the cache, so that it's already available for usage (eg getting it in order to merge it already with other data). If we would only add it in cache when the channel receives it on the background routine, it will give us very subtle bugs, where data is corrupt due to failed merges or w/e. And because the new merge would have a new sequence number, that would overwrite the previous data.

nbdserver/ardb/tlog.go
```diff
@@ -0,0 +1,578 @@
...
+			op,       // opcode
+			sequence, // sequence
+			uint64(blockIndex*tls.blockSize), // offset
+			uint64(time.Now().Unix()),        // timestamp
+			content,                          // data
+			length,                           // data length
+		)
+
+		if err != nil {
+			log.Fatalf(
```
GlenDC: There is no graceful way at the moment for a backend to exit prematurely. But this panic will be catched, and it will be logged. So at least the user will be stopped from using a vdisk which transactions are not properly stored within the tlog server. Is that not what you want?

nbdserver/ardb/tlog.go
GlenDC: Good catch. I now do it over a channel, on a seperate goroutine (similar to how it used to be done). This means the nbdserver is slowed down, but that does mean that at least the tlogserver has more time to finish. And now I can work even with big images 👍 very good!

nbdserver/ardb/tlog.go
GlenDC: Oh wait @iwanbk you are right, we should probably actually throw a panic, as log.Fatalf will actually exit the nbdserver. Good catch, will update this!

GlenDC:
It seems that the `TestTlogStorageDeadlockWithDeduped` is not stable, and fails now and then on travis. Need to investigate this, as it might indicate a bug. If it does indicate a bug, we probably also need an extra unit test (case) so this kind of issue can be discovered consistently, rather than by cheer luck.

Here is the log:

```
t=2017-05-31T23:26:42+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:38729" module=global caller=redis.go:47
t=2017-05-31T23:26:42+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:34920" module=global caller=redis.go:47
t=2017-05-31T23:26:42+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:36447" module=global caller=redis.go:47
t=2017-05-31T23:26:42+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:44908" module=global caller=redis.go:47
t=2017-05-31T23:26:42+0000 lvl=info msg="Started listening on local address 127.0.0.1:35382" module=global caller=server.go:55
t=2017-05-31T23:26:42+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:35874" module=global caller=redis.go:47
t=2017-05-31T23:26:42+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:46219" module=global caller=redis.go:47
t=2017-05-31T23:26:42+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:34591" module=global caller=redis.go:47
t=2017-05-31T23:26:42+0000 lvl=dbug msg="received handshake request from incoming connection" module=global caller=server.go:140
t=2017-05-31T23:26:42+0000 lvl=dbug msg="incoming connection checks out with version: 1.1.0-alpha" module=global caller=server.go:150
t=2017-05-31T23:26:42+0000 lvl=dbug msg="create vdisk with expectedSequence:0" module=global caller=vdisk.go:140
t=2017-05-31T23:26:42+0000 lvl=info msg="vdisk a reset first sequence to 0" module=global caller=vdisk.go:210
t=2017-05-31T23:26:42+0000 lvl=dbug msg="handshake phase successfully completed" module=global caller=server.go:175
t=2017-05-31T23:26:42+0000 lvl=dbug msg="replying handshake with status: OK" module=global caller=server.go:202
t=2017-05-31T23:26:42+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:42+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:43+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:43+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:44+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:44+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:45+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:45+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:46+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:46+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:47+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:47+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:48+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:48+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:49+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:49+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:50+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:50+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:51+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:51+0000 lvl=dbug msg="vdisk a's tlog server has received force flush message" module=global caller=tlog.go:244
t=2017-05-31T23:26:51+0000 lvl=dbug msg="background context for tlogStorage a aborting" module=global caller=tlog.go:261
t=2017-05-31T23:26:51+0000 lvl=dbug msg="transaction sender for tlogStorage a aborting" module=global caller=tlog.go:292
t=2017-05-31T23:26:51+0000 lvl=info msg="connection from 127.0.0.1:43892 dropped" module=global caller=server.go:105
t=2017-05-31T23:26:51+0000 lvl=dbug msg="abort current sendResp goroutine for vdisk:a" module=global caller=server.go:287
--- FAIL: TestTlogStorageDeadlockWithDeduped (9.28s)
	ardb_test.go:466: 6 9 unexpected content received [249 227 152 120 85 77 40 233 214 130 234 18 197 31 243 52 60 51 176 213 0 91 228 26 133 149 8 156 76 199 210 95 175 189 189 156 156 52 98 223 238 179 111 63 203 217 151 43 94 126 165 201 195 224 29 157 69 101 182 101 118 217 254 55 30 132 117 107 220 201 2 109 54 124 195 221 4 177 128 251 53 58 68 194 25 156 65 78 72 14 35 97 182 159 84 76 231 84 90 209 212 103 79 78 108 78 104 1 8 147 51 157 249 0 232 74 40 217 66 145 24 131 167 131 68 14 98 74] [14 195 84 4 233 238 110 15 218 152 18 97 156 81 193 59 101 42 160 164 78 133 89 247 98 203 135 134 86 199 203 115 94 44 223 126 195 88 149 146 176 140 12 200 1 91 6 50 27 223 210 39 125 183 182 226 169 24 11 127 31 73 225 72 184 1 207 255 252 9 12 249 27 154 185 103 239 181 90 199 69 160 106 49 60 42 162 1 183 134 29 166 102 83 123 69 119 183 99 189 100 23 92 148 211 37 1 185 220 175 122 253 92 93 121 204 130 198 29 134 3 99 136 21 50 0 225 168]
	ardb_test.go:466: 6 11 unexpected content received [95 66 16 72 4 219 56 49 35 194 3 103 44 40 4 213 251 198 188 224 184 110 21 233 124 28 92 95 81 42 221 48 209 210 200 158 169 162 60 125 110 129 178 224 13 209 109 76 198 192 205 111 31 46 109 217 198 152 228 229 175 113 65 153 82 63 232 79 150 202 49 204 178 20 95 51 82 171 152 229 194 11 163 146 150 141 111 147 251 103 25 215 96 52 17 41 157 102 204 197 44 179 17 171 78 153 41 212 90 53 8 151 129 60 126 179 119 81 87 77 163 142 212 27 90 252 50 74] [232 80 76 223 203 151 60 116 233 215 123 163 181 147 177 174 74 107 217 77 232 4 34 227 239 201 228 224 71 103 23 207 196 97 78 92 109 195 120 253 198 131 96 128 225 194 25 209 137 173 37 181 129 105 209 76 118 214 188 28 199 17 26 205 238 197 42 37 163 27 242 109 83 209 141 51 128 159 133 190 45 79 224 182 119 17 165 170 168 222 246 63 50 81 117 237 12 240 88 204 119 121 30 74 135 58 116 188 134 156 143 82 155 251 59 52 157 100 245 235 174 124 86 151 187 48 112 179]
	ardb_test.go:466: 6 15 unexpected content received [73 202 161 145 168 184 189 253 19 97 124 94 125 107 225 182 33 77 121 26 80 244 56 214 18 75 17 246 86 25 113 245 138 217 253 86 37 234 51 163 110 189 218 189 178 156 184 134 133 115 149 153 156 13 161 89 99 12 247 238 144 198 90 185 16 118 121 24 188 40 17 126 141 64 21 134 208 173 244 162 149 67 126 186 185 247 232 49 241 131 126 121 139 8 220 253 105 122 125 174 56 6 70 35 61 164 222 19 210 84 243 84 9 137 85 112 121 249 81 240 107 135 70 74 47 206 205 52] [164 81 224 158 152 133 101 68 84 181 162 175 76 130 130 22 190 248 173 131 73 141 230 18 122 167 208 60 152 113 38 89 198 209 151 118 117 107 65 148 183 153 179 226 140 9 70 134 197 127 88 155 83 12 138 244 23 107 63 12 27 146 228 3 71 222 16 24 224 42 92 15 200 20 175 111 172 200 51 155 201 5 167 98 81 81 57 28 146 188 97 12 183 139 52 219 245 238 254 222 74 194 69 137 137 253 53 118 221 190 230 71 114 20 139 150 10 199 95 93 236 29 195 11 58 226 185 223]
```

iwanbk:
> It seems that the TestTlogStorageDeadlockWithDeduped is not stable,

What this test is all about?

> Here is the log:

what the logs told us?

GlenDC:
> What this test is all about?

Originally the inner test (`testBackendStorageDeadlock`) was designed as a response to issue https://github.com/g8os/blockstor/issues/89, as to make sure that no storage would ever get stuck in a deadlock, after many operations (async), due to for example not closing connections. Which was a bug back in the days in the dedupedStorage.

Now I use it for every storage just to guarantee that, and it also is a nice test that sets, merges and gets a lot of data at once, which helps a lot to detect bugs by accident.

> what the logs told us?

The log doesn't tell us much about the actual issue I'm afraid.

It does tell us that at some point, the content it has (using `Get`) is different the content it just has set (using `Set`). Which makes me think there is still somewhere in my `ardb/tlog.go` code a race condition or something that can potentially fail based on some multiple async operations at once. Which is of course bad, as it should always work.

That's the beauty of these generic `ardb_test.go` tests, they allow us to make sure that we can use every storage the same way, which is important as the backend does exactly that.

GlenDC:
@iwanbk i think the issues were more related to some non-thread safe unit test code. Pushed a commit to fix this, and it seems to pass all tests once again. Tried to test it many times too, and it seems to always pass now, so I think we're OK.

GlenDC:
Oh, guess I spoke too early @iwanbk. Ran the test suites a couple of times using `make testcgo` as well on my laptop, with 5% battery. And now I do get a fail due to an EOF panic: ``

This is triggered from here: https://github.com/g8os/blockstor/blob/ccee4b9dcffdd8e26777a64ec3d83b3bba23d9e0/nbdserver/ardb/tlog.go#L229-L231

Perhaps we should filter out EOF errors at that stage, just as the server does. Or is the current code correct and should there never be an EOF error at that stage. If however an EOF error at that stage is normal, we could change that code to:

```go
if res.Err != nil {
  if res.Err == io.EOF {
      continue // EOF in this stage is not an error
  }
  panic(fmt.Errorf(
      "tlog server resulted in error for vdisk %s: %s",
      tls.vdiskID, res.Err))
}
```

I think this error is being triggered at the end of a `testBackendStorageDeadlock` tlog test. But not too sure.

GlenDC:
I do think that this issue is unrelated to this PR though, as this one seems to be more due to an issue somewhere in the tlogclient/tlogserver code. So perhaps we can make this into a separate issue, and review+merge this PR first?

nbdserver/ardb/tlog.go
iwanbk: created new issue for this

PR #213, opened 2017-05-30
Force flush

### LINKED ISSUE #188: add command to force tlog flush

iwanbk:
Closes #188

### DIFF
--- a/tlog/tlogserver/server/server.go
+++ b/tlog/tlogserver/server/server.go
@@ -209,26 +209,51 @@ func (s *Server) handle(conn *net.TCPConn) error {
 	go s.sendResp(ctx, conn, vdisk.ID(), vdisk.ResponseChan())
 
 	for {
-		// decode
-		block, err := s.ReadDecodeBlock(br)
+		msgType, err := tlog.ReadCheckMessageType(br)
 		if err != nil {
-			log.Errorf("failed to decode tlog: %v", err)
+			if err != io.EOF { // EOF in this stage is not an error
+				log.Errorf("failed to read message type: %v", err)
+			}
 			return err
 		}
 
-		// check hash
-		if err := s.hash(block, vdisk.vdiskID); err != nil {
-			log.Debugf("hash check failed:%v\n", err)
+		switch msgType {
+		case tlog.MessageForceFlush:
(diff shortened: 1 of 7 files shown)


iwanbk:
Changes, based on plan mentioned at https://github.com/g8os/blockstor/issues/188#issuecomment-304620462:

- add message type before sending the actual data (tlog block or command)
- force flush doesn't need any data, so we only need to send the message type
- force flush response is the last sequence flushed by vdisk

tlog/tlogserver/server/server.go
```diff
@@ -209,28 +209,58 @@ func (s *Server) handle(conn *net.TCPConn) error {
 	go s.sendResp(ctx, conn, vdisk.ID(), vdisk.ResponseChan())
 
 	for {
-		// decode
-		block, err := s.ReadDecodeBlock(br)
+		msgType, err := tlog.ReadCheckMessageType(br)
 		if err != nil {
-			log.Errorf("failed to decode tlog: %v", err)
+			if err != io.EOF { // EOF in this stage is not an error
```
GlenDC: If it's not an error, why do we _do_ return it still as an error?

iwanbk: you're right. Maybe the `handle` func shouldn't return error at all, the returned error is useless anyway

GlenDC: Ah ok, but in general I think you can just return nil in here.

So we could change this code as:

```go
if err != nil {
  if err == io.EOF { // EOF in this stage is not an error
      err = nil
  }
  return err
}
```

and then we should probably change the code @ 
https://github.com/g8os/blockstor/blob/38e31931090986de17e6c03ba1475c0431053224/tlog/tlogserver/server/server.go#L103 to something like:

```go
err := s.handle(tcpConn)
if err == nil {
    log.Infof("connection from %s dropped", addr.String())
} else {
    log.Errorf("connection from %s dropped with error: %s", addr.String(), err.Error())
}
```

If you agree with this I can push a small commit for this to master, so that I can rebase it into my branch as well.

iwanbk: OK

PR #233, opened 2017-06-02
Force flush seq

### LINKED ISSUE #228: Add optional sequence param to tlog ForceFlush
See https://github.com/g8os/blockstor/issues/188#issuecomment-304852303

> 
> Was thinking about adding optional sequence param to the force flush command. Was not sure that it is useful, but i'm sure that it is useful now.
> 
> If specified, the tlog server needs to wait until it have the given sequence in (ready to) flush buffer before doing the force flush.
> (in current implementation, force flush means doing the flush right now).

iwanbk:
Closes #228


iwanbk:
oh, this line 
https://github.com/zero-os/0-Disk/blob/master/tlog/tlogserver/server/vdisk.go#L372-L373

missed the `continue`, will fix it on zero-os repo

PR #234, opened 2017-06-03
Fix force flush race

### LINKED ISSUE #228: Add optional sequence param to tlog ForceFlush

iwanbk:
Closes #228

### DIFF
--- a/tlog/tlogserver/server/common_test.go
+++ b/tlog/tlogserver/server/common_test.go
@@ -1,0 +1,21 @@
+package server
+
+import (
+	"testing"
+
+	"github.com/zero-os/0-Disk/tlog"
+)
+
+func createTestServer(t *testing.T, flushTime int) (*Server, *Config, error) {
+	conf := DefaultConfig()
+	conf.ListenAddr = "127.0.0.1:0"
+	if flushTime > 0 {
+		conf.FlushTime = flushTime
+	}
+
+	// create inmemory redis pool factory
+	poolFactory := tlog.InMemoryRedisPoolFactory(conf.RequiredDataServers())
+
+	s, err := NewServer(conf, poolFactory)
+	return s, conf, err
+}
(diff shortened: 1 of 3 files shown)


tlog/tlogserver/server/common_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"testing"
+
+	"github.com/zero-os/0-Disk/tlog"
+)
+
+func createTestServer(t *testing.T, flushTime int) (*Server, *Config, error) {
+	conf := DefaultConfig()
+	conf.ListenAddr = "127.0.0.1:0"
```
GlenDC: I've said this before, but you can simply set it empty, `conf.ListenAddr = ""`, will achieve you the same, except that it will also support IPv6 if available. And you know, it's shorter to type 💃

tlog/tlogserver/server/vdisk.go
```diff
@@ -363,14 +365,13 @@ func (vd *vdisk) runFlusher() {
 			switch cmdType {
 			case vdiskCmdForceFlushAtSeq:
 				seqToForceFlush = flusherCmd.sequence
-				if vd.expectedSequence <= seqToForceFlush { // we don't have it yet
+				if maxSeq < seqToForceFlush { // we don't have it yet
```
GlenDC: seems like you can refactor this to `needForceFlushSeq = maxSeq < seqToForceFlush`, no?

GlenDC: Ahno, wait nevermind. The more important one is the continue statement, as to not flush, so nevermind my comment :| *hides*

PR #238, opened 2017-06-05
nbdserver tlogStorage flush rework

### LINKED ISSUE #227: nbdserer tlogStorage.Flush : fix the ForceFlush call
As pointed by me in https://github.com/g8os/blockstor/pull/209/files#r119353644

> 
> It at least abuses the ForceFlush facility.
> Just realise that my unit test might give bad example

iwanbk:
Fixes #227 
- fix misuse of global variable inside a closure in ardb_test
- replace the `ForceFlush` + busy waiting with `ForceFlushAtSeq` + condition variable.


iwanbk:
The travis is failing with incorrect hash issue, i think i need to work on that issue because it is getting annoying

PR #246, opened 2017-06-07
Tlogclient reconnect from read

### LINKED ISSUE #226: tlogclient : Filter out io.EOF error
Possible bug was reported at https://github.com/g8os/blockstor/pull/209#issuecomment-305367738

iwanbk:
Fixes #226

### DIFF
--- a/tlog/tlogclient/client.go
+++ b/tlog/tlogclient/client.go
@@ -19,8 +19,8 @@ import (
 const (
 	sendRetryNum     = 5
 	sendSleepTime    = 500 * time.Millisecond // sleep duration before retrying the Send
-	resendTimeoutDur = 5 * time.Second        // duration to wait before re-send the tlog.
-	readTimeout      = 5 * time.Second
+	resendTimeoutDur = 2 * time.Second        // duration to wait before re-send the tlog.
+	readTimeout      = 2 * time.Second
 )
 
 // Response defines a response from tlog server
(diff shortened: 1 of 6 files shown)


tlog/tlogclient/client.go
```diff
@@ -80,19 +82,37 @@ func New(addr, vdiskID string, firstSequence uint64, resetFirstSeq bool) (*Clien
...
 
-func (c *Client) connect(firstSequence uint64, resetFirstSeq bool) error {
+// reconnect to server
+// it must be called under wLock
+func (c *Client) reconnect(closedTime time.Time) (err error) {
+	if c.lastConnected.After(closedTime) { // another goroutine made the connection work again
+		return
+	}
+
+	if err = c.connect(c.blockBuffer.MinSequence(), false); err == nil {
```
GlenDC: I would put a debug log statement for the fail case (` } else { /* log */ }`), as might be interesting to have as part of the verbose logging.

tlog/tlogclient/client.go
```diff
@@ -327,5 +369,12 @@ func (c *Client) send(op uint8, seq, offset, timestamp uint64,
 // It is user responsibility to call this function.
 func (c *Client) Close() error {
 	c.cancelFunc()
+
+	c.wLock.Lock()
+	defer c.wLock.Unlock()
+
+	c.rLock.Lock()
+	c.rLock.Unlock()
```
GlenDC: shouldn't this be `defer c.rLock.Unlock()`, otherwise it seems a bit useless to Lock it in the first case.

tlog/tlogclient/client.go
iwanbk: the caller do the logging

tlog/tlogclient/client.go
iwanbk: oh, you're right, my bad.
Fortunately the other goroutine already dead.
But it still a mistake

PR #249, opened 2017-06-07
Fix nbdserver freeze

### LINKED ISSUE #247: nbdserver freeze when converting big image
When converting ubuntu image, the process couldn't be finished.
And i found that there are many too many logs/blocks in the block buffer.
Possible issues:
- bug in client's block buffer
- bug in client receiver
- bug in nbdserver's tlogStorage

iwanbk:
Fixes #247

### DIFF
--- a/nbdserver/ardb/tlog.go
+++ b/nbdserver/ardb/tlog.go
@@ -17,7 +17,7 @@ import (
 
 const (
 	// maximum duration we wait for flush to finish
-	tlogFlushWait = 10 * time.Second
+	tlogFlushWait = 10 * time.Millisecond
 )
 
 func newTlogStorage(vdiskID, tlogrpc string, blockSize int64, storage backendStorage) (backendStorage, error) {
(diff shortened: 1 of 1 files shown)


iwanbk:
Another improvement we can make is on the tlogclient side #250

nbdserver/ardb/tlog.go
```diff
@@ -208,27 +218,25 @@ func (tls *tlogStorage) Flush() (err error) {
 		// if timeout, signal the condition variable anyway
 		// to avoid the goroutine blocked forever
 		tls.cacheEmptyCond.Signal()
```
GlenDC: Maybe I misunderstand something, but doesn't this only unblock the goroutine only once, because it will still be in an infinite loop `for` loop, no? Thus next iteration it will be waiting again, as we are stopping only because of a timeout, meaning the cache will still not be empty, right?

Don't we need another mechanism to relaly exit that goroutine?

iwanbk: you're right, it was `if`, not sure why i changed it to `for`. Fixing it now

Issue #252, opened 2017-06-07
TestForceFlush got stuck in a travis build


GlenDC:
Log:

```
t=2017-06-07T15:07:32+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:41711" module=global caller=redis.go:47
t=2017-06-07T15:07:32+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:41255" module=global caller=redis.go:47
t=2017-06-07T15:07:32+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:35015" module=global caller=redis.go:47
t=2017-06-07T15:07:32+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:36941" module=global caller=redis.go:47
t=2017-06-07T15:07:32+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:35499" module=global caller=redis.go:47
t=2017-06-07T15:07:32+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:41612" module=global caller=redis.go:47
t=2017-06-07T15:07:32+0000 lvl=dbug msg="received handshake request from incoming connection" module=global caller=server.go:140
t=2017-06-07T15:07:32+0000 lvl=dbug msg="incoming connection checks out with version: 1.1.0-alpha" module=global caller=server.go:150
t=2017-06-07T15:07:32+0000 lvl=dbug msg="create vdisk with expectedSequence:0" module=global caller=vdisk.go:148
t=2017-06-07T15:07:32+0000 lvl=dbug msg="handshake phase successfully completed" module=global caller=server.go:175
t=2017-06-07T15:07:32+0000 lvl=dbug msg="replying handshake with status: OK" module=global caller=server.go:202
panic: test timed out after 5m0s

goroutine 81 [running]:
testing.startAlarm.func1()
	/home/travis/.gimme/versions/go1.8.linux.amd64/src/testing/testing.go:1023 +0xf9
created by time.goFunc
	/home/travis/.gimme/versions/go1.8.linux.amd64/src/time/sleep.go:170 +0x44

goroutine 1 [chan receive, 4 minutes]:
testing.(*T).Run(0xc4202dad00, 0xd23127, 0xe, 0xd41e08, 0xc420175d20)
	/home/travis/.gimme/versions/go1.8.linux.amd64/src/testing/testing.go:698 +0x2f4
[...]

PR #255, opened 2017-06-08
integrate hotreloader into tlog's redis pool

### LINKED ISSUE #177: tlogserver: support SIGHUP reloading for config hot reloading
For the solution of issue #170, the config is only loaded at startup time. For long running tlogservers however it would also require it to support hot-reloading (eg at SIGHUP).

GlenDC:
Fixes #177.

### DIFF
--- a/tlog/redispool.go
+++ b/tlog/redispool.go
@@ -78,13 +78,13 @@ func StaticRedisPoolFactory(requiredDataServerCount int, storageServers []config
 // such that at creation of a RedisPool, it is validated that the
 // storage cluster in question has sufficient data servers available.
 func ConfigRedisPoolFactory(requiredDataServerCount int, configPath string) (RedisPoolFactory, error) {
-	cfg, err := config.ReadConfig(configPath)
+	reloader, err := config.NewHotReloader(configPath)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create RedisPoolFactory: %s", err.Error())
 	}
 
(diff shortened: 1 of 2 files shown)


GlenDC:
Travis failed because of the same reason as #252, restarted the specific `TIMEOUT=5m make testcgo` jobs, should make it pass.

GlenDC:
All test pass, ready for review @iwanbk :)

iwanbk:
Does this change affect the tlog's flusher?

GlenDC:
In a way, but not in a way that you should be able to notice. It simply reloads the config, when receiving a SIGHUP, and applies that to the redisPools which are active.

For long running servers, this allows for data to migrate to different storage clusters, and let the server know about it.

But it shouldn't really affect you, no.

iwanbk:
> In a way, but not in a way that you should be able to notice. It simply reloads the config, when receiving a SIGHUP, and applies that to the redisPools which are active.

Oh, i thought we need to do something to the tlog's flusher, like:
- restart flusher
- re-assign the flusher's pool  variable because it is not pointer

GlenDC:
Nope, because the reloading happens, within that `tlog.redisPool`, IFF we use the config-based redis pool. If you for example use the `-memorystorage` or `tlogrpc` flags, you would use a static redisPool instead. But in production it is assumed that the config version is used (`-config` flag, which also has a default=`./config.yml`).

iwanbk:
cool

PR #262, opened 2017-06-09
Make the tlogclient's blockbuffer scalable.

### LINKED ISSUE #250: tlogclient : handle slow user's receiver
related to #247

iwanbk:
Closes #250 

In previous version, it scans the whole Go map to find the timed out blocks.
In this version, it added an array of sequence numbers ordered by it's timeout.
Which can be used by blockbuffer to find the timed out blocks without scanning the whole map.

### DIFF
--- a/tlog/tlogclient/blockbuffer/buffer.go
+++ b/tlog/tlogclient/blockbuffer/buffer.go
@@ -1,6 +1,7 @@
 package blockbuffer
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"math"
@@ -19,6 +20,7 @@ var (
 type Buffer struct {
 	lock            sync.Mutex
 	entries         map[uint64]*entry
+	seqToTimeout    []uint64
 	readyChan       chan *schema.TlogBlock
 	timeout         time.Duration
 	maxRetry        int
@@ -69,6 +71,11 @@ func (b *Buffer) Add(block *schema.TlogBlock) {
 
 	seq := block.Sequence()
 
+	// add to array of ordered timeout
+	b.seqToTimeout = append(b.seqToTimeout, seq)
+
+	// only need to update the timeout
+	// if already exist in the buffer
 	ent, exist := b.entries[seq]
 	if exist {
 		ent.update(b.timeout, 1)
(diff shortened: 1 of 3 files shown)


tlog/tlogclient/blockbuffer/buffer.go
```diff
@@ -92,33 +99,59 @@ func (b *Buffer) Delete(seq uint64) {
...
+	}
+
+	var i int
+	var seq uint64
+
+	defer func() {
+		b.seqToTimeout = b.seqToTimeout[i+1:] // truncate processed blocks
+	}()
 
-			// extend the timeout value, to prevent it being timed out
```
GlenDC: If you go down this road, I think you should simply limit the slice upfront, as in:

```go
seqToTimeout = b.seqToTimeout
if len(seqToTimeout) > 1000 { // be nice with others by not taking all the cpu
  seqToTimeout = seqToTimeout[:1000]
}
for _, seq := range seqToTimeout {
   // if not exist anymore in the map
   // ...
```

tlog/tlogclient/blockbuffer/buffer.go
```diff
@@ -92,33 +99,59 @@ func (b *Buffer) Delete(seq uint64) {
...
+func (b *Buffer) TimedOut(ctx context.Context) <-chan *schema.TlogBlock {
 	go func() {
 		for {
-			time.Sleep(time.Second)
-			ents := b.getTimedOut()
-			for _, ent := range ents {
+			select {
+			case <-ctx.Done():
+				return
+			default:
```
GlenDC: Instead of going via an intermediate slice `seqToTimeout`, wouldn't it be better to go for a channel (of sequences) directly, this way you could just wait here for a sequence to be available, instead of having to randomly sleep and wait for data to be available within the slice?

tlog/tlogclient/blockbuffer/buffer.go
iwanbk: Because this func only want to take one timed out blocks.
We can of course modify it to take array of timed out blocks

tlog/tlogclient/blockbuffer/buffer.go
iwanbk: I've tried to use channel, the problem with channel is that it is blocking after it reach it's max size and then create deadlock.

Some committed code can be seen at https://travis-ci.org/zero-os/0-Disk/builds/240691107. More code still exist at my PC.

Maybe i did some stupid mistake that make it deadlock

tlog/tlogclient/blockbuffer/buffer.go
GlenDC: What? What I'm proposing is exactly what you're doing. The only thing it improves is that you don't require the `if i == 1000 { return nil }` check every iteration.

iwanbk: And what to do when i want to return before i = 1000? Prepend the slice back?

Sorry, i don't understand :). Feel free to send a PR

GlenDC: You don't have to prepend the slice back, when you make a new slice, you don't make a copy, still shares the same memory, here is a diff of what I mean:

```diff
diff --git a/tlog/tlogclient/blockbuffer/buffer.go b/tlog/tlogclient/blockbuffer/buffer.go
index 5a2239a..aff0457 100644
--- a/tlog/tlogclient/blockbuffer/buffer.go
+++ b/tlog/tlogclient/blockbuffer/buffer.go
@@ -115,11 +115,12 @@ func (b *Buffer) getOne() *entry {
 		b.seqToTimeout = b.seqToTimeout[i+1:] // truncate processed blocks
 	}()
 
-	for i, seq = range b.seqToTimeout {
-		if i == 1000 { // be nice with others by not taking all the cpu
-			return nil
-		}
+	seqToTimeout := b.seqToTimeout
+	if len(seqToTimeout) > 1000 { // be nice with others by not taking all the cpu
+		seqToTimeout = seqToTimeout[:1000]
+	}
 
+	for i, seq = range seqToTimeout {
 		// if not exist anymore in the map
 		// it means the blocks already delivered
 		ent, ok := b.entries[seq]
@@ -129,6 +130,7 @@ func (b *Buffer) getOne() *entry {
 
 		return ent
 	}
+
 	return nil
 }
```

As far as I know, that does the same as what you do, but without having to check every for-loop iteration if `i == 1000`. I think it's a bit too small for a single PR, haha.

iwanbk: Thanks for the diff, i can see it now.
Just realized that my prepend question was a bit stupid.

It seems to work, feel free to make the needed changes or i can do it tomorrow.

Issue #256, opened 2017-06-08
tlogclient : add resend and reconnect to ForceFlushAtSeq


iwanbk:
- [x] reconnect (#263)
- [ ] resend

iwanbk:
The reconnect feature added at #263,

iwanbk:
The resend feature need code addition to the blockbuffer package, which only designed to be used for block resend.

I think the better way is to make all commands to be in separate tcp connection or  use capnp rpc.
It has the advantage to not mix command with data.

iwanbk:
Can't find a good solution for `resend` without add too much complexity to the code.
I think the best solution are:
- update the docs, states that user need to do the retry
- update nbdserver code to do the retry

GlenDC:
That would just move the complexity rather than preventing it, no? We still write the nbdserver code, so we would simply be writing the complexity in the nbdserver code, where I'm not sure this resend code really belongs.

iwanbk:
> That would just move the complexity rather than preventing it, no

The complexity is different, in a way that i can't solve it in a clean way in tlog client.

But yeah, your other points are right.

PR #263, opened 2017-06-09
Generic reconnect

iwanbk:
Part of #256.
Add reconnect to `ForceFlushAtSeq`

It also add generic method to do the reconnect

### DIFF
--- a/tlog/tlogclient/reconnect_test.go
+++ b/tlog/tlogclient/reconnect_test.go
@@ -91,6 +91,49 @@ func TestReconnectFromRead(t *testing.T) {
 	}
 }
 
+func (c *Client) forceFlushAtSeq(seq uint64) error {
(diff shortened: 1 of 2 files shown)


tlog/tlogclient/reconnect_test.go
```diff
@@ -91,6 +91,49 @@ func TestReconnectFromRead(t *testing.T) {
 	}
 }
 
+func (c *Client) forceFlushAtSeq(seq uint64) error {
```
GlenDC: Why can the reconnect_test not have the public API (and thus have the reconnect logic)?

iwanbk: This func is used by `TestReconnectFromRead` (test reconnect from reader/receiver).
If we use the public API (which support reconnect),  the reconnect might happen from the sender.

Issue #38, opened 2017-03-31
Failover to slave ardb's if the main ardb is failing


robvanmieghem:
- [x] tlog : add nbdserver's ardb slave config to server (#270)
- [x] tlog : also flush (using tlogreplay) to ardb slave in async way (#270)
- [x] tlog : on (re) start, make sure to keep ardb slave synced (#282)
- [x] tlog : add command to sync the ardb slave (#277 )
- [x] tlog : slave syncer must be able to self heal in case of error (#282)
- [x] nbdserver : detect failed master (#278)
- [x] nbdserver : ask tlogserver to sync the slave, need to pause/reject all operations during this sync process (#278)
- [x] nbdserver : switch to ardb slave, may need to restart vdisk or clean all of it's data/cache (#278)
- [x] nbdserver : notify the orchestrator about this failover (moved to #284)
- [x] slave syncer : add hot reload support (#25)

Redis replication is asynchronous, leaving a window for dataloss. I don't want to introduce paxos (or variation upon) complexity so a scenario in which the tlog server is integrated to handle this window seems like a better approach.

GlenDC:
@robvanmieghem I want to (try to) help you with thinking about a solution for this issue. I was wondering if you could give some more information on when, where and why we use this feature?

Also are the slaves read-only or do they write as well? Could you also tell more about how the replication setup is configured at the moment?

GlenDC:
I don't have too many details about our configuration and setup, so I can't elaborate too much yet. I do however think that we'll never be able to have strong consistency with a 100% _redis_-backed solution. Still, _redis_ has a couple of tricks and tools that might be able to assist us...

The [`WAIT`](https://redis.io/commands/wait) _redis_ command can make the client-side synchronous, which would solve the first half of _strong consistency_.

The second property could be partly solved, using [`redis-sentinel`](https://redis.io/topics/sentinel). It could help us with monitoring, notifying, discovery and even auto-recovering in case of failure. Auto-recovering in the sense that it can promote a slave to master in case the master fails. Also if one slave fails it shouldn't be an issue as long as we have enough other slaves. If a master and slave fails, it should still not be an issue (I think), as long as we still have a couple of other slaves available.

robvanmieghem:
As discussed on Telegram, solving it through normal master/slave replication will be hard to do. Better to include the tlog server in the architecture and let that one fill the slaves.

iwanbk:
[...]

PR #278, opened 2017-06-15
nbdserver : failover to slave.

### LINKED ISSUE #276: nbdserver : switch to ardb slave
Part of #38

@GlenDC  How is the best way for the nbdserver to switch to ardb slave?
(assumed that slave is ready to be used)

I think nbdserver can simply re-create the redis pool, to point to the slave. (need to add config for this)

iwanbk:
Fixes #276 as part of #38 

It adds support for ndbserver to switch to ardb slave in case of failing ardb master.

Note that this PR  only deals with nbdserver part.
The tlogserver and client already partially done in previous PRs.

Summary of changes:
- Add `SlaveStorageCluster` to the `0-Disk/config.Config`
- add ConfigPath to the nbd backend factory and tlogStorage
  We need it in order to tlogStorage be able to rewrite the config
- in case of ardb/storage operation failed, it do these:
	- make sure we have slave storage cluster
	- wait for slave to be fully synced by tlog
	- override the config, set the slave as master
	- hot reload the config

Limitations:
- ~~the failing detection currently only happens at `flushCachedContent`, other places could be added easily~~ It doesn't retry or reconnect in case of failed ardb operation, because i think it should be done by storage module
- it needs to wait for the hot reload to be finished, it currently do this by sleeping for 1 second, let me know if there is better way to do this
- orchestrator part is not included in this PR
- ~~test code is super mess~~

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -335,15 +335,16 @@ type StorageServerConfig struct {
 
 // VdiskConfig defines the config for a vdisk
 type VdiskConfig struct {
-	BlockSize          uint64    `yaml:"blockSize" valid:"optional"`
-	ReadOnly           bool      `yaml:"readOnly" valid:"optional"`
-	Size               uint64    `yaml:"size" valid:"optional"`
-	StorageCluster     string    `yaml:"storageCluster" valid:"optional"`
-	RootStorageCluster string    `yaml:"rootStorageCluster" valid:"optional"`
-	RootVdiskID        string    `yaml:"rootVdiskID" valid:"optional"`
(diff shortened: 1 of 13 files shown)


iwanbk:
@GlenDC this PR is not ready, but i need you to validate the basic idea.

It already works in unit test, it simulates failing master by closing the master's redis pool

iwanbk:
> the failing detection currently only happens at flushCachedContent

It doesn't do retry or reconnect in case of failed ardb operation, because i think it should be done by storage module

iwanbk:
update:
- docs
- some improvement to the tlogserver side

This PR is ready now

PR #290, opened 2017-06-22
Tlogclient multiple server

iwanbk:
Handle these cases (with unit tests):
- master crashed,  all received blocks already flushed to storage
- master crashed,  there are unflushed blocks
- master not crash, but failed to flush


iwanbk:
The client now need array of tlog server addresses instead of single address.
Client is going to use first address and then move to next addresses if the first address is failed.

PR #293, opened 2017-06-23
Fix hot reload data race.

iwanbk:
Fixes #293

Simplify the locking by only use one mutex instead of two.
Two mutexs might give very small performance gain on not critical path.
But it makes the locking mechanism too complex and hard to maintain, and lead to this issue

### DIFF
--- a/config/hotreload.go
+++ b/config/hotreload.go
@@ -68,7 +68,6 @@ type defaultHotReloader struct {
 	cfg          *Config
 	fullReloadCh chan struct{}
 	done         chan struct{}
-	handlerMux   sync.Mutex
 	mux          sync.RWMutex
 }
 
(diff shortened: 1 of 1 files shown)


config/hotreload.go
```diff
@@ -210,10 +208,11 @@ func (hr *defaultHotReloader) reloader(ctx context.Context) {
 // and sends the specific vdisk cluster configs to each registered notifier
 // using the freshly cached configs
 func (hr *defaultHotReloader) reloadVdiskClusterConfigs() error {
+	hr.mux.Lock()
+	defer hr.mux.Unlock()
+
 	// load config
 	err := func() error {
```
GlenDC: The only reason why this was wrapped in a closure is so that we could defer this unlock, as you removed it, the added complexity of this closure is no longer needed, and thus you could/should have removed this as well when working on this PR.

PR #295, opened 2017-06-23
Feat slavesync hot reload

iwanbk:
Part of #38 

Hot reload the tlog slave syncer

### DIFF
--- a/tlog/tlogserver/server/vdisk.go
+++ b/tlog/tlogserver/server/vdisk.go
@@ -3,11 +3,15 @@ package server
 import (
 	"context"
 	"net"
+	"os"
+	"os/signal"
 	"sync"
+	"syscall"
 	"time"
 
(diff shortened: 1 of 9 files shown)


tlog/tlogserver/server/vdisk.go
```diff
@@ -471,3 +474,86 @@ func tlogBlockComparator(a, b interface{}) int {
 		return 0
 	}
 }
+
+// handle SIGHUP
+func (vd *vdisk) handleSighup(ctx context.Context) {
+	sigs := make(chan os.Signal)
```
GlenDC: Why handle the SIGHUP signal here directly, instead of using the config.HotReloader? Seems like you could just use that one.

GlenDC: Nevermind, your comment https://github.com/zero-os/0-Disk/issues/38#issuecomment-310306307 seems to answer it.

PR #312, opened 2017-07-07
add semideduped storage type

### LINKED ISSUE #298: Space-Optimized Storage Type (deduped/nondeduped hybrid)
At the time of writing the `boot` vdisk type is a pure `deduped` storage. This is great and intended as it means we can dedupe all blocks, of which a lot exist for an OS image. However it has the unintended effect that the after-install (user-specific) data is also being deduped, while this content is most likely unique and used by that user only. Microsoft's Windows is an extreme example of this, as it is constantly updating its registers. And while these registers are on a relatively few amount of shards, it produces a lot of different block content versions, and thus spamming the deduped storage full of data.

Because of this, a new hybrid storage is proposed, which would be a combination of a deduped and nondeduped storage.

The image (and thus template) data would still (possibly) be read and stored into a deduped storage. But all write-like operations triggered by the user (nbdserver) would write to a nondeduped storage.

This would result in the following operation flow:

```go
// NOTES:
//   + The nondeduped storage would be created without template support;
//   + The deduped storage would be created with template support;
//   + Both the deduped and nondeduped storage would write to the same data storage servers;
type bootStorage struct

// nondedupedStorage.Set
(bs *bootStorage) Set(...) (...)

// bs.Get -> Merge Content -> bs.Set
(bs *bootStorage) Merge(...) (...)

// nondedupedStorage.Get -> OK
//    |
//    --> FAIL --> dedupedStorage.Get
(bs *bootStorage) Get(...) (...)

// nondedupedStorage.Delete && dedupedStorage.Delete
(bs *bootStorage) Delete(...) (...)

// nondedupedStorage.Flush && dedupedStorage.Flush
(bs *bootStorage) Flush(...) (...)

// nondedupedStorage.Close && dedupedStorage.Close
(bs *bootStorage) Close(...) (...)

// nondedupedStorage.GoBackground && dedupedStorage.GoBackground
(bs *bootStorage) GoBackground(...) (...)
```

### LINKED ISSUE #326: Fix broken zeroctl list vdisks command (remove prefixes)
Since PR #312, no hashmap is anymore _just_ identified by a vdisk. Instead we now add prefixes (`nondedup`, `lba`), as to not get any naming conflicts in scenarios where different kind of such hash maps are stored on the same storage servers for the same vdisk (which is the case for semi deduped storage).

This is all great and such, but we should:

+ Have these prefixes available as public constants;
+ Make sure to remove in each found hashmap key, such prefixes, if found;

GlenDC:
+ fixes #298;
+ fixes #326;

**warning**: this does contain breaking changes. As the storage key for LBA data and nondeduped data are now prefixed, such as to prevent name collisions between the two, when stored on the same storage server. As we're still in alpha, this shouldn't matter though, as it's not the first time we introduce breaking changes in an alpha release.

### DIFF
--- a/nbdserver/ardb/semideduped.go
+++ b/nbdserver/ardb/semideduped.go
@@ -5,16 +5,26 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/siddontang/go/log"
+	"github.com/garyburd/redigo/redis"
+	"github.com/zero-os/0-Disk/log"
 	"github.com/zero-os/0-Disk/nbdserver/lba"
 )
 
-func newSemiDedupedStorage(vdiskID string, blockSize int64, provider redisConnectionProvider, vlba *lba.LBA) backendStorage {
-	return &semiDedupedStorage{
+func newSemiDedupedStorage(vdiskID string, blockSize int64, provider redisConnProvider, vlba *lba.LBA) backendStorage {
+	storage := &semiDedupedStorage{
 		templateStorage: newDedupedStorage(vdiskID, blockSize, provider, true, vlba),
 		userStorage:     newNonDedupedStorage(vdiskID, "", blockSize, false, provider),
+		vdiskID:         vdiskID,
 		blockSize:       blockSize,
+		provider:        provider,
 	}
+
+	err := storage.readBitMap()
+	if err != nil {
+		log.Debugf("couldn't read semi deduped storage %s's bitmap: %v", vdiskID, err)
+	}
+
+	return storage
 }
 
 // semiDedupedStorage is a backendStorage implementation,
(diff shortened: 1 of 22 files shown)


zaibon:
LGM, but before I merge. Is there any impact on the configuration file for a dedupe vdisk?
Is this new SemiDedupe type is by default use for any boot disk now ?

GlenDC:
> Is there any impact on the configuration file for a dedupe vdisk?
> Is this new SemiDedupe type is by default use for any boot disk now ?

No impact, the semi dedupe storage is used for any boot disk which also has a rootStorageCluster defined.

> I'll block this PR until @GlenDC can have a look at the comment : #298 (comment)

Already checked that very valid comment. Will work on it today, as to not block this PR too long.

GlenDC:
@zaibon could you re-review this PR? I added the bitmap to make the semi dedup storage more efficient. Please do not merge in yet, as I'll do a last review myself of the code in detail tomorrow, + some extra thorough (manual) testing of it all. But please could you simply validate the ideas and code as you can see? And please approve if it all looks good according to you.

Once you approve, and I can confirm that the code is good as it is (after testing tomorrow), I'll merge it into master myself than tomorrow.

nbdserver/ardb/semideduped.go
```diff
@@ -29,12 +39,49 @@ type semiDedupedStorage struct {
...
+	// we know it has to be retreived from the user storage
+	//
+	// NOTE: for now this bit is never unset,
+	// as it is assumed that once data is overwritten (if it's overwritten at all),
+	// it is custom forever. It would be weird if suddenly out of the blue (after a delete operation for example),
+	// the template (original) data would be used once again,
+	// I don't think that's something a user would expect at all.
+	sds.userStorageBitMap.Set(int(blockIndex))
+
+	// delete content from dedup storage, as it's no longer needed there
```
zaibon: Why do we delete content from the deduped storage ? I'm not sure I get it.

Issue #310, opened 2017-07-07
broken tlogclient multiple server unit test


GlenDC:
A tlogclient unit test has failed now a couple of times on master code, log:

```
t=2017-07-07T01:42:14+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:33302" module=global caller=redis.go:47
t=2017-07-07T01:42:14+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:40954" module=global caller=redis.go:47
t=2017-07-07T01:42:14+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:45127" module=global caller=redis.go:47
t=2017-07-07T01:42:14+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:33602" module=global caller=redis.go:47
t=2017-07-07T01:42:14+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:37472" module=global caller=redis.go:47
t=2017-07-07T01:42:14+0000 lvl=info msg="embedded LedisDB Server ready and listening at 127.0.0.1:42434" module=global caller=redis.go:47
t=2017-07-07T01:42:14+0000 lvl=dbug msg="received handshake request from incoming connection" module=global caller=server.go:157
t=2017-07-07T01:42:14+0000 lvl=dbug msg="incoming connection checks out with version: 1.1.0-alpha" module=global caller=server.go:167
t=2017-07-07T01:42:14+0000 lvl=dbug msg="create vdisk with expectedSequence:0" module=global caller=vdisk_mgr.go:65
t=2017-07-07T01:42:14+0000 lvl=dbug msg="handshake phase successfully completed" module=global caller=server.go:192
t=2017-07-07T01:42:14+0000 lvl=dbug msg="replying handshake with status: OK" module=global caller=server.go:219
t=2017-07-07T01:42:18+0000 lvl=info msg="tlogclient : reconnect from read because of : EOF" module=global caller=client.go:336
FAIL
FAIL	github.com/zero-os/0-Disk/tlog/tlogclient	56.603s
```

GlenDC:
The cause of this fail seems to happen most commits I push to GitHub, so it shouldn't be too hard to reproduce this locally hopefully.

iwanbk:
is this the same error as https://travis-ci.org/zero-os/0-Disk/jobs/250433968? I can't reproduce it locally

[...]

PR #320, opened 2017-07-11
Fix tlog test multipleserver

iwanbk:
#310 fix attempt

- tlogclient always move to other server in any case of failures
- tlogserver's vdisk exit on failed flush
- better test logging by using the logger package inside the test. It is because test logger always printed after our logger package, which make the debugging to be hard.

It passed 9 travis builds :)

### DIFF
--- a/tlog/tlogclient/common_test.go
+++ b/tlog/tlogclient/common_test.go
@@ -25,6 +25,7 @@ func testClientWaitSeqFlushed(ctx context.Context, t *testing.T, respChan <-chan
 		select {
 		case <-time.After(20 * time.Second):
 			t.Fatal("testClientWaitSeqFlushed timeout")
+			return
 		case re := <-respChan:
 			if re.Err != nil {
 				t.Logf("recv err = %v", re.Err)
--- a/tlog/tlogserver/server/vdisk.go
+++ b/tlog/tlogserver/server/vdisk.go
@@ -43,6 +43,8 @@ type vdiskFlusherCmd struct {
 	respCh   chan error
 }
 
+type vdiskCleanupFunc func(vdiskID string)
+
 type vdisk struct {
 	vdiskID     string
 	lastHash    []byte              // current in memory last hash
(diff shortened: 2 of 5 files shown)


tlog/tlogclient/common_test.go
```diff
@@ -25,6 +25,7 @@ func testClientWaitSeqFlushed(ctx context.Context, t *testing.T, respChan <-chan
 		select {
 		case <-time.After(20 * time.Second):
 			t.Fatal("testClientWaitSeqFlushed timeout")
+			return
```
GlenDC: Is this return statement needed? it seems like Fatal exits no? At least https://golang.org/src/testing/testing.go#L522 seems like it's doing that.

tlog/tlogserver/server/vdisk.go
```diff
@@ -142,15 +144,28 @@ func newVdisk(ctx context.Context, aggMq *aggmq.MQ, vdiskID, configPath string,
 	if err := vd.manageSlaveSync(); err != nil {
 		return nil, err
 	}
+	ctx, cancelFunc := context.WithCancel(context.Background())
 
 	// run vdisk goroutines
-	go vd.runFlusher(ctx)
-	go vd.runBlockReceiver(ctx)
-	go vd.handleSighup(ctx)
+	go vd.runFlusher(parentCtx, ctx, cancelFunc)
```
GlenDC: Why give both contexes? Fairly sure you pass on the new `ctx` no? As far as I know this should trigger the `Done` whether that's triggered by the cancelFunc, or because of something with the `parentCtx`. Especially as you don't seem to be doing something special besides also checking if it's done.

tlog/tlogserver/server/vdisk.go
```diff
@@ -142,15 +144,28 @@ func newVdisk(ctx context.Context, aggMq *aggmq.MQ, vdiskID, configPath string,
...
+	go vd.cleanup(parentCtx, ctx, cleanup)
 
 	return vd, nil
 }
 
+// do all necessary cleanup for this vdisk
+func (vd *vdisk) cleanup(parentCtx, ctx context.Context, cleanup vdiskCleanupFunc) {
+	defer cleanup(vd.vdiskID)
+	select {
+	case <-parentCtx.Done():
```
GlenDC: See my previous comment, but I think a snippet like this and the other methods where you do it, could be replaced by:

```
case <-ctx.Done():
    return
```

Making it not required (or even recommended) to pass both contexts to these functions.

iwanbk: thanks, you're right, fixed on #327

PR #330, opened 2017-07-12
Adds etcdconfig

### LINKED ISSUE #316: remove usage of config files, go to etcd
config files are against the stateless nature of our Zero-OS

we should do the following
- move all configuration settings to etcd
- each daemon (service) starts with list of etcd to work with & his own key
    - $pathToServiceBinary $ip1,$ip2,$ip3 $servicename $instancename
    - key generated = $servicename_$instancename (in relation to ays)
- use watching of keys to see when config changes
    - https://coreos.com/etcd/docs/latest/learning/api.html#watch-api
    - if it changes reload
    - an also use this to send some commands to a daemon e.g. have in toml file: cmds section
suggested requirements
- store config for each service as toml file
- run max 5 etcd daemons (ays manages the setup of these daemons), they are in 1 cluster

todo
- spec the required config formats per service e.g. nbdserver

chrisvdg:
+ adds support to load configs from either ETCD or file;
+ ETCD should be used in production;
+ refactor config completely to accommodate the ETCD support;
+ remove the outdated (non-std) context package (and its use) for now in all dependencies;
+ add updated config documentation;

closes #316


configV2/config.go
```diff
@@ -0,0 +1,270 @@
...
+	res, err := yaml.Marshal(cfg)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// deSerialize deserializes a subconfig from YAML 1.2 content (for now)
+// error returned if type was not of subconfig type, mashal error or fails to validate
+func (v *VdiskConfig) deSerialize(cfg interface{}, data []byte) (interface{}, error) {
```
GlenDC: please name this `deserialize`

configV2/config.go
```diff
@@ -72,19 +96,24 @@ type SlaveConfig struct {
...
 
-// serialize provided subconfig into YAML 1.2 content (for now)
+// serializeVdiskSubCfg serializes provided subconfig into YAML 1.2 content (for now)
 func serializeVdiskSubCfg(cfg interface{}) ([]byte, error) {
-	res, err := yaml.Marshal(cfg)
+	return marchalYAML(cfg)
+}
+
+// marchalYAML is a wrapper function for yaml.Marchal
+func marchalYAML(data interface{}) ([]byte, error) {
```
GlenDC: `marshalYAML`*,
and why are you wrapping it?

configV2/config.go
```diff
@@ -141,18 +170,117 @@ func deSerializeVdiskSubCfg(cfg interface{}, data []byte) (interface{}, error) {
...
 func validateVdiskSubCfg(cfg interface{}) error {
-	// TODO: Add validation implementation
+	_, err := valid.ValidateStruct(cfg)
+	if err != nil {
+		return err
+	}
+
 	switch cfg.(type) {
 	case BaseConfig:
+		err := validateBase(cfg.(BaseConfig))
```
GlenDC: See, this is the reason why I'm not sure that working with a wrapped vdisk config is a good idea. We can't just validate all separate structs, as it wouldn't be uncommon for only one or a couple of them to be defined, not all of them. I think you should keep those structs separate, and validate them on serialization and deserialization of the structs.

configV2/config_test_data.go
```diff
@@ -30,7 +33,123 @@ vdisks:
...
+        dataStorage: 
+          - address: 192.168.58.149:2000
+            db: 4
+        metadataStorage:
+          address: 192.168.58.146:2001
+          db: 8
+  `,
+}
+
+// TOML
```
GlenDC: toml can be removed, let's focus on the real task here, and that is supporting ETCD as well as file confgs. So for now let's just go for YAML, as discussed in the meeting.

configV2/yaml.go
```diff
@@ -44,7 +48,7 @@ func (YAMLConfig) fromYAMLBytes(bytes []byte) (*Config, error) {
 
 	err = cfg.validate()
 	if err != nil {
-		return cfg, fmt.Errorf("Could not validate configuration: %s", err)
+		return cfg, fmt.Errorf("Invalidation configuration: %s", err)
```
GlenDC: errors shouldn't start with a capital according to golang standards, AFAIK. (nitpick)

configV2/config.go
chrisvdg: But some validation need other structs (baseconfig)?
e.g: there should be an ndb rootstorage when template server is supported for which I need to check the VdiskType from the base config.

Or could/should the validation for the other types also get the base config?

GlenDC: But when you automatically validate during deserialization of those specific structs (as needed), you'll get that validation as well.

That's why I was thinking that each struct has a constructor which just allows you to serialize from bytes, no matter where those bytes came from. Perhaps this is complicated when working with the file config, not sure.

What is more important is that you can simply give the constructor the `nbdConf` the `vdiskType` (on top of the bytes you might give it). The other structs don't need the type or any other info of the baseConf AFAIK.

Also I don't think the user should ever create those structs himself. I think that's more something that the Source Interface should done for the user.

So the user will always work with some Source interface I think, which allows you to get those structs via provided methods, and than on top of that, it also (optionally) provides hot reloading.

chrisvdg: Ok I'll try that,
Also going to put the (de)serialize and validate methodes to each struct instead of the interface switching I'm doing atm, which is how you want it right?

GlenDC: Yeah, I think that makes more sense.

Because either we split up the data structs, or we keep them together, but now you're trying to do both, which doesn't make much sense at all.

configV2/config.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+
+	return nil
+}
+
+// TemplateSupport returns whether or not
+// this vdisk supports a template (root) server,
+// to get the data in case the data isn't available on
+// the normal (local) storage cluster.
+func TemplateSupport(vtype VdiskType) bool {
```
GlenDC: I think it makes more sense to provide this as a method of `VdiskType`. Same goes for `TlogSupport`.

configV2/config.go
```diff
@@ -0,0 +1,442 @@
...
+// based on /docs/README.md#zero-os-0-disk
+const (
+	VdiskTypeBoot  = propDeduped | propPersistent | propTlogSupport | propTemplateSupport
+	VdiskTypeDB    = propPersistent | propTlogSupport | propTemplateSupport
+	VdiskTypeCache = propPersistent
+	VdiskTypeTmp   = propTemporary
+)
+
+// GetStorageType returns the type of storage this vdisk uses
+func (t VdiskType) GetStorageType() StorageType {
```
GlenDC: Correct me if I'm wrong, but I think it's more Go Idiomatic to call a function like this `StorageType`. That's also how it used to be called.

configV2/config.go
```diff
@@ -0,0 +1,442 @@
...
+	StorageNondeduped
+)
+
+// UInt8 returns the storage type as an uint8 value
+func (st StorageType) UInt8() uint8 {
+	return uint8(st)
+}
+
+// String returns the name of the storage type
+func (st StorageType) String() string {
```
GlenDC: You need to sync with master, as the storage type has changed a bit since you started working on this feature.

configV2/config.go
```diff
@@ -0,0 +1,442 @@
...
+		return fmt.Errorf("baseconfig has invalid type: %s", err.Error())
+	}
+
+	return nil
+}
+
+// NDBConfig represents an ndb storage configuration
+type NDBConfig struct {
+	StorageCluster     StorageClusterConfig `yaml:"storageCluster" valid:"optional"`
+	RootStorageCluster StorageClusterConfig `yaml:"rootStorageCluster" valid:"optional"`
```
GlenDC: you can call this `templateStorageCluster`. From now on use the word `template` where you previously used `root`. This is to stay consist with naming, as to fix the mess where we can't properly communicate as a team because we use different names for the same concept.

configV2/config.go
```diff
@@ -0,0 +1,442 @@
...
+	RootVdiskID        string               `yaml:"rootVdiskID" valid:"optional"`
```
GlenDC: you can call this `templateVdiskID`. See above.

configV2/config.go
```diff
@@ -0,0 +1,442 @@
...
+	_, err := valid.ValidateStruct(slave)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// TemplateSupport returns whether or not
+// this vdisk supports a template (root) server,
```
GlenDC: `(root)` can be removed here.

configV2/config_test_data.go
```diff
@@ -0,0 +1,181 @@
...
+    tlogConfig:
+      tlogStorageCluster:
+        dataStorage: 
+          - address: 192.168.58.149:2000
+            db: 4
+        metadataStorage:
+          address: 192.168.58.146:2001
+          db: 8
+  `,
+	// bad root storage dail address
```
GlenDC: `dial`*

configV2/toml.go
```diff
@@ -0,0 +1 @@
+package configV2
```
GlenDC: can be removed

GlenDC:
Also make sure to rebase master into your branch. Especially merged PR #333 will fix one of your failing travis CI tests.

configV2/config.go
```diff
@@ -2,13 +2,22 @@ package configV2
...
 	"fmt"
-	"strconv"
 
 	valid "github.com/asaskevich/govalidator"
 	"github.com/go-yaml/yaml"
 	"github.com/siddontang/go/log"
 )
 
+// ConfigSource specifies a config source interface
+type ConfigSource interface {
```
GlenDC: Where are the methods to get the configs? The config implementations shouldn't have any public fields, as we're not working with those, we're always working with the interface instead.

configV2/config.go
```diff
@@ -2,13 +2,22 @@ package configV2
 
 import (
 	"fmt"
-	"strconv"
 
 	valid "github.com/asaskevich/govalidator"
 	"github.com/go-yaml/yaml"
 	"github.com/siddontang/go/log"
```
GlenDC: you're using the wrong logger, it's "github.com/zero-os/0-Disk/log"

configV2/config.go
```diff
@@ -2,13 +2,22 @@ package configV2
...
 )
 
+// ConfigSource specifies a config source interface
+type ConfigSource interface {
+	Close() error               // closes source connection and goroutines if present
+	SetBase(BaseConfig) error   // sets a new base config and writes it to the source
+	SetNDB(NDBConfig) error     // sets a new NDB config and writes it to the source
+	SetTlog(TlogConfig) error   // sets a tlog base config and writes it to the source
+	SetSlave(SlaveConfig) error // sets a slave base config and writes it to the source
+	Validate() error            // Validates all of the current configs
```
GlenDC: Always keep an API as simple as possible, please remove this validate method. Validation should be done implicitly by the source when (de)serializing a config.

configV2/config.go
```diff
@@ -2,13 +2,22 @@ package configV2
...
+}
+
 // Config for the zerodisk backends
 type Config struct {
```
GlenDC: Why is this still here?

configV2/config.go
```diff
@@ -369,18 +378,18 @@ func (t VdiskType) StorageType() StorageType {
 }
 
 // Validate this vdisk type
-func (t VdiskType) Validate() error {
-	switch t {
+func (vtype VdiskType) Validate() error {
```
GlenDC: I don't think `vtype` is much better then `t`, why the change? All functions are pretty short, so it's pretty obvious to see that it is of type `VdiskType`.

configV2/config.go
chrisvdg: replaced

configV2/config.go
chrisvdg: They're added

configV2/config.go
chrisvdg: Consistency with the code I've added, didn't occur to me at first t was used for VdiskType

configV2/etcd.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+)
+
+// ETCDConfig represents a config using ETCD as source
+type ETCDConfig struct {
```
GlenDC: should be a private struct

configV2/etcd.go
```diff
@@ -0,0 +1,69 @@
...
+	NDB       NDBConfig
+	Tlog      TlogConfig
+	Slave     SlaveConfig
+	endpoints []string
+	cli       *clientv3.Client
+	closeFunc func() error
+}
+
+// NewETCDConfig returns a new ETCDConfig from provided endpoints
+func NewETCDConfig(vdiskID string, endpoints []string) (*ETCDConfig, error) {
```
GlenDC: should return `(ConfigSource, error)`

configV2/etcd.go
```diff
@@ -0,0 +1,69 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cfg *ETCDConfig) baseKey() string {
+	return fmt.Sprintf("%s:%s", cfg.VdiskID, "conf")
```
GlenDC: If we store plan to keep a source static in so much that we give the vdiskID for any source we create, it would be better to store these 4 keys (`baseKey`, `nbdKey`, `tlogKey`, `slaveKey`) at construction time. `fmt.Sprintf` isn't a cheap function to perform at all, so if we only have to compute each key once, that would save us time, even though I know we won't require these keys _that_ much, it would still prevent us from "the dead of a thousand cuts".

configV2/yaml.go
```diff
@@ -0,0 +1,190 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/go-yaml/yaml"
+)
+
+// YAMLConfig represents a config using a YAML file as source
+type YAMLConfig struct {
```
GlenDC: should be a private type

configV2/yaml.go
```diff
@@ -0,0 +1,190 @@
...
+
+type configWrapper struct {
+	Base  BaseConfig  `yaml:"baseConfig" valid:"required"`
+	NDB   NDBConfig   `yaml:"ndbConfig" valid:"required"`
+	Tlog  TlogConfig  `yaml:"tlogConfig" valid:"required"`
+	Slave SlaveConfig `yaml:"slaveConfig" valid:"optional"`
+}
+
+// NewYAMLConfig returns a new YAMLConfig from provided file path
+func NewYAMLConfig(vdiskID string, path string) (*YAMLConfig, error) {
```
GlenDC: Should return `(ConfigSource, error)`

configV2/yaml.go
```diff
@@ -0,0 +1,190 @@
...
+	cfg.ndb = ndb
+
+	// write to file
+	return cfg.writeToSource()
+}
+
+// SetTlog implements ConfigSource.SetTlog
+// Sets a new tolog config and writes it to the source
+func (cfg *YAMLConfig) SetTlog(tlog TlogConfig) error {
+	cfg.tlog = tlog
```
GlenDC: make sure to validate `tlog` first

configV2/yaml.go
```diff
@@ -0,0 +1,190 @@
...
+	cfg.base = base
+
+	// write to file
+	return cfg.writeToSource()
+}
+
+// SetNDB implements ConfigSource.SetNDB
+// Sets a new ndb config and writes it to the source
+func (cfg *YAMLConfig) SetNDB(ndb NDBConfig) error {
+	cfg.ndb = ndb
```
GlenDC: make sure to validate `nbd` first

configV2/yaml.go
```diff
@@ -0,0 +1,190 @@
...
+
+// Slave implements ConfigSource.Slave
+func (cfg *YAMLConfig) Slave() SlaveConfig {
+	return cfg.slave
+}
+
+// SetBase implements ConfigSource.SetBase
+// Sets a new base config and writes it to the source
+func (cfg *YAMLConfig) SetBase(base BaseConfig) error {
+	cfg.base = base
```
GlenDC: make sure to validate `base` first

configV2/yaml.go
```diff
@@ -0,0 +1,190 @@
...
+		return cfg, err
+	}
+
+	// set up buffer to be able to set private fields
+	cfg.base = buf.Base
+	cfg.ndb = buf.NDB
+	cfg.tlog = buf.Tlog
+	cfg.slave = buf.Slave
+
+	err = cfg.validate()
```
GlenDC: not sure if you need a method for this, really you only need the `validate` internal method here, so you could just as well do it directly here.

configV2/etcd.go
```diff
@@ -0,0 +1,69 @@
...
+	cfg.cli = cli
+
+	// fetch all subconfigs
+
+	return cfg, nil
+}
+
+// Close implements ConfigSource.Close
+func (cfg *ETCDConfig) Close() error {
+	if cfg.cli != nil {
```
GlenDC: you would save an indention if you would simply write:

```
if cfg.cli == nil {
   return nil
}
```

configV2/etcd.go
```diff
@@ -0,0 +1,112 @@
...
+	"time"
+
+	"github.com/coreos/etcd/clientv3"
+)
+
+// etcdConfig represents a config using ETCD as source
+type etcdConfig struct {
+	vdiskID   string
+	base      BaseConfig
+	ndb       NDBConfig
```
GlenDC: `nbd`, `tlog` and `slave` subconfigs should be stored by pointer instead of by value, as they are all 3 optional. So this way you know which ones you were able to load. Which will also help you in only validating those configs that you have.

configV2/yaml.go
```diff
@@ -0,0 +1,198 @@
...
+	"os"
+
+	"github.com/go-yaml/yaml"
+)
+
+// yamlConfig represents a config using a YAML file as source
+type yamlConfig struct {
+	vdiskID   string
+	base      BaseConfig
+	ndb       NDBConfig
```
GlenDC: `nbd`, `tlog` and `slave` subconfigs should be stored by pointer instead of by value, as they are all 3 optional. So this way you know which ones you were able to load. Which will also help you in only validating those configs that you have.

configV2/etcd.go
```diff
@@ -0,0 +1,112 @@
...
+	tlog      TlogConfig
+	slave     SlaveConfig
+	baseKey   string
+	ndbKey    string
+	tlogKey   string
+	slaveKey  string
+	endpoints []string
+	cli       *clientv3.Client
+	closeFunc func() error
```
GlenDC: what is this for? I think you can close it, seems kind-of useless, and not even used.

configV2/etcd.go
```diff
@@ -0,0 +1,112 @@
...
+	base      BaseConfig
+	ndb       NDBConfig
+	tlog      TlogConfig
+	slave     SlaveConfig
+	baseKey   string
+	ndbKey    string
+	tlogKey   string
+	slaveKey  string
+	endpoints []string
+	cli       *clientv3.Client
```
GlenDC: better to provide your own interface for the ETCD client, this will allow you to also do some stubbed unit testing

configV2/yaml.go
```diff
@@ -0,0 +1,198 @@
...
+	tlog      TlogConfig
+	slave     SlaveConfig
+	path      string
+	closeFunc func() error
+}
+
+type configWrapper struct {
```
GlenDC: rename to yamlConfigFormat, as that's what really defines, and this will also avoid any unintended name collisions.

configV2/config.go
```diff
@@ -0,0 +1,462 @@
...
+import (
+	"fmt"
+
+	valid "github.com/asaskevich/govalidator"
+	"github.com/go-yaml/yaml"
+	"github.com/zero-os/0-Disk/log"
+)
+
+// ConfigSource specifies a config source interface
+type ConfigSource interface {
```
GlenDC: Nitpick, but I think I would prefer the following order, with some separation:

```go
type ConfigSource interface {
	Base() BaseConfig           // Returns the current base config
	NDB() NDBConfig             // Returns the current ndb config
	Tlog() TlogConfig           // Returns the current tlo
	Slave() SlaveConfig         // Returns the current Slave config

	SetBase(BaseConfig) error   // sets a new base config and writes it to the source
	SetNDB(NDBConfig) error     // sets a new NDB config and writes it to the source
	SetTlog(TlogConfig) error   // sets a tlog base config and writes it to the source
	SetSlave(SlaveConfig) error // sets a slave base config and writes it to

	Close() error               // closes source connection and goroutines if present
}
```

configV2/yaml.go
```diff
@@ -0,0 +1,198 @@
...
+	Base  BaseConfig  `yaml:"baseConfig" valid:"required"`
+	NDB   NDBConfig   `yaml:"ndbConfig" valid:"required"`
```
GlenDC: specify NBD, Tlog and Slave properties by pointer rather then by value, as these are optional.

configV2/config.go
```diff
@@ -0,0 +1,462 @@
...
+
+// NDBConfig represents an ndb storage configuration
+type NDBConfig struct {
+	StorageCluster         StorageClusterConfig `yaml:"storageCluster" valid:"optional"`
+	TemplateStorageCluster StorageClusterConfig `yaml:"templateStorageCluster" valid:"optional"`
+	TemplateVdiskID        string               `yaml:"templateVdiskID" valid:"optional"`
+}
+
+// NewNDBConfig creates a new NDBConfig from byte slice in YAML 1.2 format
+func newNDBConfig(data []byte, vID string, vtype VdiskType) (*NDBConfig, error) {
```
GlenDC: stay consistent in naming with the existing code base: `vID` -> `vdiskID`

also I would rename `vtype` -> `vdiskType`

configV2/config.go
```diff
@@ -0,0 +1,462 @@
...
+func (vtype VdiskType) TlogSupport() bool {
+	return vtype&propTlogSupport != 0
+}
+
+// TemplateSupport returns whether or not
+// this vdisk supports a template server,
+// to get the data in case the data isn't available on
+// the normal (local) storage cluster.
+func (vtype VdiskType) TemplateSupport(ndb NDBConfig) bool {
+	return vtype&propTemplateSupport != 0 ||
```
GlenDC: you can redefine/reduce the body as/to: `return vtype&propTemplateSupport != 0`, as the body you have now is still old code, that I forgot to update after disabling the `semiDedupedStorage` for the `boot` vdisk type

configV2/config.go
```diff
@@ -0,0 +1,462 @@
...
+func (base *BaseConfig) serialize() ([]byte, error) {
+	res, err := yaml.Marshal(base)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// deserialize tries to convert provided data into a baseConfig
+func (base *BaseConfig) deserialize(data []byte) error {
```
GlenDC: this should be removed, instead you should construct a new subconfig using the constructor.

configV2/config.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	base.validate()
+	if err != nil {
+		return nil, err
+	}
+	return base, nil
+}
+
+// serialize converts baseConfig in byte slice in YAML 1.2 format
+func (base *BaseConfig) serialize() ([]byte, error) {
```
GlenDC: Make it public, as it is required for people implementing their own ConfigSource in another package, and rename it to `ToBytes`, as that is more consistent with the rest of the 0-Disk codebase.

(this comment applies to all subconfigs)

configV2/etcd.go
chrisvdg: if hotreading would be set it could set this as it's closing function which would then be called when the config is closed, but will see if it's needed once I start with implementing hotreloading/watch

GlenDC: Fairly sure it's not required, it's definitely not required ever in the `yamlConfig` (which should be renamed to `fileConfig` as I commented earlier).

configV2/config.go
chrisvdg: is it just for readability or is there any other reason for this?

GlenDC: mix of readability and preference

configV2/etcd.go
chrisvdg: I don't really get this one

iwanbk: Try to answer, i hope this is what Glen thought.

Lets say we only use 2 funcs from `*clientv3.Client` : `First() bool`  and `Second() error`.

We create the interface

```go
type cliInterface interface {
   First() bool
   Second() error
}
```

The above struct become
```go
endpoints []string
cli       cliInterface
```

Then you can assign anything to `cli`, as long as it has those 2 funcs above

GlenDC: Yes that's what I mean indeed. @iwanbk explains the idea like I had it in mind, even though I would call it `etcdClient` or something like that. Than again, naming is hard, especially for all of us who are not English native.

configV2/etcd.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Close implements ConfigSource.Close
+func (cfg *etcdConfig) Close() error {
+
+	// close client if created
+	if cfg.cli == nil {
+		return nil
+	}
+	err := cfg.cli.Close()
```
GlenDC: You can simply do `return cfg.cli.close()`, as you are adding a conditional (= extra cost), without gaining any value of it.

configV2/etcd.go
```diff
@@ -0,0 +1,212 @@
...
+	//base
+	resp, err := cli.Get(context.TODO(), cfg.baseKey)
+	if err != nil {
+		return nil, fmt.Errorf("could not get base from ETCD: %v", err)
+	}
+	if len(resp.Kvs) < 1 {
+		return nil, fmt.Errorf("base config was not found on the ETCD server")
+	}
+	base, err := newBaseConfig(resp.Kvs[0].Value)
+	cfg.base = *base
```
GlenDC: You don't seem to handle the branch where creating `base` fails (thus where `err != nil`.

configV2/etcd.go
iwanbk: > even though I would call it etcdClient or something like that. Than again, naming is hard, especially for all of us who are not English native.

Actually i chose a bad name, so chris can choose his own

GlenDC: Ah ok. You do have to know, if you didn't already, that people tend to copy sample code even though it wasn't meant to be copied. A dangerous example of that is, when people give an example of a bad encryption code, and ignorant people copy that and use it in production :( Therefore I usually only add code to my comments if I'm OK with it being copied and used.

configV2/etcd.go
chrisvdg: Didn't know I could do that, C# habits

GlenDC: What are you talking about? You can do that in C# as well o.o

chrisvdg: Was taught to do null checks before closing streams etc, as it might throw a nullpointer exception

configV2/etcd.go
chrisvdg: Don't seem to be able to pull this one off, clientv3.Client has public fields and I get errors that it implements context from the standard library while my interface takes the context from Vendor.
Are there ways around this?

configV2/etcd.go
```diff
@@ -87,42 +101,156 @@ func NewETCDConfig(vdiskID string, endpoints []string) (ConfigSource, error) {
...
+	defer cfg.baseRWLock.RUnlock()
 	return cfg.base
 }
 
 // NBD implements ConfigSource.NBD
 func (cfg *etcdConfig) NBD() (*NBDConfig, error) {
+	cfg.nbdRWLock.RLock()
+	defer cfg.nbdRWLock.RUnlock()
+	if cfg.nbd == nil {
+		return nil, nil
```
GlenDC: This should return an error. Whoever is calling `NBD` is clearly assuming there is an NBD config available. In this case, no NBD config available, and thus that is an error. If you are going to start returning `nil` even when `error` is `nil`, you always need to do 2 checks, instead of 1. There are cases where this is desired, but this is not one of those. You can make a "constant" error `ErrConfigNotAvailable` for this one (or whatever the name convention is).

configV2/etcd.go
```diff
@@ -87,42 +101,156 @@ func NewETCDConfig(vdiskID string, endpoints []string) (ConfigSource, error) {
...
+	}
 	return cfg.nbd.Clone(), nil
 }
 
 // Tlog implements ConfigSource.Tlog
 func (cfg *etcdConfig) Tlog() (*TlogConfig, error) {
+	cfg.tlogRWLock.RLock()
+	defer cfg.tlogRWLock.RUnlock()
+	if cfg.tlog == nil {
+		return nil, nil
```
GlenDC: the same thing as I commented on ` func (cfg *etcdConfig) NBD()` applies here

configV2/etcd.go
```diff
@@ -87,42 +101,156 @@ func NewETCDConfig(vdiskID string, endpoints []string) (ConfigSource, error) {
...
+	}
 	return cfg.tlog.Clone(), nil
 }
 
 // Slave implements ConfigSource.Slave
 func (cfg *etcdConfig) Slave() (*SlaveConfig, error) {
+	cfg.slaveRWLock.RLock()
+	defer cfg.slaveRWLock.RUnlock()
+	if cfg.slave == nil {
+		return nil, nil
```
GlenDC: the same thing as I commented on ` func (cfg *etcdConfig) NBD()` applies here

GlenDC:
Can you rename `fileConfig.go` to `file_config.go` please, it's the convention that file names are written in snake case. Right now `fileConfig.go` is the only outlier in this codebase I'm aware of.

configV2/etcd.go
```diff
@@ -150,13 +280,17 @@ func (cfg *etcdConfig) SetBase(base BaseConfig) error {
...
 func (cfg *etcdConfig) SetNBD(nbd NBDConfig) error {
+	cfg.baseRWLock.RLock()
 	err := nbd.validate(cfg.vdiskID, cfg.base.Type)
+	cfg.baseRWLock.RUnlock()
 	if err != nil {
 		return fmt.Errorf("provided nbd config was not valid: %s", err)
 	}
 
 	// write to etcd
+	cfg.nbdRWLock.RLock()
```
GlenDC: Why are you locking this? This doesn't make any sense at all. It's a local variable passed by the user, this variable should thus be unique to this goroutine, unless the user is doing some weird shit that shouldn't be happening in the first place.

configV2/etcd.go
```diff
@@ -177,7 +311,9 @@ func (cfg *etcdConfig) SetTlog(tlog TlogConfig) error {
 	}
 
 	// write to etcd
+	cfg.tlogRWLock.RLocker()
```
GlenDC: Why are you locking this? This doesn't make any sense at all. It's a local variable passed by the user, this variable should thus be unique to this goroutine, unless the user is doing some weird shit that shouldn't be happening in the first place.

configV2/etcd.go
```diff
@@ -198,7 +334,9 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
 	}
 
 	// write to etcd
+	cfg.slaveRWLock.RLock()
```
GlenDC: Why are you locking this? This doesn't make any sense at all. It's a local variable passed by the user, this variable should thus be unique to this goroutine, unless the user is doing some weird shit that shouldn't be happening in the first place.

configV2/etcd.go
```diff
@@ -135,7 +263,9 @@ func (cfg *etcdConfig) SetBase(base BaseConfig) error {
 	}
 
 	// write to etcd
+	cfg.baseRWLock.RLock()
```
GlenDC: Why are you locking this? This doesn't make any sense at all. It's a local variable passed by the user, this variable should thus be unique to this goroutine, unless the user is doing some weird shit that shouldn't be happening in the first place.

configV2/etcd.go
```diff
@@ -150,13 +280,17 @@ func (cfg *etcdConfig) SetBase(base BaseConfig) error {
 // SetNBD implements ConfigSource.SetNBD
 // Sets a new nbd config and writes it to the source
 func (cfg *etcdConfig) SetNBD(nbd NBDConfig) error {
```
GlenDC: A comment which applies to all your SetSubConfig methods: It seems like you are not assigning the given subconfig to the local variable cached inside the `ConfigSource` implementation (e.g. `etcdConfig`)? Instead it seems like the only thing you're doing is writing it to the source;

configV2/fileConfig.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Base implements ConfigSource.Base
+func (cfg *fileConfig) Base() BaseConfig {
+	return cfg.base
+}
+
+// NBD implements ConfigSource.NBD
+func (cfg *fileConfig) NBD() (*NBDConfig, error) {
+	return cfg.nbd.Clone(), nil
```
GlenDC: Doesn't take into account the error case in which the NBD subconfig might not have been available in the first place.

configV2/fileConfig.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Tlog implements ConfigSource.Tlog
+func (cfg *fileConfig) Tlog() (*TlogConfig, error) {
+	return cfg.tlog.Clone(), nil
```
GlenDC: Doesn't take into account the error case in which the Tlog subconfig might not have been available in the first place.

configV2/fileConfig.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Slave implements ConfigSource.Slave
+func (cfg *fileConfig) Slave() (*SlaveConfig, error) {
+	return cfg.slave.Clone(), nil
```
GlenDC: Doesn't take into account the error case in which the Slave subconfig might not have been available in the first place.

configV2/fileConfig.go
```diff
@@ -0,0 +1,199 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/go-yaml/yaml"
+)
+
+// fileConfig represents a config using a YAML file as source
+type fileConfig struct {
```
GlenDC: `fileConfig`'s `get` and `set` methods are not goroutine safe.

configV2/etcd.go
chrisvdg: It doesn't, for now I was relying on the watcher to update my values for me, since it also triggers on events by the same client.
Btw, should values be check if they are the same when they come in from ETCD or is the extra effort not worth it and just always update the value?

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
 
 	return nil
 }
+
+// Set a watcher for a field
+func (cfg *etcdConfig) watch() {
+	// watch for fields starting with the config's vdiskID
+	watch := cfg.cli.Watch(context.TODO(), cfg.vdiskID, clientv3.WithPrefix())
```
GlenDC: I think it's better to just watch for configs with prefix `<vdiskID>:conf`. This will prevent watching keys that have nothing to do with our config keys, as I have no idea what other stuff they might store on those ETCD servers.

According to the docs and checking the code, this will even work for your base-key `<vdiskID>:conf`. However, I think you should rename the base config to contain the word `base`, might come in handy in future to have this under an explicit key, such that _all_ our configs have the format `<vdiskID>:conf:<type>`, because right now `base` is the only outlier. This also allows us to easily replace `base` with a new config in the feature, while still supporting the old base keys for legacy reasons during a transition period. Well in theory you could still do that even you choose to use the generic `<vdiskID>:conf` format, but still, I would add the base postfix :)

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
...
+				if ok {
+					log.Errorf("Watch channel encountered an error: %v", resp.Err())
+				}
+				log.Infof("Watch channel for %s closed", cfg.vdiskID)
+				return
+			}
+			// update value of key
+			for _, ev := range resp.Events {
+				log.Infof("Values for %s received an update: %s", ev.Kv.Key, ev.Kv.Value)
+				switch string(ev.Kv.Key) {
```
GlenDC: is going to be faster if you cut off the prefix, and just switch on the constants `base`, `nbd`, `slave` and `tlog`. By coincidence it will already able to discard an option based on just the first character, so switching just on these constants should be pretty fast.

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
...
+						log.Errorf("Did not update %s of %s: %s", cfg.baseKey, cfg.vdiskID, err)
+					}
+					cfg.baseRWLock.Lock()
+					cfg.base = *newBase
+					cfg.baseRWLock.Unlock()
+				case cfg.nbdKey:
+					// check if empty, if so make the subconfig nil
+					if len(ev.Kv.Value) < 1 {
+						cfg.nbdRWLock.Lock()
+						cfg.nbd = nil
```
GlenDC: please log an `Info` message for this, so that it is available in the logs that this nbd config was at some point deleted, in case it was done by accident. Make sure the log message is descriptive enough.

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
...
+						cfg.nbdRWLock.Unlock()
+						continue
+					}
+
+					// create subconfig and check validity
+					cfg.baseRWLock.RLock()
+					newNBD, err := newNBDConfig(ev.Kv.Value, cfg.vdiskID, cfg.base.Type)
+					if err != nil {
+						log.Errorf("Did not update %s of %s: %s", cfg.nbdKey, cfg.vdiskID, err)
```
GlenDC: no reason to also log the vdiskID, it's already part of the nbd key, no?

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
...
+						continue
+					}
+					cfg.baseRWLock.RUnlock()
+					cfg.nbdRWLock.Lock()
+					cfg.nbd = newNBD
+					cfg.nbdRWLock.Unlock()
+				case cfg.tlogKey:
+					if len(ev.Kv.Value) < 1 {
+						cfg.tlogRWLock.Lock()
+						cfg.tlog = nil
```
GlenDC: please log an `Info` message for this, so that it is available in the logs that this tlog config was at some point deleted, in case it was done by accident. Make sure the log message is descriptive enough.

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
...
+						cfg.tlogRWLock.Unlock()
+						continue
+					}
+
+					newTlog, err := newTlogConfig(ev.Kv.Value)
+					if err != nil {
+						log.Errorf("Did not update %s of %s: %s", cfg.tlogKey, cfg.vdiskID, err)
```
GlenDC: same as my error log comment for nbdKey

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
...
+						continue
+					}
+					cfg.tlogRWLock.Lock()
+					cfg.tlog = newTlog
+					cfg.tlogRWLock.Unlock()
+				case cfg.slaveKey:
+					if len(ev.Kv.Value) < 1 {
+						cfg.slaveRWLock.Lock()
+						cfg.slave = nil
```
GlenDC: please log an `Info` message for this, so that it is available in the logs that this slave config was at some point deleted, in case it was done by accident. Make sure the log message is descriptive enough.

configV2/etcd.go
```diff
@@ -347,3 +266,94 @@ func (cfg *etcdConfig) SetSlave(slave SlaveConfig) error {
...
+						cfg.slaveRWLock.Unlock()
+						continue
+					}
+
+					newSlave, err := newSlaveConfig(ev.Kv.Value)
+					if err != nil {
+						log.Errorf("Did not update %s of %s: %s", cfg.slaveKey, cfg.vdiskID, err)
```
GlenDC: same as my error log comment for the nbdKey

configV2/etcd.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+// NewETCDConfig returns a new etcdConfig from provided endpoints
+func NewETCDConfig(vdiskID string, endpoints []string) (ConfigSource, error) {
+	cfg := new(etcdConfig)
+	cfg.vdiskID = vdiskID
+	cfg.endpoints = endpoints
+
+	// setup keys
+	cfg.baseKey = fmt.Sprintf("%s:%s", cfg.vdiskID, "conf")
```
GlenDC: Instead of formatting them manually everywhere, I would just declare 4 functions like this. All along the following lines:

```
func etcdBaseKey(vdiskID string) string {
    return vdiskID + ":conf:base"
}

// ... add 1 func per conf
```

Also see how you don't even require the fmt.Sprintf function, as the only unique part is the prefix `vdiskID`, so no need to use such a heavy function.

Having it as functions allows you to use it everywhere you need it (including test files), making it easier to understand and change this kind of stuff.

configV2/etcd.go
```diff
@@ -0,0 +1,359 @@
...
+	// start watcher
+	cfg.watch()
+
+	return cfg, nil
+}
+
+// Close implements ConfigSource.Close
+func (cfg *etcdConfig) Close() error {
+	// close client
+	err := cfg.cli.Close()
```
GlenDC: Just make it a oneline body: `return cfg.cli.Close()`, or do something with your error, because now your conditional branch is just pointless...

configV2/config.go
```diff
@@ -62,29 +62,30 @@ func (base *BaseConfig) ToBytes() ([]byte, error) {
...
 		return fmt.Errorf("%d is an invalid size", base.Size)
 	}
-	err = base.Type.Validate()
+	err := base.Type.Validate()
 	if err != nil {
 		return fmt.Errorf("baseconfig has invalid type: %s", err.Error())
+
+	}
+	// check valid tags
+	_, err = valid.ValidateStruct(base)
```
GlenDC: Why you moved this down?

configV2/config.go
```diff
@@ -128,19 +129,15 @@ func (nbd *NBDConfig) ToBytes() ([]byte, error) {
...
 // Should only be used for ConfigSource implementation
 func (nbd NBDConfig) Validate(vdiskType VdiskType) error {
-	_, err := valid.ValidateStruct(nbd)
-	if err != nil {
-		return fmt.Errorf("invalid NBD config: %v", err)
-	}
 
 	if len(nbd.StorageCluster.DataStorage) <= 0 {
-		return fmt.Errorf("no nbd datastorage was found")
+		return fmt.Errorf("nbd datastorage was empty")
```
GlenDC: Why you moved `ValidateStruct` down? I'm fairly sure that that test would ensure that this check is not needed at all!

configV2/config.go
```diff
@@ -128,19 +129,15 @@ func (nbd *NBDConfig) ToBytes() ([]byte, error) {
...
 	}
 
 	// Check if templatestorage is present when required
 	if vdiskType.TemplateSupport(nbd) {
 		if len(nbd.TemplateStorageCluster.DataStorage) <= 0 {
```
GlenDC: This test is not correct. It's not because a vdisk type supports a template, that a specific vdisk of that type has to specify template servers.

configV2/config.go
```diff
@@ -149,6 +146,12 @@ func (nbd NBDConfig) Validate(vdiskType VdiskType) error {
 	if metadataUndefined && vdiskType.StorageType() == StorageDeduped {
 		return fmt.Errorf("metadata storage not found while required")
 	}
+
+	_, err := valid.ValidateStruct(nbd)
```
GlenDC: Why did you move this down?

configV2/test_data.go
```diff
@@ -0,0 +1,200 @@
+package configV2
```
GlenDC: Rename this file to `data_test.go` please, as the Go toolchain knows it should only compile and include files ending in the `_test` suffix for testing purposes. While in your case that's not the case. On top of that, your data structs can now be used by (foolish) accident in production code.

configV2/file_test.go
```diff
@@ -120,7 +120,7 @@ func testFileSourceWithWatcher(t *testing.T) {
 	timeout := make(chan bool, 1)
 	defer close(timeout)
 	go func() {
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(1 * time.Second)
 		recievedLock.Lock()
 		if !received {
```
GlenDC: can you put `defer receivedLock.Unlock()` please, instead of having it in the bottom. Whenever you can try to have Unlocking locks as a deferred action, this way it is ensured not to get stuck in a deadlock situation.

configV2/file_test.go
```diff
@@ -75,10 +75,10 @@ func TestFileSource(t *testing.T) {
 	}
 }
 
-func testFileSourceWithWatcher(t *testing.T) {
+func TestFileSourceWithWatcher(t *testing.T) {
 	// setup yaml file
 	// write
-	testpath := "./test.yaml"
+	testpath := "/tmp/test.yaml"
```
GlenDC: Don't do that, this will break on certain platforms (e.g. Windows). Use an std function such as https://golang.org/pkg/io/ioutil/#TempFile instead. Also the name of a tmp file should also not be hardcoded, as there are scenarios where that can break. Which is why tmp file/dir functions generate a unique file (even though you can usually define some suffix/prefix, optionally).

configV2/file.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	return nil
+}
+
+// Base implements ConfigSource.Base
+func (cfg *fileConfig) Base() BaseConfig {
+	cfg.baseRWLock.RLock()
+	defer cfg.baseRWLock.RUnlock()
+	return cfg.base
```
GlenDC: You also have to clone the base cfg, as it contains slices which in Go are passed by reference. Thus cloning is required as to ensure those slices are unique.

configV2/file.go
```diff
@@ -0,0 +1,326 @@
...
+	cfg.slave = &slave
+	cfg.slaveRWLock.Unlock()
+
+	// write to file
+	return cfg.writeToSource()
+}
+
+// watch starts watching for SIGHUP signal, then update the config
+func (cfg *fileConfig) watch() {
+	sighub := make(chan os.Signal)
```
GlenDC: why do you call this `sighub` instead of `sighup`?

configV2/file.go
```diff
@@ -0,0 +1,326 @@
...
+	cfg := new(fileConfig)
+	buf := new(yamlConfigFormat)
+	// unmarshal the yaml content
+	err := yaml.Unmarshal(bytes, buf)
+	if err != nil {
+		return nil, fmt.Errorf("Could not unmarshal provided bytes: %v", err)
+	}
+
+	// debug
+	//log.Info("NBD Values: %v", buf.NBD)
```
GlenDC: can be removed, no? Also for debugging purposes there is `log.Debug` and `log.Debugf`

configV2/file.go
```diff
@@ -0,0 +1,326 @@
...
+	cfg.base = buf.Base
+	cfg.nbd = buf.NBD
+	cfg.tlog = buf.Tlog
+	cfg.slave = buf.Slave
+
+	err = cfg.base.Validate()
+	if err != nil {
+		return nil, fmt.Errorf("invalid base configuration: %s", err)
+	}
+	if cfg.nbd != nil {
```
GlenDC: This nil check can be placed inside the Validate method, so that this nil check is not needed here, keeping your logic here a bit more concentrated and clean :)

So if your nbd config Validate method was implemented as:

```go
func (nbd *NBDConfig) Validate(vdiskType VdiskType) error {
	if nbd == nil {
		return nil
	}

	// ... validation

	return nil
}
```

Than you can omit the `if cfg.nbd != nil {` check here, making the nbd config much more pleasant to use. Oh by the way, declare your methods by value, otherwise you'll be copying values on the stack for each call. Sometimes its useful to have them by value, but this is not this case.

And in general, you should ensure your methods can handle the case where the object itself is nil, making it even more useful, only if it makes sense of course. The talk at https://www.youtube.com/watch?v=ynoY2xz-F8s explains this concept and more pretty well I think.

This concept can be applied in your entire module. Only where it makes sense of course. It's one of those things people don't seem to understand, which makes them complain they have to do nil-checks everywhere, even though there are ways to keep code clean without needing to write nil checks everywhere. This is one of those ways.

GlenDC:
@chrisvdg all in all we're getting there. Tomorrow will be a busy day for you, here is what I hope you manage to do:

1. Finish the API implementation of it (see my proposed API below);
2. Get the latest config from master, make sure your configv2 contains all  changes from Config;
3. Move configv2 to Config and integrate it in the entire codebase;
4. Finish unit testing;

All in all you're on the right track. However I think we have to split up the ConfigSource as well, rather than have 1 big one:

+ There is no advantage in having 1 big one, as we'll never require a big one anyhow, all config users only ever need 1 config (nbd, tlog or slave);
+ It makes the API of the Source interface really big, and that will only be bigger once we also add the required `Subscribe`/`Unsubscribe` methods;

I was hoping to keep the new config API simpler, however I don't think we can live without the `Subscribe`/`Unsubscribe` methods. This allows you to subscribe to (watch) a source and automatically retrieve it a new config if it is updated. Without this model, we'll be cloning a lot I'm afraid, especially in the NBD storage code. The provider requires the config each time a vdisk operation is applied, sometimes multiple times for 1 operation. All those clones are really going to hurt our performance, but without the subscribe/unsubscribe model it's required, as a config update in the middle of using a config will really give weird results. The subscribe/unsubscribe model makes it that we only have to clone once per subscriber per reload. A lot less thus.

With all that in mind, here is the modified API I have in mind:

```go
func BaseConfigFromFile(vdiskID string, path string) (*BaseConfig, error)
func BaseConfigFromETCD(vdiskID string, endpoints []string) (*BaseConfig, error)

type BaseConfig struct {
	BlockSize uint64    `yaml:"blockSize" valid:"optional"`
	ReadOnly  bool      `yaml:"readOnly" valid:"optional"`
	Size      uint64    `yaml:"size" valid:"optional"`
	Type      VdiskType `yaml:"type" valid:"optional"`
}

func (cfg *BaseConfig) Validate() error
func (cfg *BaseConfig) Clone() *BaseConfig
func (cfg *BaseConfig) ToBytes() ([]byte, error)

func NBDConfigFromFile(vdiskID string, path string) (*NBDConfig, error)
func NBDConfigFromETCD(vdiskID string, endpoints []string) (*NBDConfig, error)

type NBDConfig struct {
	StorageCluster         StorageClusterConfig `yaml:"storageCluster" valid:"required"`
	TemplateStorageCluster StorageClusterConfig `yaml:"templateStorageCluster" valid:"optional"`
	TemplateVdiskID        string               `yaml:"templateVdiskID" valid:"optional"`
}

func (cfg *NBDConfig) Validate(VdiskType) error
func (cfg *NBDConfig) Clone() *NBDConfig
func (cfg *NBDConfig) ToBytes() ([]byte, error)

func NBDConfigFileSource(vdiskID string, path string) (NBDConfigSource, error)
func NBDConfigETCDSource(vdiskID string, endpoints []string) (NBDConfigSource, error)

type NBDConfigSource interface {
	NBDConfig() (*NBDConfig, error) // Returns the latest NBD config version
        SetNBDConfig(NBDConfig) error
	Close() error                     // closes source connection and goroutines if present

	// all NBDConfigSource implementations will probably contain
	// a static BaseConfig which gets loaded at creation time,
	// and does never need to be hot reloaded,
	// as none of that info should ever be supported
	VdiskType() VdiskType

	// see config.HotReloader for inspiration for Subscribe/Unsubscribe logic
	Subscribe(c chan<- NBDConfig) error
	Unsubscribe(c chan<- NBDConfig) error
}

func TlogConfigFromFile(vdiskID string, path string) (*TlogConfig, error)
func TlogConfigFromETCD(vdiskID string, endpoints []string) (*TlogConfig, error)

type TlogConfig struct {
	TlogStorageCluster StorageClusterConfig `yaml:"tlogStorageCluster" valid:"optional"`
	SlaveSync          bool                 `yaml:"tlogSlaveSync" valid:"optional"`
}

func (cfg *TlogConfig) Validate() error
func (cfg *TlogConfig) Clone() *TlogConfig
func (cfg *TlogConfig) ToBytes() ([]byte, error)

func TlogConfigFileSource(vdiskID string, path string) (TlogConfigSource, error)
func TlogConfigETCDSource(vdiskID string, endpoints []string) (TlogConfigSource, error)

type TlogConfigSource interface {
	TlogConfig() (*TlogConfig, error) // Returns the latest tlog config version
        SetTlogConfig(TlogConfig) error
	Close() error                     // closes source connection and goroutines if present

	// see config.HotReloader for inspiration for Subscribe/Unsubscribe logic
	Subscribe(c chan<- TlogConfig) error
	Unsubscribe(c chan<- TlogConfig) error
}

func SlaveConfigFromFile(vdiskID string, path string) (*SlaveConfig, error)
func SlaveConfigFromETCD(vdiskID string, endpoints []string) (*SlaveConfig, error)

// SlaveConfig represents a backup storage configuration
type SlaveConfig struct {
	SlaveStorageCluster StorageClusterConfig `yaml:"slaveStorageCluster" valid:"optional"`
}

func (cfg *SlaveConfig) Validate() error
func (cfg *SlaveConfig) Clone() *SlaveConfig
func (cfg *SlaveConfig) ToBytes() ([]byte, error)

func SlaveConfigFileSource(vdiskID string, path string) (SlaveConfigSource, error)
func SlaveConfigETCDSource(vdiskID string, endpoints []string) (SlaveConfigSource, error)

type SlaveConfigSource interface {
	SlaveConfig() (*SlaveConfig, error) // Returns the latest slave config version
        SetSlaveConfig(SlaveConfig) error
	Close() error                       // closes source connection and goroutines if present

	// see config.HotReloader for inspiration for Subscribe/Unsubscribe logic
	Subscribe(c chan<- SlaveConfig) error
	Unsubscribe(c chan<- SlaveConfig) error
}
```

configV2/config.go
```diff
@@ -0,0 +1,501 @@
...
+// has to send to the tlog server, to log its transactions.
+func (vdiskType VdiskType) TlogSupport() bool {
+	return vdiskType&propTlogSupport != 0
+}
+
+// TemplateSupport returns whether or not
+// this vdisk supports a template server,
+// to get the data in case the data isn't available on
+// the normal (local) storage cluster.
+func (vdiskType VdiskType) TemplateSupport(nbd NBDConfig) bool {
```
GlenDC: There is no need to pass the `nbd NBDConfig` here. I guess it's outdated code. Either way, please remove this parameter.

configV2/config.go
chrisvdg: Since the error message I check for myself are more descriptive, especially when testing nbd I wasn't quite sure sometimes what ValidateStruct was invalidating

configV2/file.go
chrisvdg: BaseConfig doesn't contain slices (2x uint64, a bool and a vdisktype which is a uint8)

config/doc.go
```diff
@@ -2,29 +2,33 @@
...
+When calling this function it will return an error if no connection could be made or the current data is invalid. If no error occurred it will return a channel and send the current subconfig to the channel.
+The configs stored at the etcd keys are YAML formatted serialized versions of the subconfig.
+
+NOTE: For the NBDConfig there should be a valid BaseConfig available in the source (e.g. etcd), as it is required to validate the NBDConfig.
 
 File source
 
-The file config source is recommended only for development use and gets it's config data from a provided YAML file location. 1 YAML file can represents the data of 1 vdisk formatted to the configFileFormat type.
-It will automatically devide the data in the different subconfigs after reading from the source file.
+The fileConfig source is recommended for development use and gets it's config data from a provided YAML file path. 1 YAML file represents the data of 1 vdisk formatted to the yamlConfigFormat type.
```
GlenDC: Replace `1 vdisk` with `one or multiple vdisks`. Also use the `one`, `two`, `three`, etc instead of `1`, `2`, `3`, when talking about quantity of things in documentation. Only use them in the form of `1` when talking about maths-related stuff, or when saying something like `4 KiB`.

config/doc.go
```diff
@@ -2,29 +2,33 @@
...
+	slave: <VdiskID>:conf:slave
 
-NOTE: For the NBDConfig there should be a valid BaseConfig in the source to validate NBDConfig
+The Get Read<Sub>ConfigETCD functions will connect to the etcd store and get the data for the subconfig. It uses the subconfig's constructor to create and validate the subconfig and returns an error if this or the connection to the etcd store fails.
+
+The Watch<Sub>ConfigETCD functions uses the etcdv3 Watch API to listen for updates of the  subconfig and send that updated value if valid to the channel returned from the function.
+When calling this function it will return an error if no connection could be made or the current data is invalid. If no error occurred it will return a channel and send the current subconfig to the channel.
+The configs stored at the etcd keys are YAML formatted serialized versions of the subconfig.
+
+NOTE: For the NBDConfig there should be a valid BaseConfig available in the source (e.g. etcd), as it is required to validate the NBDConfig.
```
GlenDC: you seem to be inconsistent between using `NBDConfig` and `nbd`. I think to be more compatible with your previous usage of `nbd`, you could use `nbd config` here.

config/doc.go
```diff
@@ -56,7 +60,12 @@ YAML file example:
 			address: 192.168.58.146:2001
 			db: 8
 
-NOTE: For the NBDConfig there should be a valid BaseConfig in the source to validate NBDConfig
+The Get Read<Sub>ConfiFile functions will the source file and get the data for the subconfig. It uses the subconfig's constructor to create and validate the subconfig and returns an error if this or the reading the file fails.
+
+The Watch<Sub>ConfigFile functions listens to the SIGHUP signal for updates of the configs and send that updated value if valid to the channel returned from the function.
+When calling this function it will return an error if it could not read the source file or the current data is invalid. If no error occurred it will return a channel and send the current subconfig to the channel.
+
+NOTE: For the NBDConfig there should be a valid BaseConfig available in the source (e.g. file), as it is required to validate the NBDConfig.
```
GlenDC: same as previous comment on the NOTE

config/etcd_test.go
```diff
@@ -59,13 +59,13 @@ func testETCDConfig(t *testing.T) {
 		return
 	}
 
-	nbd, err := ReadNBDConfigETCD(vdiskID, endpoints)
+	nbdBa, nbd, err := ReadNBDConfigETCD(vdiskID, endpoints)
```
GlenDC: What does the `B` stand for in this named local variable, I really have no clue.

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
+The data is split up in this way so that for the use cases of 0-Disk, most of the time only one subconfig is required. Should another subconfig be required, it can easily be fetched from a source and stored together with the other subconfigs.
 
 * NBD Server uses: BaseConfig, NBDConfig, SlaveConfig
 * Tlog Server (server.vdisk) uses: TlogConfig
 
 The source implementations of the config (etcd) allows for data being read from the source and include a watch/update feature that returns an updated subconfig as it is updated by the 0-orchestrator. This feature is not implemented for BaseConfig as that data should be quite static.
 
+## Subconfigs
+### Base config
+* BaseConfig: stores general vdisk information:
```
GlenDC: I would somehow like the layout as follows more:

```
Stores general information for a vdisk.

* `BlockSize`: Size of a block on the vdisk
* `ReadOnly`: Defines if vdisk is readonly
* `Size`: vdisk size
* `Type`: Type of vdisk (Boot, DB, Cache, tmp)
* `TlogRPC`: Defines a list of network addresses of tlogserver(s)
```

Also in `./docs` folder we now have a `glossary.md` file, and also other docs. Please link to a terminology in the `glossary.md` if possible (e.g. `vdisk`, `boot`, `db`, `cache`, `tmp`. Also be consistent with caps in these 4 last terminologies). You can see an example of how to do the linking in the `./docs/nbd/storage` directory.

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
-The data is split up in this way so that for the use cases of 0-Disk, most of the time only the BaseConfig and/or one of the other subconfigs would be required. Should another subconfig be required, it can easily be fetched from a source and stored together with the other subconfigs.
+The data is split up in this way so that for the use cases of 0-Disk, most of the time only one subconfig is required. Should another subconfig be required, it can easily be fetched from a source and stored together with the other subconfigs.
 
 * NBD Server uses: BaseConfig, NBDConfig, SlaveConfig
 * Tlog Server (server.vdisk) uses: TlogConfig
 
 The source implementations of the config (etcd) allows for data being read from the source and include a watch/update feature that returns an updated subconfig as it is updated by the 0-orchestrator. This feature is not implemented for BaseConfig as that data should be quite static.
 
+## Subconfigs
+### Base config
```
GlenDC: Can you please _always_ add 1 empty line below headers? Sorry I know markdown renderers handles this fine, but programmers often read markdown files in their raw format, at least I do, and than having a space makes it a bit easier on the eyes.

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
+* BaseConfig: stores general vdisk information:
+	* BlockSize: Size of a block on the vdisk
+	* ReadOnly: Defines if vdisk is readonly
+	* Size: vdisk size
+	* Type: Type of vdisk (Boot, DB, Cache, tmp)
+	* TlogRPC: Defines a list of network addresses of tlogserver(s)
+
+* YAML example:
+```yaml
```
GlenDC: same in cases like these, having 1 empty line below `YAML example` make it somehow easier on the eyes.

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
+blockSize: 4096 # can not be zero or uneven
+readOnly: false
+size: 10 	# can not be 0
+type: db	# should be a valid vdisk type (boot, db, cache or tmp)
+tlogrpc: 192.168.58.123:2000,192.168.58.125:2000
```
GlenDC: explain that you can give one or multiple addresses, and if you give multiple they have to be comma separated (better be explicit in docs you know)

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
+```
+
+* Used by:
+	* [NBD Server][nbdServerConfig]:
```
GlenDC: just do `* [NBD Server][nbdServerConfig] (e.g. used for the BackendFactory)`. Also provide a link the `backend_factory.go` file for the BackendFactory word.

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
+		* ardb uses it for the BackendFactory
+
+* Godoc: [BaseConfig][baseconfigGodoc]
+
+### NBD config
```
GlenDC: My comments for the Base Config apply for the NBD Config as well.

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
+      db: 10
+```
+
+* Used by:
+	* [NBD Server][nbdServerConfig]:
+		* ardb uses it for the BackendFactory
+
+* Godoc: [NBDConfig][nbdconfigGodoc]
+
+### Tlog config
```
GlenDC: My comments for the Base Config apply for the Tlog Config as well.

docs/config.md
```diff
@@ -4,36 +4,119 @@ All 0-Disk services are configured using a YAML config source, written by the [0
...
+```
+
+* Used by:
+	* [Tlog Server][tlogServerConfig]:
+		* server.vdisk uses the tlogSlaveSync
+		* slavesync uses the TlogStorageCluster
+
+* Godoc: [TlogConfig][tlogconfigGodoc]
+
+### Slave config
```
GlenDC: My comments for the Base Config apply for the Slave Config as well.

docs/config.md
```diff
@@ -42,43 +125,45 @@ The source implementation for etcd provides methodes to read subconfigs from an
 In the etcd cluster it will look for the following keys for the subconfigs:
 
 ```
-	BaseConfig: <VdiskID>:conf:base
-	NBDConfig: <VdiskID>:conf:nbd
-	TlogConfig: <VdiskID>:conf:tlog
-	SlaveConfig: <VdiskID>:conf:slave
+BaseConfig: <VdiskID>:conf:base
```
GlenDC: for each config here, and wherever you reference a subconfig within this document, you can reference to the header, by writing something like `([BaseConfig](#base-config)`.

docs/config.md
```diff
@@ -42,43 +125,45 @@ The source implementation for etcd provides methodes to read subconfigs from an
...
+NBDConfig: <VdiskID>:conf:nbd
+TlogConfig: <VdiskID>:conf:tlog
+SlaveConfig: <VdiskID>:conf:slave
 ```
 
 The values stored in those keys are the subconfigs serialised in YAML format. This can be done using the subconfig's ToBytes() method.
 
-When the values are read from etcd, the subconfig will be created with the New<Sub>Config constructor, which will also check if the provided byte slice is a valid subconfig.
+When the values are read from etcd, the subconfig will be created with the subconfig's constructor, which will also check if the provided byte slice is valid.
```
GlenDC: `subconfigs`*

`subconfig's constructor` == `subconfig his constructor`, which is not what you meant I assume. I guess it's because of your mother tongue that you are inclined to put an apostrophe in front of the `s` in cases like this, but please pay attention as to not to do it.

docs/config.md
```diff
@@ -42,43 +125,45 @@ The source implementation for etcd provides methodes to read subconfigs from an
...
-
 e.g.:
 ```go
 func WatchNBDConfigETCD(ctx context.Context, vdiskID string, endpoints []string) (<-chan NBDConfig, error)
 ```
-WatchNBDConfigETCD will return a channel that sends NBDConfig's. As seen as this function is called, it will fetch and validate the NBDConfig on the etcd cluster. If successful it will return the channel and send the current NBDConfig on etcd. If not and error is returned.
-WatchNBDConfigETCD also launches a Goroutine and using the etcd Watch API it will send an updated NBDConfig to the returned channel when they are received from the Watch API and validated. If the updated value on etcd is not valid, no new NBDConfig will be send to the channel and the error will be logged.
-
+WatchNBDConfigETCD will return a channel that sends a NBDConfig instance right at the start, and each time it is updated following that. If the NBDConfig didn't exist yet at startup or any other error occurred while doing the initial fetching, this method will return an error instead.
```
GlenDC: + When talking about names in codebases (e.g. constants, functions, types), put them around back ticks, as in \`WatchNBDConfigETCD\`;
+ Link to `NBDConfig`, as in `[NBDConfig](#nbd-config)`.

docs/nbd/config.md
```diff
@@ -14,7 +14,7 @@ The NBD Server will use the methodes provided by the config package to communica
 
 Within the NBD server, it uses the configs in following places:
   * ardb: use the BaseConfig and NBDConfig for the BackendFactory
-  * Tlogstorage: when switching to ardb slave the SlaveConfig will be used
+  * nbdserver.tlog : when switching to the ardb slave the SlaveConfig will be used
```
GlenDC: tlogStorage*, just keep using that, you already mention `Within the NBD server`, so that should be clear enough that they have to look in the `nbdserver` dir for these things.

docs/nbd/config.md
```diff
@@ -14,7 +14,7 @@ The NBD Server will use the methodes provided by the config package to communica
 
 Within the NBD server, it uses the configs in following places:
   * ardb: use the BaseConfig and NBDConfig for the BackendFactory
```
GlenDC: + Link to baseConfig and NBDConfig headers;
+ Link to BackendFactory file;

docs/nbd/config.md
GlenDC: + Link the tlogStorage file;
+ Link to SlaveConfig header;

docs/config.md
chrisvdg: The apostrophe is valid in English as well, only they have decided it doesn't matter if you put it there or not anymore. https://www.englishclub.com/grammar/nouns-possessive.htm

GlenDC: Oh sorry, rereading your sentence made me realize I was wrong. When I was reading it originally I thought you were talking here about multiple subconfigs. My bad.

khaledkbadr:
I can't find any zeroctl related info in the `config.md`, we're not using `etcd` for that?

docs/config.md
```diff
@@ -1,10 +1,284 @@
...
+### Reading data
+
+The source implementation for [etcd][etcd] provides methodes to read and watch subconfigs from an etcd cluster. 
+In the etcd cluster it will look for the following subconfig keys:
+
+* [BaseConfig][baseconf]: `<VdiskID>:zerodisk:conf:base`;
+* [NBDConfig][nbdconf]: `<VdiskID>:zerodisk:conf:nbd`;
+* [TlogConfig][tlogconf]: `<VdiskID>:zerodisk:conf:tlog`;
+* [SlaveConfig][slaveconf]: `<VdiskID>:zerodisk:conf:slave`;
+* [VdisksConfig][vdisksconf]: `zerodisk:conf:vdisks`;
```
grimpy: since this is a yaml store in the etcd config we will have issues attomicly updating this values.
Removing/adding disks.

GlenDC:
> I can't find any zeroctl related info in the config.md, we're not using etcd for that?

You're right, @khaledkbadr, I should add documentation for the "configuration" of that tool as well. Will do it.

> > All 0-Disk services expose a -config flag

The paragraph following the above statement, applies to all 0-Disk services, including the zeroctl tool. So that's already information you have. But you're right, it should also have nice config-related information to be explicit about all of it.

docs/config.md
GlenDC: First of all, etcd has no data types (in contrast to something like redis), but I guess that is besides you question.

To get to your question, @grimpy. This is something you'll only have to fix on _your_ side, as you're the only one writing. We just read and watch. However, the good news is that since etcd v3 you can easily resolve this using etcd3 transactions. https://coreos.com/blog/transactional-memory-with-etcd3.html talks a bit about this.

GlenDC:
After discussion with @khaledkbadr and @grimpy, some changes in the design are in order, and the following is left to be done to finish this PR:

### improved documentation

+ improve the 0-Disk config overview (dot) graph;
+ add info on what (sub)configs are used in zeroctl in the 0-Disk config overview markdown doc;
+ add zeroctl configuration, in a similar style as the NBD Server- and Tlog Server configuration docs, explaining what is used where and how it effects the usage of the relevant zeroctl commands;

### etcd key format

The new key formats would be as follows:

* VdiskStaticConfig: `<VdiskID>:vdisk:conf:static` (used to be `<VdiskID>:zerodisk:conf:base`);
* VdiskNBDStorageConfig: `<VdiskID>:vdisk:conf:storage:nbd` (used to be `<VdiskID>:zerodisk:conf:nbd`);
* VdiskTlogStorageConfig: `<VdiskID>:vdisk:conf:storage:tlog` (used to be `<VdiskID>:zerodisk:conf:tlog`);

* NBDVdisksConfig: `<ID>:nbdserver:conf:vdisks` (used to be `zerodisk:conf:vdisks`);

Where the `<ID>` can be specified for each _nbdserver_ using an `-id` flag.
If not given explicitly it will be defaulted to `default`.
An `<ID>` should be unique per nbdserver process if possible.

* StorageClusterConfig: `<ID>:cluster:conf:storage` (didn't exist yet);
* TlogClusterConfig: `<ID>:cluster:conf:tlog` (didn't exist yet);

Where the `<ID>` is to be unique per cluster.

### shared [ARDB] storage cluster definitions

Currently each [ARDB] storage cluster is explicitly and fully specified per vdisk subconfig. Meaning that if we have 25 vdisks which use the same [primary] storage cluster, we would have specified it 25 times. Now instead we only define each cluster once, and simply refer to it for each vdisk that uses it.

For a vdisk `foo` a NBDVdiskStorageConfig `foo:vdisk:conf:storage:nbd` (using the new proposed key format) will now be stored as follows:

```yaml
storageClusterID: a9823a788388027d9a7a8c19f49e786b # required
templateStorageClusterID: ae3e83e2fab3a7d8683d8eefabd1e74d
templateVdiskID: testtemplate # optional, same as VDisk ID if not defined,
                              # only used for nondeduped vdisks
```

For a vdisk `foo` a TlogVdiskStorageConfig `foo:vdisk:conf:storage:tlog` (using the new proposed key format) will now be stored as follows:

```yaml
storageClusterID: asdadad32342sdsd345adasdae23425asd2 # required
slaveStorageClusterID: sdafgwr454wrsfsfsfsdsadawararfaffsa # optional
tlogServerClusterID: sadadadrwresdsadadsadqweasdadaseqr # optional
```

A storage cluster `mycluster`'s config `mycluster:cluster:conf:storage` is new and stored as:

```yaml
dataStorage:  # at least one server is required
  - address: 192.168.1.146:2000 # has to be a valid dial string
    db: 10 # optional, 0 by default
  - address: 192.123.123.1:2001
    db: 10
metadataStorage: # not used for nondeduped vdisks, required for the others
  address: 192.168.1.146:2001
  db: 11
```

And finally a tlog server cluster `mytlog`'s config `mytlog:cluster:conf:tlog` is new and stored as:

```yaml
servers:
  - address: 192.168.2.1
  - address: 192.168.2.2
  - address: 192.168.2.3
```

### validation at startup

When starting an NBDServer:

+ Is is tested that the given etcd cluster is reachable;
+ For each given vdiskID is is tested that the full given NBD-related Vdisk Configuration is valid;

When starting a TLogserver:

+ Is is tested that the given etcd cluster is reachable;
+ For each given vdiskID is is tested that the full given TLog-related Vdisk Configuration is valid;

GlenDC:
We also need to ensure error messages related to invalid configs are descriptive enough.

docs/config.md
```diff
@@ -1,10 +1,284 @@
...
+    address: 192.168.1.1:1001
+    db: 18
+slaveSync: true # optional, false by default
+```
+
+Used by the [Tlog Server][tlogServerConfig].
+
+See the [TlogConfig Godoc][tlogconfigGodoc] for more information.
+
+### Slave Config
```
khaledkbadr: Is config updated?

PR #337, opened 2017-07-18
nbdserver sigterm handler.

### LINKED ISSUE #323: exit nbdserver gracefully upon SIGTERM
As reported in #288, the Cairo team seems to send a `SIGTERM` signal to the `nbdserver`, when done with it. Once the `tlogserver` and `nbdserver` processes have both exited, the VM is shut down. The `SIGTERM` signal is sent right after being done with the Vdisk.

We have to handle the `SIGTERM`, such that we can gracefully exit the nbdserver and shutdown any active vdisks. This will probably result in some errors in case there were active processes, but at least now it will be done gracefully rather than abrupt if not handled properly.

iwanbk:
Fixes #323 
Wait for all tlog vdisks completion.

### DIFF
--- a/nbdserver/main.go
+++ b/nbdserver/main.go
@@ -6,7 +6,9 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/signal"
 	"sync"
+	"syscall"
 
 	_ "net/http/pprof"
 
@@ -168,6 +170,8 @@ func main() {
 		LBACacheLimit:     lbacachelimit,
 		ConfigPath:        configPath,
 	})
+	handleSigterm(backendFactory, cancelFunc)
+
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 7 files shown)


nbdserver/main.go
```diff
@@ -197,6 +201,33 @@ func main() {
...
+func handleSigterm(bf *ardb.BackendFactory, cancelFunc context.CancelFunc) {
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGTERM)
+
+	go func() {
+		<-sigs
+
+		errs := bf.Wait()
+		for _, err := range errs {
+			// TODO : log to stderr properly
```
GlenDC: What about this TODO?

iwanbk: It is explained in the next line 
```
// depends on : https://github.com/zero-os/0-Disk/issues/300
```

iwanbk:
```
However: i think the SIGTERM handling should be moved from the tlogStorage to the nbd.Backend. This way all storage types and thus vdisks would be supported. The nbd.Backend can then call first the Flush method (having a time limit does seem acceptable I guess), to then call the Close method.
```
Indeed.

I currently doesn't have enough knowledge to do this in this tight sprint
I created #339 to address this. I might work on it later.

```
Besides that it looks good enough for now.
```
OK, i merge it

GlenDC:
Thanks Iwan 👍

Issue #322, opened 2017-07-11
exit tlogserver gracefully upon SIGTERM


GlenDC:
As reported in #288, the Cairo team seems to send a `SIGTERM` signal to the `tlogserver`, when done with it. Once the `tlogserver` and `nbdserver` processes have both exited, the VM is shut down. The `SIGTERM` signal is sent right after being done with the Vdisk.

We have to handle the `SIGTERM`, such that we can gracefully exit the tlogserver and finish the critical things we're doing before exiting. I propose we do the following things upon `SIGTERM`:

+ Block any new incoming connections (we could provide a new `HandshakeStatus` for this;
+ Close any open connections (otherwise this connection might keep sending stuff forever, while we want to quit), this shouldn't be an issue as the nbdserver (the client) is also gracefully exiting anyhow;
+ Finish turning all received transactions into aggregations, serialize them, and exit afterwards, return an exit error code in case the this step couldn't finish for some reason, this way 0-Orchestrator or whoever is using tlogserver can handle such an error (make sure to document this of course);

Putting this on milestone 6 as critical, as without this feature, tlog will keep failing for the Cairo team.

iwanbk:
Suggest to wait until we have clear problem on #288, especially regarding the data left in OS cache and flush

iwanbk:
> Putting this on milestone 6 as critical, as without this feature, tlog will keep failing for the Cairo team.

Moved to milestone 7 and remove 'critical' priority.

It is because nbdserver's flush always wait for tlog's flush.

iwanbk:
Move it to milestone 8 with these reasons:
- it wont be finished on time
- not really needed for now

PR #340, opened 2017-07-19
move nbd sigterm handling to backend

GlenDC:
Here is an implementation of what I proposed in your merged `SIGTERM` PR.

Let me know what you think, and feel free to merge it if you like it.

I still think we should also handle the `SIGTERM` signal in the `tlogserver`, to ensure that the `tlogserver` is also gracefully quit. I know, in normal cases it should be fine without, but in production we don't always know what is normal. If you agree, we could make an issue for it, and plan it for milestone 7.


nbdserver/ardb/backend_sigterm_test.go
```diff
@@ -0,0 +1,90 @@
...
+	if !assert.NoError(t, err) || !assert.Equal(t, blockSize, bw) {
+		return
+	}
+
+	// sending SIGTERM
+	syscall.Kill(syscall.Getpid(), syscall.SIGTERM)
+
+	// make sure we get completion error object
+	errs := vComp.Wait()
+	assert.Equal(t, 0, len(errs), "expected no errs, but received: %v", errs)
```
iwanbk: In this new test, it get no errors at all.

Here is the test from previous PR
```go
 // make sure we get completion error object
    errs := vComp.Wait()
    assert.Equal(t, 1, len(errs))
    assert.Nil(t, errs[0])
```

iwanbk:
> I still think we should also handle the SIGTERM signal in the tlogserver, to ensure that the tlogserver is also gracefully quit. I know, in normal cases it should be fine without, but in production we don't always know what is normal. If you agree, we could make an issue for it, and plan it for milestone 7.

We already have the issue.

nbdserver/ardb/backend_sigterm_test.go
GlenDC: Yes, I know. It's not an oversight. I just think it's weird to return nil as error values. Returning those even when nil would only make sense if you know for sure there is some kind of order. However here there is no order, so there is no real value in having `nil` values. Hence, I'm only returning errors, if there are indeed errors.

GlenDC:
> We already have the issue.

Ah, you're correct! I forgot about https://github.com/zero-os/0-Disk/issues/322.

nbdserver/ardb/backend_sigterm_test.go
iwanbk: > Returning those even when nil would only make sense if you know for sure there is some kind of order. However here there is no order, so there is no real value in having nil values. Hence, I'm only returning errors, if there are indeed errors.

Hmm....i was thinking that when a `func` has `error` as return value, it will return `err` or `nil`, not `err` or `nothing`. But i have no strong reason here, so i'm OK whatever way we go.

But in this case, the test is not clear whether the backend exited under SIGTERM handler or not.
In previous PR, if we have 2 backends exited under SIGTERM handler then we have 2 elements in the `errors`

GlenDC: It's still clear, no? All backends support the SIGTERM handler. Thus if we have no errors, than all active vdisks were successfully flushed and closed. I don't think it's important to know how many vdisks were closed (even though this can be seen from the info logs). But if the flushing timed out, or an error happened during either/both the flushing or/and the closing, these errors are returned, and can be logged (using the info or formatted logger).

PR #356, opened 2017-07-28
isolate BlockStorage and bring order to nbd module

### LINKED ISSUE #350: refactor core storage code into public centralized module
The deduped, nondeduped and semideduped storage should be moved to the new `nbd/ardb/storage`. `lba` can be moved to `nbd/ardb/storage/lba` and `nbdserver` can be moved to `nbd/nbdserver` following closer the way we structure the tlogserver.

`backendStorage` will be in the storage module as well and be renamed to the public `BlockStorage interface:

```go
type BlockStorage interface {
    SetBlock(index int64, content []byte) (err error)
    GetBlock(index int64) (content []byte, err error)
    DeleteBlock(index int64) (err error)
    
    Close() error
    Flush() error
}
```

Fetching multiple blocks at once (`GetBlocks`) won't be available for now. This is planned for a later milestone (see: #352). There is also no `SetBlocks` variant, as we'll optimize this function in a later milestone (see: #351).

The following public functions will be available for each core storage (`deduped`, `nondeduped` and `semideduped`):

+ A constructor to create them;
+ A function to copy an entire `vdisk` (on the same or different server);
+ A function to delete an entire `vdisk`;

The `Backend` will remain in the `ardb` module together with the `tlogStorage`. The `Backend` will still wrap around a `BlockStorage`.

By having this logic we will now only ever use the `Backend` for the `gonbdserver`, and use the `BlockStorage` (and its companion functions) for everything else:

+ `zeroctl`: copy, delete, as well as import and export (see: #317);
+ `tlogreplay`: uses backend now, will use blockStorage instead (might mean we need to adapt capnp schema a bit, as we would now keep it in blocks, but I think that's fine, as the offset is now more or less forced anyhow);

Doing this will allow us to keep all the internal ardb storage logic within the `storage` module, rather than having it spread over the `ardb` and mutliple `zeroctl` modules.

This issue is pretty easy (and should be fast to resolve), as it doesn't introduce any new code, it just refactors what is already there into new modules and slightly new names.

GlenDC:
fixes #350

+ nbd.backend, tlogStorage and inMemoryStorage are now private and only known to nbdserver;
+ tlogplayer makes now use of BlockStorage instead of backend;
+ tlog-related tests previously using nbd backend, now use tlogStorage;
+ some tlog-related tests have moved to nbdserver main package, as tlogStorage is private;
+ testRedisProvider is now promoted to redisstub.InMemoryRedisProvider and easier to use;
+ Copy, Delete and List logic related to vdisks, now is centralized and defined within the ardb.storage module;
+ zeroctl doesn't define any real logic anylonger and is more of an interface, using the ardb.storage module;


nbd/ardb/storage/deduped.go
```diff
@@ -0,0 +1,393 @@
...
+	if err != nil {
+		return
+	}
+	defer conn.Close()
+
+	exists, err := redis.Bool(conn.Do("EXISTS", hash.Bytes()))
+	if err == nil && !exists {
+		_, err = conn.Do("SET", hash.Bytes(), content)
+		success = err == nil
+	}
```
iwanbk: Why not using SETNX? https://redis.io/commands/setnx

nbd/ardb/storage/nondeduped.go
```diff
@@ -0,0 +1,368 @@
...
+// Set implements BlockStorage.Set
+func (ss *nonDedupedStorage) SetBlock(blockIndex int64, content []byte) (err error) {
+	// get a connection to a data storage server, based on the modulo blockIndex
+	conn, err := ss.provider.DataConnection(blockIndex)
+	if err != nil {
+		return
+	}
+	defer conn.Close()
+
+	// don't store zero blocks,
```
iwanbk: What if the content is really zero?

GlenDC: This is not new code. That's fine, the backend handled this since ever. When you get that block, it will just return an all zeroes block, so that's not an issue. You will still get zeroes if you really wanted zeroes.

nbd/ardb/storage/deduped.go
GlenDC: This is old code. But the goal is not to not overwrite a block, as this is deduped, so as long as the hash matches, the content will be the same anyhow (unless the hashing algo is broken). It's more about not having to send the entire block data over the wire if it's already set anyhow.

But.. now that I think about it. I wonder if this really is an optimization, or if it doesn't just make everything slower, by requiring 4 roundtrips, instead of 2, in case you really have to set the block...

iwanbk: If the packet size is less than the MTU, then i'm sure that it just make things slower.
But i'm sure the packet size is bigger

GlenDC: Not sure if it is larger. Especially when we use 4 KiB blocks, which is what the size is used for deduped storage in production. So perhaps we should remove this check...

GlenDC: When benchmarking I'm fairly sure it's going to be pretty invisible if you test it with a local redis server. Probably only becomes really clear when benchmarked with remote redis servers.

PR #382, opened 2017-08-09
Adds 0-log to Config

### LINKED ISSUE #300: Add For-Orchestrator-Failure (stderr) logger
Part of the self-healing as described in https://github.com/zero-os/0-orchestrator/blob/master/specs/selfhealing/storage.md

In certain scenarios the nbdserver has to communicate with the 0-orchestrator:

+ a tlog server crashed;
+ a storage server crashed, rebalancing of data is required;

In these cases it will do so by using a special logger, which will log the failure/success messages via stderr, which will then be picked up by the 0-watchdog and delegated over to the 0-orchestrator.

The proposed format by @FastGeert is:

```
FAILURE: {"type": "storageengine-failure", "data": "10.100.0.1:22005"}
```

and

```
SUCCESS: {"type": "storageengine-rebalance", "data": "10.100.0.1:22005"}
```

The proposed format might not be final, but it should at the very least serve as a starting point.

---

Note: as this feature is going to be used by all 0-services, we might want to move this into a more common repo, perhaps even 0-orchestrator, but for now we could simply have it as a new root module in the 0-Disk repo.

### LINKED ISSUE #363: notify 0-Orchestrator when a hot-reloaded config is invalid
Whenever a config which is hot reloaded because of an update trigged by the 0-Orchestrator, results in an invalid config for the vdisk in question, we ignore that config for usage, as it would break our mounted vdisk.

While this is fine, we do also need to notify the 0-Orchestrator, so its aware of this failure and can fix it with a correction (if possible). Make sure to add to the failure the key of the config, and any other useful information, that might allow the 0-Orchestrator to auto-fix the issue or handle it in some way.

This issue can be resolved together with #300.

### LINKED ISSUE #370: notify 0-orchestrator when etcd (config) cluster is down
The 0-orchestrator should be handled in case the etcd (config) cluster is down, as this would prevent us from updating (see: hotreloading) any watched config values (although I guess if the cluster is down they are no longer watched). But more importantly, it would prevent us from mounting a vdisk, as the configuration comes from etcd.

Sooo... We should notify 0-orchestrator about this, using the logger described in #300, such that it can make the given cluster work again, or reboot the nbdserver using a newly given config cluster.

Until 0-orchestrator has resolved this notification, no new vdisks can be mounted and no configs of active vdisks can be updated.

chrisvdg:
Implements 0-log to log failing config for 0-Orchestrator
closes #363 , closes #300 , closes #370

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -355,6 +355,10 @@
 			"Rev": "2243d714d6c94951d8ccca8c851836ff47d401c9"
 		},
 		{
+			"ImportPath": "github.com/zero-os/0-log",
+			"Rev": "9173ce07ea291ea5c99897e290394ee4122fee81"
+		},
+		{
 			"ImportPath": "golang.org/x/net/http2",
 			"Rev": "b3756b4b77d7b13260a0a2ec658753cf48922eac"
 		},
--- a/config/config_api.go
+++ b/config/config_api.go
@@ -126,8 +126,14 @@ func ReadConfig(source Source, id string, keyType KeyType) ([]byte, error) {
 	if err == ErrConfigUnavailable {
 		source.MarkInvalidKey(configKey, "")
 	} else if err == ErrSourceUnavailable {
-		// TODO: notify 0-Orchestrator that the etcd cluster is down
-		// see: https://github.com/zero-os/0-Disk/issues/370
+		log.Errorf("couldn't fetch config %v: %v", configKey, err)
+		if source.Type() == "etcd" {
+			log.Broadcast(
+				log.StatusClusterTimeout,
+				log.SubjectETCD,
+				source.SourceConfig(),
+			)
+		}
 	} else if _, isInvalid := err.(*InvalidConfigError); isInvalid {
 		source.MarkInvalidKey(configKey, "")
 	}
--- a/config/source.go
+++ b/config/source.go
@@ -40,6 +40,11 @@ type Source interface {
 	// in case the invalid state is related to
 	// it being used for a specific vdisk.
 	MarkInvalidKey(key Key, vdiskID string)
+	// SourceConfig returns this source configuration.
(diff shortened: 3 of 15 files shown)


chrisvdg:
Better suggestions for the naming are always welcome

config/logger.go
```diff
@@ -0,0 +1,171 @@
...
+
+// MessageState represents the MessageSubject state
+type MessageState uint8
+
+// states
+const (
+	// StateSuccess represents a success state
+	StateSuccess MessageState = iota
+	// StateRebalanced represents a rebalanced state
+	StateRebalanced
```
GlenDC: You can delete this, as this is not used anywhere for now, AFAIK.

config/logger.go
```diff
@@ -0,0 +1,171 @@
...
+	SubStorageengine MessageSubject = iota
+	// SubETCDCluster represents an etcd cluster subject
+	SubETCDCluster
+	// SubTlogRPC represents a Tlog RPC subject
+	SubTlogRPC
+)
+
+// subjects
+const (
+	subStorageengineStr = "storageengine"
```
GlenDC: find a way to separate your words, either caps, or underscores, but choose something, and be consistent with the rest of the ecosystem.

config/logger.go
```diff
@@ -0,0 +1,171 @@
...
+
+// logFormat represents a json struct that can be used by 0-Log
+// to notify the 0-Orchestrator of errors
+type logFormat struct {
+	Type string `json:"type"`
+	Data string `json:"data"`
+}
+
+// SendJSONLog send a logFormat formated JSON message through 0-log with provided type and data
+func SendJSONLog(t string, d string) error {
```
GlenDC: you can delete this function, it's not needed.

GlenDC:
Also why did you add this to the config module? The log module is a much better fit imho. It's about logging! Not about configuration o.o

PR #393, opened 2017-08-15
do not restore a vdisk if it still exists

### LINKED ISSUE #392: do not restore vdisk while it still exists in the ARDB cluster
Currently you can restore a vdisk using the `zeroctl restore vdisk` command, even though its (meta)data still exists on the ARDB cluster.

This is a problem when you use the `zeroctl restore vdisk` command to go back in time (see: rollback), as you would still have newer data available on the ARDB server, which will give undefined behavior, data-wise.

We can fix this by first checking if the vdisk already/still exists, and if it does, exit with an error.

However, we should also add a `-f, --force` optional flag, such that when it is given, we'll delete all the existing (meta)data on the cluster for the given vdisk prior to restoring the vdisk in question, if that would be the user's intend.

GlenDC:
fixes #392

### DIFF
--- a/zeroctl/cmd/restore/vdisk.go
+++ b/zeroctl/cmd/restore/vdisk.go
@@ -8,6 +8,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/zero-os/0-Disk/config"
 	"github.com/zero-os/0-Disk/log"
+	"github.com/zero-os/0-Disk/nbd/ardb/storage"
 	"github.com/zero-os/0-Disk/tlog/tlogclient/decoder"
 	"github.com/zero-os/0-Disk/tlog/tlogclient/player"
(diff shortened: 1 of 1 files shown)


zeroctl/cmd/restore/vdisk.go
```diff
@@ -75,6 +82,60 @@ func restoreVdisk(cmd *cobra.Command, args []string) error {
 	return err
 }
 
+// checkVdiskExists checks if the vdisk in question already/still exists,
+// and if so, and the force flag is not specified, delete the (meta)data.
```
iwanbk: > and the force flag is not specified, delete the (meta)data.

is **NOT** specified?

I think it is going to be deleted if the flag is specified

GlenDC: Oh you're right, that comment should be `is specified`.

PR #396, opened 2017-08-16
adds http support to ServiceAddress validation

### LINKED ISSUE #394: config : `serviceaddress` tag doesn't recognize http server adddress
```go
 invalid ZeroStorClusterConfig: Address: http://127.0.0.1:39555 does not validate as serviceaddress
```

chrisvdg:
closes #394

### DIFF
--- a/config/validation.go
+++ b/config/validation.go
@@ -18,11 +18,17 @@ func IsServiceAddress(str string) bool {
 		return ok && pt == valid.Unix
 	}
 
+	if strings.HasPrefix(str, httpPrefix) || strings.HasPrefix(str, httpsPrefix) {
+		return valid.IsURL(str)
+	}
+
 	return valid.IsDialString(str)
 }
 
 const (
 	unixPrefix    = "unix://"
+	httpPrefix    = "http://"
+	httpsPrefix   = "https://"
 	unixFileStart = len(unixPrefix) - 1
 )
 
--- a/config/validation_test.go
+++ b/config/validation_test.go
@@ -17,6 +17,11 @@ func TestIsServiceAddress(t *testing.T) {
 		"[2001:db8:0:1:1:1:1:1]:123",
 		"http://etcdserver.com:123",
 		"https://etcdserver.com:123",
+		"http://127.0.0.1:123",
+		"https://127.0.0.1:123",
+		"http://etcdserver.com:123/somepath",
+		"https://etcdserver.com:123/somepath",
+		"http://etcdserver.com:123/somepath#someid",
 	}
 	for _, validCase := range validCases {
 		assert.Truef(IsServiceAddress(validCase), "%v", validCase)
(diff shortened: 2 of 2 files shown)


config/validation.go
```diff
@@ -18,11 +18,17 @@ func IsServiceAddress(str string) bool {
 		return ok && pt == valid.Unix
 	}
 
+	if strings.HasPrefix(str, httpPrefix) || strings.HasPrefix(str, httpsPrefix) {
```
GlenDC: use a regex which checks for `^https?://`, so you prevent having to scan twice through the string in order to find it if it's https or not.

config/validation_test.go
```diff
@@ -15,6 +15,8 @@ func TestIsServiceAddress(t *testing.T) {
 		"unix://42",
 		"127.0.0.1:3000",
 		"[2001:db8:0:1:1:1:1:1]:123",
+		"http://etcdserver.com:123",
+		"https://etcdserver.com:123",
```
GlenDC: also add some tests with some deeper paths after the port, it will pass as the `IsURL` function accepts this at the moment (and probably always will)(, but better to add it to protect against future changes.

config/validation_test.go
```diff
@@ -26,6 +28,7 @@ func TestIsServiceAddress(t *testing.T) {
 		"localhost:foo",
 		"unix:/foo",
 		"unix:boo",
+		"http://,foo.com",
```
GlenDC: add some more subtle typo errors as well.

config/validation.go
chrisvdg: I don't like using regex as I've been taught they are expensive and I don't find they are very clean (for people who don't know regex well).
As I wanted to test how expensive they are myself I've made a quick little test (results are at the bottom)
https://gist.github.com/chrisvdg/2d05b1f2932b1c4c0f4a9dd623cf79a7
sidenote: when `str` is set to "http://" the double string check takes half as much time.

GlenDC: Your first point is in general true, your second point is a weak excuse. Regex is easy enough, especially a regex as simple as the one suggested, so I'm going to ignore your second point.

Surprisingly your test does check out, so you catch me off guard here. The golang regex implementation doesn't support back-tracing, so it only ever scans forward, so I'm surprised that still makes it that more expensive than a linear non-regex scan.

As I don't really have much more time to dive deeper into this, and because this really isn't a critical path, so optimization here is really not a necessity, I'll accept your PR as it is.

chrisvdg: Yeah, the second point is much more personal opinion

GlenDC: For people who don't know regexp very well, and they want to learn it quickly I can recommend going through all levels off https://regexcrossword.com. Pretty easy to bootstrap yourself like that. Because there are plenty of situations where a solution without the usage of regexp would be pretty cumbersome.

GlenDC:
@iwanbk this PR seems to be ready for merging in, feel free to review it yourself & merge it in when you feel ready for it.

iwanbk:
thanks @chrisvdg & @GlenDC

PR #397, opened 2017-08-17
WIP: Adds stringer implementations to config.Key and config.KeyType

### LINKED ISSUE #389: Make config.Key and config.KeyType comply to Stringer interface
Provide `String` methods for the `Key` and `KeyType` structs of the `0-Disk/config` module.

For the `KeyType`'s `String` method I propose that we go for the following format: `<TypeString>(<ID>)`, but if you have a better idea, feel free to go for your idea instead.

This should make log messages more clear, as for now it will log stuff like `{mycluster 3}`, which forces the reader of logs to look up each number each time, which isn't a very good use of my time.

chrisvdg:
closes #389

### DIFF
--- a/config/source.go
+++ b/config/source.go
@@ -63,6 +63,10 @@ type Key struct {
 	Type KeyType
 }
 
+func (k Key) String() string {
+	return fmt.Sprintf("%s(%s)", k.Type, k.ID)
+}
+
 // KeyType defines the type of a key,
 // which together with the ID makes it unique.
 type KeyType uint8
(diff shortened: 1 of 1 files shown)


chrisvdg:
KeyType string representations are placeholder, suggestions are welcome

config/source.go
```diff
@@ -63,6 +63,10 @@ type Key struct {
 	Type KeyType
 }
 
+func (k Key) String() string {
+	return fmt.Sprintf("%s(%s)", k.Type, k.ID)
```
GlenDC: Yeah that's what I had in mind, for now that's as good as a format as any. This is anyhow just for debugging purposes so as long as it shows all the information clearly, I'm OK with it.

PR #406, opened 2017-08-21
Adds statistics logger

### LINKED ISSUE #385: integrate 0-log statistics log into 0-Disk/log
A small layer around the 0-log stats logger functionality, to encapsulate that logic, and to make it a bit easier to use, serving 0-Disk specific needs.

Blocked by https://github.com/zero-os/0-log/issues/7.

chrisvdg:
closes #385

### DIFF
--- a/docs/log.md
+++ b/docs/log.md
@@ -114,64 +114,19 @@ This message is send in the hope that the config can be made valid by receiving 
 
 ## Broadcast statistics 
 
-The `BroadcastStatistics` function in the `0-Disk/log` package logs statistical messages using the [0-Log library][zeroLog] to broadcast messages for the [0-core log monitor][zeroCoreLogMonitor] using the [Statistics Log message format spec][StatLogSpec].
+The `BroadcastStatistics` function in the `0-Disk/log` package logs statistical messages using the [0-Log library][zeroLog] to broadcast messages for the [0-core log monitor][zeroCoreLogMonitor] using the [Statistics Log message format spec][StatLogSpec]. The broadcasted statistics messages are send at [log level 10 (statistics/monitoring message)][loglevels]. 
 
 More in-depth information about the actual implementation in 0-Disk can be found in [the log module Godocs][zeroDiskLogGodcs].
 
-### Message format
+### Logged statistics
 
-All broadcasted statistics messages are sent using [level 10 (loglevel Statistics/monitoring)][loglevels] and takes a `zerolog.MsgStatistics` struct with the following fields:
-
-* Key (`string`): represents the key for the statistic
-* Value (`float`): represents the data value of the statistic
-* Operation (`zerolog.AggregationType`): represents the aggregation operation for the statistic
-* Tags(`zerolog.MetricTags`): represents additional tags for the statistic
-
-More indepth information about 0-log can be found on the [0-log documentation][zerolog]
-
-The `BroadcastStatistics` function takes the following arguments:
-* vdiskID (`string`): represents the ID of the vdisk where the statistics are from.  
-    Used to build the `0-log` `Key`.
-* key (`StatisticsKey`): represents of what the statistic is from.  
-    Used to build the `0-log` `Key`.
-* value (`float`): Used for the `0-log` `Value`.
-* op (`AggregationType`): Used for the `0-log` `Operation`.
-* tags (`MetricTags`, optional): Used for the `0-log` `Tags`.
-
-These values will be put in a `zerolog.MsgStatistics` struct to be logged. 
-
-The `vdiskID` and `key` are used to format the `MsgStatistics` key to the [monitor metrics spec][monMetricsSpec] specific for 0-Disks.
-
-Wrappers for `zerolog.AggregationType` (`AggregationType`) and `zerolog.MetricTags` (`MetricTags`) are provided so there is no need to directly access the 0-log lib anywhere else within `0-Disk`.
-
-An error can be returned in following cases:
-* `vdiskID` is empty (`ErrNilVdiskID`)
-* `StatisticsKey` (`Key`) is invalid (`ErrInvalidStatisticsKey`)
-* `AggregationType` (`zerolog.MsgStatistics.Operation`) is invalid (`zerolog.ErrInvalidAggregationType`)
-
-### Usage
-```go
-// without tags
-BroadcastStatistics("vdisk1", StatisticsKeyIOPSWrite, 1.234, AggregationAverages, nil)
-// outputs: 10::vdisk.iops.write@virt.vdisk1:1.234000|A
-
-// with tags
-tags := MetricTags{
-    "foo":   "world",
-    "hello": "bar",
-}
-BroadcastStatistics("vdisk2", StatisticsKeyTroughputRead, 2.345, AggregationDifferentiates, tags)
-// outputs: 10::vdisk.throughput.read@virt.vdisk2:2.345000|D|foo=world,hello=bar
-```
+    // TODO: describe format of statistics broadcasted + services that broadcast them
 
 [zeroLog]: https://github.com/zero-os/0-log/
-[zeroStor]: https://github.com/zero-os/0-stor/
 [loglevels]: https://github.com/zero-os/0-log/blob/master/README.md#supported-log-levels
-[sourceinterface]: https://godoc.org/github.com/zero-os/0-Disk/config/#Source
 [zeroCoreLogMonitor]: https://github.com/zero-os/0-core/blob/master/docs/monitoring/README.md#monitoring
 [zeroOrchestrator]: https://github.com/zero-os/0-orchestrator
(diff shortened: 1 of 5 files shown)


log/broadcast_stats.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	return statkey.String() + "@virt." + vdiskID, nil
+}
+
+// StatisticsKey represents a statistics key
+type StatisticsKey string
+
+// StatisticsKey options
+const (
+	StatisticsKeyIOPSRead       = StatisticsKey("vdisk.iops.read")
```
GlenDC: Better turn this into an enum, is cheaper:

+ Strings are passed by value, so each time we pass it to a statics log call, we'll be copying an entire string on the stack;
+ Validation is also a lot cheaper, as you'll only have to compare 1 byte (uint8), compared to an entire string comparison;

The string function will get a bit more expensive, but not that much, as it's a very simply switch case. You could even use `golang.org/x/tools/cmd/stringer` as described in https://blog.golang.org/generate, should you want. But writing it by hand is fine as well, I'm fine either way. So far in this repo we have always written it by hand, but it's all the same for me.

log/broadcast_stats.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"errors"
+
+	"github.com/zero-os/0-log"
+)
+
+var (
+	// ErrNilVdiskID represents an error where an empty vdiskid was provided
+	ErrNilVdiskID = errors.New("empty statistics ID")
```
GlenDC: error message says a different thing than the error name. Better to have the message as `"unexpected nil vdiskID"` or something similar.

log/broadcast_stats.go
```diff
@@ -0,0 +1,82 @@
...
+func BroadcastStatistics(vdiskID string, key StatisticsKey, value float64, op AggregationType, tags MetricTags) error {
+	k, err := createKey(vdiskID, key)
+	if err != nil {
+		return nil
+	}
+
+	msg := zerolog.MsgStatistics{
+		Value:     value,
+		Key:       k,
+		Operation: zerolog.AggregationType(op),
```
GlenDC: Once 1.9 comes out we can create a separate file for this function `pre1.9` and `1.9` as I'm fairly sure this can be done without casting since 1.9, thanks to type aliasing. But let's worry about that once 1.9 releases officially.

docs/log.md
```diff
@@ -112,12 +112,66 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
 
 This message is send in the hope that the config can be made valid by receiving an(other) update from the [0-Orchestrator][zeroOrchestrator].
 
+## Broadcast statistics 
+
+The `BroadcastStatistics` function in the `0-Disk/log` package logs statistical messages using the [0-Log library][zeroLog] to broadcast messages for the [0-core log monitor][zeroCoreLogMonitor] using the [Statistics Log message format spec][StatLogSpec].
+
+More in-depth information about the actual implementation in 0-Disk can be found in [the log module Godocs][zeroDiskLogGodcs].
+
+### Message format
```
GlenDC: `Message Format` and `Usage` don't belong in this documentation. This is stuff that belongs in the Godocs documentation, not here. This is external documentation, very high level, they don't need to care about internal details like structure format and function APIs.

What should be described here:

+ Statistics we log;
+ For each statistic we log:
  + How we format them (OP Type, tags, etc...);
  + Which services log each statistic;

For now I guess you don't know that yet, so that's fine. But hey, feel free to move this information into Godocs format. More comments about that below...

docs/log.md
```diff
@@ -112,12 +112,66 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
...
+
+All broadcasted statistics messages are sent using [level 10 (loglevel Statistics/monitoring)][loglevels] and takes a `zerolog.MsgStatistics` struct with the following fields:
+
+* Key (`string`): represents the key for the statistic
```
GlenDC: This documentation can be described for each property in-lined with the struct properties itself.

docs/log.md
```diff
@@ -112,12 +112,66 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
...
+* Value (`float`): represents the data value of the statistic
+* Operation (`zerolog.AggregationType`): represents the aggregation operation for the statistic
+* Tags(`zerolog.MetricTags`): represents additional tags for the statistic
+
+More indepth information about 0-log can be found on the [0-log documentation][zerolog]
+
+The `BroadcastStatistics` function takes the following arguments:
```
GlenDC: should be inline documentation for the function itself, although not as explicit as you are describing here. Some properties are definitely obvious enough from their name alone.

docs/log.md
```diff
@@ -112,12 +112,66 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
...
+    Used to build the `0-log` `Key`.
+* value (`float`): Used for the `0-log` `Value`.
+* op (`AggregationType`): Used for the `0-log` `Operation`.
+* tags (`MetricTags`, optional): Used for the `0-log` `Tags`.
+
+These values will be put in a `zerolog.MsgStatistics` struct to be logged. 
+
+The `vdiskID` and `key` are used to format the `MsgStatistics` key to the [monitor metrics spec][monMetricsSpec] specific for 0-Disks.
+
+Wrappers for `zerolog.AggregationType` (`AggregationType`) and `zerolog.MetricTags` (`MetricTags`) are provided so there is no need to directly access the 0-log lib anywhere else within `0-Disk`.
```
GlenDC: This can be added as inline documentation for the type definitions.

docs/log.md
```diff
@@ -112,12 +112,66 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
...
+
+An error can be returned in following cases:
+* `vdiskID` is empty (`ErrNilVdiskID`)
+* `StatisticsKey` (`Key`) is invalid (`ErrInvalidStatisticsKey`)
+* `AggregationType` (`zerolog.MsgStatistics.Operation`) is invalid (`zerolog.ErrInvalidAggregationType`)
+
+### Usage
+```go
```
GlenDC: Can be added as Godoc examples inline with the function.

docs/log.md
```diff
@@ -112,12 +112,66 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
...
+* op (`AggregationType`): Used for the `0-log` `Operation`.
+* tags (`MetricTags`, optional): Used for the `0-log` `Tags`.
+
+These values will be put in a `zerolog.MsgStatistics` struct to be logged. 
+
+The `vdiskID` and `key` are used to format the `MsgStatistics` key to the [monitor metrics spec][monMetricsSpec] specific for 0-Disks.
+
+Wrappers for `zerolog.AggregationType` (`AggregationType`) and `zerolog.MetricTags` (`MetricTags`) are provided so there is no need to directly access the 0-log lib anywhere else within `0-Disk`.
+
+An error can be returned in following cases:
```
GlenDC: `will be`*, as this is 100% for sure in each of these scenarios;

You can add this as part of the inline godoc documentation for the function.

GlenDC: Note sure if there is a format (or guideline) for describing error cases for a function, but if there is, you should follow it.

log/broadcast_stats.go
chrisvdg: Did some quick testing on this to see the expensiveness for myself and you're right, it would start making more sense to use the strings if you're doing more than 5 String calls per Validation call (for the values that we have), but in this case it's 1/1, which makes using uint8 almost half as expensive as  using string.

https://gist.github.com/chrisvdg/0da46534fc1c377b351f6ab5880fed8c

log/broadcast_stats.go
```diff
@@ -0,0 +1,113 @@
...
+	case StatisticsKeyTroughputWrite:
+		return statisticsKeyTroughputWriteStr
+	default:
+		return statisticsKeyNilStr
+	}
+}
+
+// Validate validates the StatisticsKey
+func (sk StatisticsKey) Validate() error {
+	switch sk {
```
GlenDC: When your enum values are just sequential `iota` values you could improve this function by just checking the bounds of the sk value as in

```go
if sk < StatisticsKeyIOPSRead || sk > StatisticsKeyTroughputWrite {
    return ErrInvalidStatisticsKey
}

return nil
```

log/broadcast_stats.go
```diff
@@ -0,0 +1,113 @@
...
+	StatisticsKeyTroughputWrite
+)
+
+// StatisticsKey string representations
+const (
+	statisticsKeyIOPSReadStr       = "vdisk.iops.read"
+	statisticsKeyIOPSWriteStr      = "vdisk.iops.write"
+	statisticsKeyTroughputReadStr  = "vdisk.throughput.read"
+	statisticsKeyTroughputWriteStr = "vdisk.throughput.write"
+	statisticsKeyNilStr            = ""
```
GlenDC: No need to make a constant for this, you can simply return the empty string where needed.

PR #400, opened 2017-08-18
0-stor tlog integration

iwanbk:
Fixes #311 

Notable changes:
- use 0-stor (of course)
- remove testcgo : 0-stor is pure Go
-  Fixes #381 rework vdisk completion https://github.com/zero-os/0-Disk/issues/381#issuecomment-323253812
- ~~split the test into more jobs https://github.com/zero-os/0-Disk/issues/311#issuecomment-322694804~~ cancelled by 2f6e27d37cbd6f33a8d3e706ef6b6bec5d3a45fb, 27e2d383f6e63dd64eab3aa329f3602de420a5cd
- Fixes #374 by removing the cgo test
- update docs.

PR #410, opened 2017-08-23
backup feature (incl. zeroctl export/import commands)

### LINKED ISSUE #317: Import / Export virtual machines
supported backend
- ftp (optionally sftp)

definitions
- dedupe_map
    - capnp serialized list location/blake_hash
    - is done per 128 KB LBA block so 10,aabbccddeeffgghh is block 128 KB at LBA pos 128*10/4
    - 20GB data needs 3.5 MB of size in dedupe_map

export
- arguments: 
    - ftp url
    - encryption key
    - optional: blocksize (std 128 KB)
    - optional: which compression to use std lz4 
    - optional: name (if name not given then epoch)
- load dedupe_map
- walk over LBA storage space per 128 KB, check against dedupemap to see if changes
   - walk in parallel over the dedupe_map this allows fast comparison
- build in parallel the output dedupe_map
   - if not change, copy the original entry
   - if change write the new value
   - if new insert new value
- for each new found 128 KB block
   - compress block using [lz4](https://github.com/pierrec/lz4) or [xz](https://github.com/ulikunitz/xz)
   - encrypt the block with the encr key
   - add 4 byte CRC to end of file
   - upload block to $storagelocation/$aa/$bb/$ccddeeffgghh
   - at end upload the compressed/encr dedupe_map to $storagelocation/backups/$name

restore
- get dedupe_map
- ... to be completed

GlenDC:
+ fixes #317;
+ adds backup module which allows for exporting/importing of vdisks;
+ add import/export zeroctl commands, which act as front-ends for the backup module;
+ add unit tests to ensure the backup module work as intended;
+ add lede img blocks gzipped in a file, used for production-comparable tests and compression benchmarks;
+ fix bug in storage.ListDedupedBlockIndices;
+ move InMemoryStorage from nbdserver module to storage module, so we can use it in other modules for testing purposes as well;
+ vendors new dependencies required by backup module;
+ adds documentation for new backup module and import/export zeroctl commands;


GlenDC:
Backup feature is ready for review and merging.

GlenDC:
All unit tests pass locally and in Travis. However travis results are a bit delayed as the OSX backlog is super long due to the recent outage. However if you check the detail you can see that all the unit tests and race tests are already passing and complete with success.

iwanbk:
>  GlenDC requested a review from iwanbk 3 hours ago

Can only check it tomorrow (thursday)

GlenDC:
Rebased master into this branch, as I had to resolve some conflicts that the merging of the 0-stor integration (PR #400) caused.

GlenDC:
Rebased master after merging PR #406 (into master). This feature branch is now the only open PR, and up to date with master. All tests still pass, and the feature is still intact. Still ready for review! 👍

nbd/ardb/backup/storage_ftp_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+func TestMinimalFTPStorageConfigToString(t *testing.T) {
+	assert := assert.New(t)
+
+	validCases := []struct {
+		Input, Output string
+	}{
+		{"foo", "ftp://foo:21"},
+		{"foo/bar/baz", "ftp://foo:21/bar/baz"},
+		{"foo:22", "ftp://foo:22"},
```
iwanbk: don't we also need to support `sftp`?

iwanbk:
The usage of `bencode` is also very interesting, i had minimal knowledge about this, only knows that `bittorrent` uses it. Some questions about it which i can't found the answer on https://github.com/zero-os/0-Disk/issues/317#issuecomment-322015388

- it only has four data types, is it really OK for potential future changes?
- any idea why it is not widely used?(it is only my assumption, not sure about the fact)
- why not using more common encoding like protobuf?

GlenDC:
> it only has four data types, is it really OK for potential future changes?

Should be fine to store a deduped map, not sure what more we would ever need, as it just about mapping indices with hashes. If the deduped map would ever change in how we store it we would anyhow have 2 different formats, whether or not we use the same encoding protocol. It must be noted though that thanks to the bencoding lib we use we can use more than just the string primary type: https://github.com/zeebo/bencode/blob/master/encode.go#L74-L236 I think that should be enough for all our purposes here...

> any idea why it is not widely used?

dunno, I'm pretty sure it's not too well known among programmers for starters. But for more generic applications the bencoding format might indeed be too limited. Even though it has plenty of type support for many applications. Honestly not sure why it's not more widely used. I was for example also interested in it potential use for rivine, as that project is now using no encoding protocol at all and simply serializing structs directly...

> why not using more common encoding like protobuf?

First I was using Capnp as suggested by kds, but that was really slow. Protobuf is probably faster, but it's still a more complex protocol so I would be surprised that it's as fast as bencode to use. Bencoding was made for speed and simplicity, so I preferred that. However the backup module doesn't stand or fall with the encoding/decoding of the deduped map. We can easily change to protobuf or any other format if that is what the team wants, it's completely isolated.

docs/nbd/backup.md
```diff
@@ -0,0 +1,69 @@
...
+
+The export process is straightforward and can be summarized as follows:
+
+1. All block indices are collected from the ARDB cluster, ordered from smallest to biggest;
+2. The deduped map is loaded if it already existed, and newly created if it didn't;
+2. All content is written in a parallel and streamlined fashion:
+    1. All blocks are read in parallel from the ARDB storage;
+    2. The blocks are ordered in a single goroutine and sized to the correct export block size;
+    3. The newly sized blocks are all compressed, encrypted and stored in parallel;
+    4. If the hash of the encrypted content's hash is alreayd mapped to the block index in question, it is not written to the storage server, as it is assumed in such case that the block already exists on the storage server;
```
GlenDC: Will fix it later today, while also adding some more sections and a few graphs.

iwanbk:
sounds good.

How about the sftp support?

GlenDC:
> How about the sftp support?

We use https://github.com/secsy/goftp for FTP as recommended by @FastGeert.

It does not support SFTP (FTP over SSH), it only supports FTPS (using TLS). Right now I didn't add support for enabling it via the configuration though. Is FTPS what we want to use? Or do we want SFTP instead?

Let me know, and I'll add support for that. Although I'm not sure how you want to configure FTPS, would need to think about an easy way to configure that.

iwanbk:
>  Is FTPS what we want to use? Or do we want SFTP instead?

No idea about this, @zaibon ?

zaibon:
SFTP will give us issue cause then you need to authorize the key of the node that export the vdisk on the FTP server of the client. This become even more complex when want to import between 2 environment. We didn't find an easy way to solve this key distribution problem. So FTP with username/password is good enough for now. 

Same decision has been take for import/export of container. so we stay consistent.

PR #421, opened 2017-08-28
Change tlog aggregation timestamp to UTC

### LINKED ISSUE #407: Might need to change timestamp/epoch to UTC
Original discussion 
https://github.com/zero-os/0-Disk/pull/400#discussion_r134603721

iwanbk:
Fixes #407

### DIFF
--- a/docs/zeroctl/commands/restore.md
+++ b/docs/zeroctl/commands/restore.md
@@ -11,12 +11,12 @@ Usage:
 Flags:
       --config string              zeroctl config file (default "config.yml")
       --data-shards int            data shards (K) variable of erasure encoding (default 4)
-      --end-timestamp uint         end timestamp in nanosecond(default 0: until the end)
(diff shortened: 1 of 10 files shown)


docs/zeroctl/commands/restore.md
```diff
@@ -11,12 +11,12 @@ Usage:
 Flags:
       --config string              zeroctl config file (default "config.yml")
       --data-shards int            data shards (K) variable of erasure encoding (default 4)
-      --end-timestamp uint         end timestamp in nanosecond(default 0: until the end)
+      --end-timestamp uint         end UTC timestamp in nanosecond(default 0: until the end)
```
GlenDC: If these values are stored somewhere I suppose it's no issue, but I've always wondered why we don't take into a formatted string such as an edited version of `RFC3339Nano`, where we could make any non given information optional by `Now()`-filling it. And where if the timezone wouldn't be given UTC could be assumed to be it.

PR #424, opened 2017-08-29
Notify zerostor down

### LINKED ISSUE #371: tlogserver: notify 0-Orchesrator when 0-Stor server is down
In case a 0-stor server used by the tlogserver for a vdiskID is unavailable due to whatever critical error, we should notify the 0-Orchestrator about this, using the logger described in #300, so it can do something about it.

This issue is blocked until #300 is resolved.

iwanbk:
Fixes #371

### DIFF
--- a/log/broadcast.go
+++ b/log/broadcast.go
@@ -38,6 +38,8 @@ func (s MessageSubject) String() string {
 		return subjectETCDStr
 	case SubjectTlog:
 		return subjectTlogStr
+	case SubjectZeroStor:
+		return subjectZeroStorStr
 	default:
 		return subjectNilStr
 	}
@@ -61,14 +63,17 @@ const (
 	SubjectETCD
 	// SubjectTlog identifies the messages has to do with tlog
 	SubjectTlog
+	// SubjectT0Stor identifies the messages has to do with 0-stor
+	SubjectZeroStor
 )
 
 // subjects
 const (
-	subjectStorageStr = "ardb"
-	subjectETCDStr    = "etcd"
-	subjectTlogStr    = "tlog"
-	subjectNilStr     = ""
+	subjectStorageStr  = "ardb"
+	subjectETCDStr     = "etcd"
+	subjectTlogStr     = "tlog"
+	subjectZeroStorStr = "0-stor"
+	subjectNilStr      = ""
 )
 
 // MessageStatus represents the status code
@@ -77,6 +82,7 @@ type MessageStatus uint
 
 // status codes
 const (
+	StatusUnknownError   MessageStatus = 400
 	StatusClusterTimeout MessageStatus = 401
 	StatusServerTimeout  MessageStatus = 402
 	StatusInvalidConfig  MessageStatus = 403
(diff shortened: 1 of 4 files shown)


log/broadcast.go
```diff
@@ -38,6 +38,8 @@ func (s MessageSubject) String() string {
 		return subjectETCDStr
 	case SubjectTlog:
 		return subjectTlogStr
+	case Subject0Stor:
```
GlenDC: Please call this `SubjectZeroStor`, having a numeric character in the middle of a name makes me feel itchy...

log/broadcast.go
```diff
@@ -77,9 +82,10 @@ type MessageStatus uint
 
 // status codes
 const (
-	StatusClusterTimeout MessageStatus = 401
-	StatusServerTimeout  MessageStatus = 402
-	StatusInvalidConfig  MessageStatus = 403
+	StatusClusterTimeout     MessageStatus = 401
+	StatusServerTimeout      MessageStatus = 402
+	StatusInvalidConfig      MessageStatus = 403
+	StatusServerUnknownError MessageStatus = 404
```
GlenDC: Just call this `StatusUnknownError`, as these status codes are meant to be as generic as possible. You can also give it the value of `400`, instead of `404`. This way it's the first kind of Error status code, useable for any situation where the actual error us unknown.

log/broadcast.go
iwanbk: fixed

log/broadcast.go
iwanbk: fixed

log/broadcast.go
```diff
@@ -61,14 +63,17 @@ const (
 	SubjectETCD
 	// SubjectTlog identifies the messages has to do with tlog
 	SubjectTlog
+	// SubjectT0Stor identifies the messages has to do with 0-stor
```
GlenDC: You forgot to fix this comment.

log/broadcast.go
```diff
@@ -61,14 +63,17 @@ const (
...
 // subjects
 const (
-	subjectStorageStr = "ardb"
-	subjectETCDStr    = "etcd"
-	subjectTlogStr    = "tlog"
-	subjectNilStr     = ""
+	subjectStorageStr  = "ardb"
+	subjectETCDStr     = "etcd"
+	subjectTlogStr     = "tlog"
+	subjectZeroStorStr = "0-stor"
```
GlenDC: are you sure we don't want to call this zerostor instead? Would keep the character set limited to lowercase alphabetical characters. Not sure if this is important, but just observing.

log/broadcast.go
iwanbk: my bad

log/broadcast.go
iwanbk: feel free to change it. Both are fine for  me

GlenDC: ok

log/broadcast.go
GlenDC: No worries, I fixed it already in a local PR that I'm opening soon.

PR #427, opened 2017-08-30
handle a timed out template ardb storage server

### LINKED ISSUE #369: nbdserver: handle template server failure
In case a template server used by a vdiskID is unavailable due to whatever critical error, we should notify the 0-Orchestrator about this, using the logger implemented in #300, so it can do something about it.

We can't do much more than just notifying the 0-orchestrator of this, as the data comes from the template server, so until this issue is resolved for us, there is nothing we can do, except to disable the template feature until the storage configuration used by that vdisk has been updated by the 0-Orchestrator.

NOTE:

+ once that failure happens, we should disable that server until the config has been reloaded, such that we only notify such a failure once, and not every time a read happens, as that would spam the 0-Orchestrator;
+ this disabling happens ideally individually for each server, rather than for the entire template cluster, as it is quite likely that even though one server is unavailable, other servers of that same template cluster are still fine;

GlenDC:
+ fixes #369;
+ part of 0-Disk's self-healing efforts;
+ also fixes some things and adds (external) documentation for merged PR #424;

### DIFF
--- a/nbd/ardb/storage/deduped.go
+++ b/nbd/ardb/storage/deduped.go
@@ -125,12 +125,16 @@ func (ds *dedupedStorage) Flush() (err error) {
 	return
 }
 
-func (ds *dedupedStorage) getDataConnection(hash zerodisk.Hash) (redis.Conn, error) {
+func (ds *dedupedStorage) getDataConnection(hash zerodisk.Hash) (ardb.Connection, error) {
 	return ds.provider.DataConnection(int64(hash[0]))
 }
 
-func (ds *dedupedStorage) getTemplateConnection(hash zerodisk.Hash) (redis.Conn, error) {
+func (ds *dedupedStorage) getTemplateConnection(hash zerodisk.Hash) (ardb.Connection, error) {
 	return ds.provider.TemplateConnection(int64(hash[0]))
+}
+
+func (ds *dedupedStorage) markTemplateConnectionInvalid(hash zerodisk.Hash) {
+	ds.provider.MarkTemplateConnectionInvalid(int64(hash[0]))
 }
 
 // getPrimaryContent gets content from the primary storage.
@@ -160,12 +164,33 @@ func (ds *dedupedStorage) getPrimaryOrTemplateContent(hash zerodisk.Hash) (conte
 	}
 
 	// try to fetch it from the template storage if available
-	content = func() (content []byte) {
+	content, err = func() (content []byte, err error) {
 		conn, err := ds.getTemplateConnection(hash)
 		if err != nil {
 			log.Debugf(
 				"content not available in primary storage for %v and no template storage available: %s",
 				hash, err.Error())
+			if status, ok := ardb.MapErrorToBroadcastStatus(err); ok {
+				log.Errorf("template server error for vdisk %s: %v", ds.vdiskID, err)
+				// broadcast the connection issue to 0-Orchestrator
+				cfg := conn.ConnectionConfig()
+				log.Broadcast(
+					status,
+					log.SubjectStorage,
+					log.ARDBServerTimeoutBody{
+						Address:  cfg.Address,
+						Database: cfg.Database,
+						Type:     log.ARDBTemplateServer,
+						VdiskID:  ds.vdiskID,
+					},
+				)
+				// mark template connection for the given block index invalid,
+				// so it remains marked invalid until next config reload.
+				ds.markTemplateConnectionInvalid(hash)
+			} else if err == ardb.ErrTemplateClusterNotSpecified {
+				err = nil
+			}
+
 			return
 		}
 		defer conn.Close()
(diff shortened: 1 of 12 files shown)


nbd/ardb/storage/deduped.go
```diff
@@ -160,48 +165,89 @@ func (ds *dedupedStorage) getPrimaryOrTemplateContent(hash zerodisk.Hash) (conte
...
 		if err != nil {
 			log.Debugf(
 				"content not available in primary storage for %v and no template storage available: %s",
 				hash, err.Error())
+			if netErr, ok := err.(net.Error); ok {
+				log.Errorf("template server error for vdisk %s: %v", ds.vdiskID, netErr)
+				// broadcast the connection issue to 0-Orchestrator
+				cfg := conn.ConnectionConfig()
+				log.Broadcast(
+					log.StatusServerTimeout,
```
iwanbk: I don't think that network error is always timeout error, there is also Temporary error, and maybe something else,not sure

nbd/ardb/storage/deduped.go
```diff
@@ -160,48 +165,89 @@ func (ds *dedupedStorage) getPrimaryOrTemplateContent(hash zerodisk.Hash) (conte
...
 			content = nil
-			log.Debugf(
-				"content for %v not available in primary-, nor in template storage: %s",
-				hash, err.Error())
+			if netErr, ok := err.(net.Error); ok {
+				log.Errorf("template server error for vdisk %s: %v", ds.vdiskID, netErr)
+				// broadcast the connection issue to 0-Orchestrator
+				cfg := conn.ConnectionConfig()
+				log.Broadcast(
+					log.StatusServerTimeout,
```
iwanbk: Same comment as before, network error doesn't always timeout error

nbd/ardb/storage/deduped.go
```diff
@@ -160,48 +165,89 @@ func (ds *dedupedStorage) getPrimaryOrTemplateContent(hash zerodisk.Hash) (conte
...
-				// we won't return error however, but just log it
-				log.Errorf("couldn't store template content in primary storage: %s", err.Error())
-				return
-			}
-
-			log.Debugf(
-				"stored template content for %v in primary storage (asynchronously)",
-				hash)
-		}()
+	// store template content in primary storage asynchronously
```
iwanbk: i know it is old logic, but why store it in async way? what if error happens?

nbd/ardb/storage/nondeduped.go
```diff
@@ -125,10 +126,30 @@ func (ss *nonDedupedStorage) getPrimaryOrTemplateContent(blockIndex int64) (cont
...
+	content, err = func() (content []byte, err error) {
 		// get a connection to a template data storage server, based on the modulo blockIndex
 		conn, err := ss.provider.TemplateConnection(blockIndex)
 		if err != nil {
+			if netErr, ok := err.(net.Error); ok {
+				log.Errorf("template server error for vdisk %s: %v", ss.vdiskID, netErr)
+				// broadcast the connection issue to 0-Orchestrator
+				cfg := conn.ConnectionConfig()
+				log.Broadcast(
+					log.StatusServerTimeout,
```
iwanbk: same comment as before

nbd/ardb/storage/nondeduped.go
```diff
@@ -137,32 +158,52 @@ func (ss *nonDedupedStorage) getPrimaryOrTemplateContent(blockIndex int64) (cont
...
-			log.Debugf(
-				"content for block %d (vdisk %s) not available in primary-, nor in template storage: %s",
-				blockIndex, ss.templateVdiskID, err.Error())
 			content = nil
+			if netErr, ok := err.(net.Error); ok {
+				log.Errorf("template server error for vdisk %s: %v", ss.vdiskID, netErr)
+				// broadcast the connection issue to 0-Orchestrator
+				cfg := conn.ConnectionConfig()
+				log.Broadcast(
+					log.StatusServerTimeout,
```
iwanbk: same comment as before

nbd/ardb/storage/deduped.go
GlenDC: We don't really care if an error happens in storing template content, next time we'll try again. The template content should remain available anyhow. So an error in this case shouldn't prevent getting the content again.

GlenDC: I mean we do care for internal purposes. So as soon as we have self-healing for this, we would obviously if needed report this to 0-Orchestrator, and if possible switch over to slave-cluster or w/e. But fetching the content was by now already successful, so there is no reason to return an error for `GetBlock`, as that did in fact happen successfully.

Storing the content in Primary content is more of a caching thing, well not really but that's how you could think of it. No reason that the user has to feel an error for that.

nbd/ardb/storage/deduped.go
GlenDC: Ah you mean that an error can _just_ be a timeout error, good one, OK I'll check it out and improve it.

GlenDC: No it seems that what i'm doing is pretty OK, as https://stackoverflow.com/a/31477004 confirms that checking for `net.Error` is good enough. The only error I'm still missing according to that answer is `io.EOF`, so will make a helper function to help with this.

nbd/ardb/storage/deduped.go
GlenDC: Oh nevermind, I now know what you mean. You are talking about the fact that is can also be a disconnect, or a temporary error. Ok got it, will fix it. Thanks 👍 👍

iwanbk:
Don't you think we need to retry in case of temporary error?

GlenDC:
We could, but what is temporary? How long should we wait? How many times should we retry? I have a feeling without any more knowledge besides "temporary error" that is just a guessing game...

But yes this was definitely floating in my head yesterday. If you have a well-constructed idea for this, please do share and we can take it from there.

GlenDC:
@iwanbk also don't forget that the redis client we use closes the connection in case of an error, so we can't simply retry using the same connection, we would have to open a new connection: https://github.com/garyburd/redigo/blob/master/redis/conn.go#L289-L299.

I also found https://blog.golang.org/error-handling-and-go, which does talk about handling transient network errors such as `Temporary`, but at the moment I wouldn't really know what a good error handling approach for it would be. I see some go repositories out there simply retrying 3 times, without a delay for example. Is it reasonable to have no delay? is 3 times enough? Or is it too much?

PR #429, opened 2017-08-30
WIP: adding IOPS and Throughput statistics logger

### LINKED ISSUE #398: Feed statistics into ledis via core-0 statistics logging
Statistics that we need to provide to core-0 :

- [x] Vdisk iops read, format on logging `vdisk.iops.read@virt.vdiskid`
- [x] Vdisk iops write, format on logging `vdisk.iops.write@virt.vdiskid`
- [x] Vdisk throughput read in KB/s, format on logging `vdisk.throughput.read@virt.vdiskid`
- [x] Vdisk throughput write in KB/s, format on logging `vdisk.throughput.write@virt.vdiskid`

vdiskid need to be replaced by the actual ID of the vdisk.

Questions : Should we also add the cluster id to the statistics ?

chrisvdg:
closes #398


chrisvdg:
First draft, created the PR to get feedback.
Also need to add tests.

docs/log.md
```diff
@@ -176,15 +176,36 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
...
+    * logged by: nbdserver
+    * broadcasts: `10::vdisk.iops.read@virt.<vdisk ID>:<value>|A`
+    * [0-core aggregation type][StatLogSpec]: Averages
+    * value unit: (partial) blocks per second
+ * IOPS write
+    * logged by: nbdserver
+    * broadcasts: `10::vdisk.iops.write@virt.<vdisk ID>:<value>|A`
+    * [0-core aggregation type][StatLogSpec]: Averages
+    * value unit: (partial) blocks per second
+ * throughput read
```
GlenDC: doesn't mention the aggregation period duration

docs/log.md
```diff
@@ -176,15 +176,36 @@ Sent when receiving an invalid config for a certain key, while reading or watchi
...
+    * logged by: nbdserver
+    * broadcasts: `10::vdisk.throughput.read@virt.<vdisk ID>:<value>|A`
+    * [0-core aggregation type][StatLogSpec]: Averages
+    * value unit: kB/s
+ * throughput write
```
GlenDC: doesn't mention the aggregation period duration

nbd/nbdserver/backend.go
```diff
@@ -60,6 +63,9 @@ func (ab *backend) WriteAt(ctx context.Context, b []byte, offset int64) (bytesWr
 	}
 
 	bytesWritten = int64(len(b))
+
+	go ab.writeStatistics.Send(bytesWritten)
```
GlenDC: starting a goroutine for this one seems really not necessary, we're simply sending a value over a channel (which should have a big enough buffer), so I don't see a need to  spawn a goroutine for each call (spawning goroutines is relatively cheap, but that's not reason to do it all the time).

statistics/broadcast_go19.go
```diff
@@ -1,14 +1,14 @@
 // +build go1.9
 
-package log
+package statistics
```
GlenDC: Ok first of all, if you want to move this into a separate package, I would have done so as `/log/statistics`, as it's still doing logging in the end.

Second of all, I don't think it was required as there shouldn't be a config dependency required in the log package. The clusterID should be given by who-ever logs it, and in fairness, this seems to be a very specific nbdserver thing, so I'm not even sure you need to have all the vdisk IOPS/throughput thing defined in the generic log package in the first place. I think having the broadcast function there was good enough, and the specialized logic should be in the nbdserver package, as it is only there required. I don't see where else we'll ever need logic to broadcast statistics about vdisks.

tlog/schema/tlog_schema.capnp.go
```diff
@@ -91,6 +91,11 @@ func (s HandshakeRequest_List) Set(i int, v HandshakeRequest) error {
 	return s.List.SetStruct(i, v.Struct)
 }
 
+func (s HandshakeRequest_List) String() string {
```
GlenDC: Again you updated and pushed a file to a PR that has nothing to do with your PR, and that you probably have done so by accident?

statistics/iops_throughput.go
```diff
@@ -0,0 +1,223 @@
...
+	"sync"
+	"time"
+
+	"github.com/zero-os/0-Disk/config"
+	"github.com/zero-os/0-Disk/log"
+)
+
+var (
+	// broadcast interval in seconds
+	interval = 1 * time.Minute
```
GlenDC: According to https://github.com/zero-os/0-core/tree/master/docs/monitoring#monitoring-metrics the aggregation duration is typically to be expected 5 minutes, is there a good reason why you are using 1 minute here?

statistics/iops_throughput.go
```diff
@@ -0,0 +1,223 @@
...
+// broadcastIOPSThroughput actually broadcast statistics based on provided data
+func broadcastIOPSThroughput(vdiskID string, blockSize int64, bytesPerSec float64, iopsKey, tpKey Key, tags MetricTags) {
+	// no need to log if 0?
+	if bytesPerSec == 0 {
+		log.Debugf("statistic value was 0, skipped broadcasting for vdisk: %s", vdiskID)
+		return
+	}
+
+	// broadcast iops
+	value := bytesPerSec / float64(blockSize)
```
GlenDC: I don't think it's a good idea to do this, there is no guarantee in the code I think that specifies the written bytes are always the block size, it could be less. Either way, it is safer to simply have a counter available.

statistics/iops_throughput.go
```diff
@@ -0,0 +1,223 @@
...
+// used for interacting with the statistics logger
+type IOPSThroughputLogger struct {
+	vdiskID       string
+	ctx           context.Context
+	dataMessenger chan<- int64
+	cancel        context.CancelFunc
+}
+
+// Send sends amount of bytes (read/written) to the log data(bytes) aggrigator
+func (l IOPSThroughputLogger) Send(data int64) {
```
GlenDC: you really want to make this a pointer type (and for all methods by the way), otherwise you're going to copy all its properties each time you make a call... I really don't think that's what you want.

statistics/iops_throughput.go
```diff
@@ -0,0 +1,223 @@
...
+	if l.ctx == nil {
```
GlenDC: I think you want to check `if l == nil` in case you really want to allow a nil-type here. Either way, I don't think that's a good idea. I think it's much better if you can somehow have the Logger as an interface available, so that you can provide a dummy one for testing purposes.

statistics/iops_throughput.go
chrisvdg: I think the aggregation this doc talks about is the 0-core one, it also aggregates at time intervals before pushing to LedisDB (5mins and 1h).
https://github.com/zero-os/0-core/blob/master/docs/monitoring/stats.md#where-do-the-statistics-go-anyway

As long as we broadcast the statistics within 5 minutes, it should be fine.

statistics/iops_throughput.go
chrisvdg: I can't do this if this should implement an interface right? Is there some way around that then?

statistics/iops_throughput.go
chrisvdg: The logger receives the amount of bytes written/ read regardless of blocksize.
At the interval the aggregated data is converted to bytes per second, again regardless of blocksize. Then it's devided by the blocksize to get the IOPS, which can also be a decimal value if it didn't end up as a round blocksize value.

That's how I tried to make it so that the writes/reads don't log full blocks for each read/write

statistics/broadcast_go19.go
chrisvdg: The backend doesn't keep track of the ClusterID and theClusterID can change since it's not in the static config no?

statistics/iops_throughput.go
GlenDC: Yeah sure, but that doc only ever seem to talk about 5 minutes exactly, it doesn't say anywhere about 5 minutes or less AFAIK. So why did you choose 1 minute?

statistics/iops_throughput.go
GlenDC: Of course you can...

statistics/iops_throughput.go
GlenDC: Sure but than your IOPS count can be wrong in the end, as you received it all regardless of the blocksize. Yet in the end you assume that the amount of operations = `totalBytesWritten / blockSize`, which might be incorrect.

statistics/broadcast_go19.go
GlenDC: Sure but it already is watching that config anyhow, so we can just as well provide a ClusterID and TemplateClusterID method.

GlenDC: Either way, the  bigger point is that this really shouldn't be a concern of the statistics logger, as that should be as dumb as possible. Its job is to log statistics for vdisks, all the input should come from the user.

statistics/iops_throughput.go
chrisvdg: Well if its `func (l *IOPSThroughputLogger) Send(data int64)`
Go complaints this doesn't implement the Logger interface

statistics/iops_throughput.go
chrisvdg: At first I didn't test with init overriding the interval, so waiting 5 minutes would be quite long, also in a discussion with Zaibon he said that logging every 1 minute was fine.

GlenDC: Sure, but not you're confusing things. The 5 minutes choice for production has nothing to do with the time you test, as you overwrite it anyway in your tests, so that is (hopefully) not the reason why you choose the 1 minute interval.

statistics/iops_throughput.go
GlenDC: The reason why that is happening (probably) is because you are creating the struct (that you wish to call the functions upon) by value, rather than by pointer. For all methods that you implement for an interface, you either have them all as by value, or by pointer, but you can't mix them. In most cases you'll choose to go by pointer, as to avoid copying values around all the time you make a call.

Check out https://play.golang.org/p/TXnz_42xzm, and try to change line 23 by `foor = foo{}`, you'll probably see the error you see in your console and which makes you think that what I'm saying is not possible.

chrisvdg: Aaah, ok, I see!

PR #440, opened 2017-09-07
update 0-stor client creation to use new 0-stor policy

zaibon:


Gopkg.lock
```diff
@@ -478,6 +478,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "b41c64c1ea6cc97c3a2d89a77a294153ecd41ed481b1faf682bef926b57b2c74"
+  inputs-digest = "8013d1253210767c6e604d4e39d9a5adf9baac73dbe7ee1faad408dcedf9eb70"
```
GlenDC: This should have been part of the vendor commit though, next time better luck 💃

tlog/stor/client.go
```diff
@@ -161,45 +157,21 @@ func (c *Client) LastHash() []byte {
...
-		},
+func newStorClientConf(conf Config) storclient.Policy {
+	return storclient.Policy{
+		Organization:           conf.Organization,
+		Namespace:              conf.Namespace,
+		DataShards:             conf.ZeroStorShards,
+		MetaShards:             conf.MetaShards,
+		IYOAppID:               conf.IyoClientID,
+		IYOSecret:              conf.IyoSecret,
+		Protocol:               "grpc",
```
GlenDC: can we turn this into constants? Doesn't have to be in this PR though, just would be nice if you can give that a thought and if you agree make an issue for it in the 0-stor repo. Perhaps you have good reason not to turn it into a constant though.

tlog/stor/client.go
```diff
@@ -161,45 +157,21 @@ func (c *Client) LastHash() []byte {
...
+		Compress:               true,
```
GlenDC: Is it still possible to choose the compression and encryption algorithm or is it now hardcoded for all users?

tlog/stor/client.go
```diff
@@ -161,45 +157,21 @@ func (c *Client) LastHash() []byte {
...
+		Encrypt:                true,
+		EncryptKey:             conf.EncryptPrivKey,
+		ReplicationNr:          0, //force to use distribution
```
GlenDC: Can you link me to docs or explain what is the difference between Replication vs distribution in this context?

Gopkg.lock
zaibon: yep, still learning the dep tool.

tlog/stor/client.go
zaibon: So more detail about how policy works
https://godoc.org/github.com/zero-os/0-stor/client#Policy

tlog/stor/client.go
zaibon: It's not configurable anymore at the moment.

tlog/stor/client.go
zaibon: Should be a constant, you are right.

tlog/stor/client.go
GlenDC: From those docs I wouldn't have guessed I have to set 0 to both properties to use distribution though.

```go
    // Number of replication to create when writting
    ReplicationNr int `yaml:"replication_nr"`
    // if data size is smaller than ReplicationMaxSize then data
    // will be replicated ReplicationNr time
    // if data is bigger, distribution will be used if configured
    ReplicationMaxSize int `yaml:"replication_max_size"`
```

Also I still don't understand the concepts of distribution vs replication in this context. I mean I have an idea, but these things can be ambiguous so I rather I have it clear in my head.

GlenDC:
@zaibon is it important that this still goes in alpha-8, instead of going straight to master?

zaibon:
Would be nice if we can integrate that in alpha8 yes.

GlenDC:
> Would be nice if we can integrate that in alpha8 yes.

I can do it, it's just that I had hoped we would stop pushing any unplanned changes to alpha-8. But if you want, I'll get this one through.

tlog/stor/client.go
zaibon: You don't need to put it to 0. You can have them both enabled. Then based on the size of the data you send, or replication or distribution will be used.

Replication is writing same data multiple time
Distribution is erasure coding,

GlenDC: Yeah I know, by default the int value in `Go` is `0`, so I suppose you're just being explicit here.

> Replication is writing same data multiple time
> Distribution is erasure coding

Thanks that clarifies it for me. As long as you only ever use Distribution and Erasure encoding, it should be simple enough to keep understanding it I suppose.

zaibon: >Yeah I know, by default the int value in Go is 0, so I suppose you're just being explicit here.

indeed, this is just to be explicit

zaibon: I'll improve the documentation regarding the policy of 0-stor and how thing works together

GlenDC: Yeah sure, even just linking to the relevant `/docs` documentation might already be good enough.

Issue #399, opened 2017-08-18
tlogserver : handle flush/0-stor failure


iwanbk:
What we can do:

- [x] notify 0-orchestrator (#424)
- [x] client move to other server (already implemented)
- [x] server's vdisk stop itself (#448)

iwanbk:
It seems the `zerostor` branch should at least `exit()`

iwanbk:
> It seems the zerostor branch should at least exit()

will be done with #371

iwanbk:
> What has to be done for a server's vdisk to restart itself? And what exactly do you mean with that @iwanbk?

It should be `stop` instead of `restart`. It might be ended up started again if the client reconnect again to this server.
It needs to : 
- kill all goroutine
- close all connected socket
- release other resource
There is unfinished work on `tlog-vdisk-stop-on-fail` branch.

GlenDC:
Once this issue is resolved make sure to open a PR with the solution using [the 1.1.0-alpha-8 branch](https://github.com/zero-os/0-Disk/tree/1.1.0-alpha-7) as the base branch. Master is from now on meant for beta code.

iwanbk:
all sub tasks has been implemented. closing it

PR #447, opened 2017-09-08
tlog : stop vdisk on failed flush.

iwanbk:
So the client could move to other server.
Or reconnect again and then we retry the flush.

Part of #399


GlenDC:
Why the `close->open->close`?

iwanbk:
into wrong branch (just realized that we could edit it)
have created another PR (#448) as replacement

Issue #425, opened 2017-08-29
VM doesn't boot correctly after restoring its boot disk


khaledkbadr:
**Steps to reproduce**:
* Copy vdisk from a template (vd2)
* Start a VM using qemu with nbd, tlog servers
* echo a file for in the VM after it boots
* Stop the VM (and stopping nbd, tlog servers)
* restore the vdisk `/bin/zeroctl restore vdisk -f vd2 --config 192.168.122.79:2379 --end-timestamp 1504018830000000000 --data-shards 1 -v --parity-shards 1`
* Start the vm again with nbd, tlog servers, it halts at this 
![image](https://user-images.githubusercontent.com/6059889/29828478-79159468-8ce5-11e7-859f-9a77f58a5a59.png)

**config in etcd**
```
vd2:vdisk:conf:static
blockSize: 4096
readOnly: false
size: 4
type: boot

vd2:vdisk:conf:storage:nbd
slaveStorageClusterID: ''
storageClusterID: myblockcluster
templateStorageClusterID: 87ec688e821f5f0006619b7207d656bd
tlogServerClusterID: tlogserver_vm0_525400123456

vd2:vdisk:conf:storage:tlog
zeroStorClusterID: myobjectcluster


myblockcluster:cluster:conf:storage
dataStorage:
- address: 192.168.122.76:2000
- address: 192.168.122.79:2000
label: myblockcluster
metadataStorage:
  address: 192.168.122.79:2001
nodes:
- '525400123456'
- '525400123459'
status: ready

myobjectcluster:cluster:conf:zerostor
iyo:
  clientID: <clientID>
  namespace: test
  org: zero-khaled
  secret: <secret>
metadataServers:
- address: 192.168.122.79:2379
servers:
- address: 192.168.122.76:2002
- address: 192.168.122.79:2002

vm0:nbdserver:conf:vdisks
vdisks:
- vd2

tlogserver_vm0_525400123456:cluster:conf:tlog
servers:
- 192.168.122.79:11211
```

iwanbk:
Did you see something in the logs of nbdserver and tlogserver?

khaledkbadr:
Nothing that raises red flags, just success negotiations....etc

iwanbk:
@khaledkbadr need logs of these:
- nbd & tlog before restore : normal logs is enough
- restoring : verbose log
- nbd & tlog after restore : verbose log

khaledkbadr:
**Logs of nbdserver before restore**
[...]

Issue #435, opened 2017-09-04
proposal: redesign the tlog module


GlenDC:
If we look at all issues created by the Cairo team (our only users for now), which aren't feature requests or due to miss communications about how the product is used or what is expected, we can see that almost all issues are related to the tlog module. On top of that there have been over the couple of months only a few cases where travis CI tests discovered hard-to-reproduce multi-goroutine-related issues, which I think all (or at least most) were about the tlog module.

So **stability** is definitely an important reason, and could already be enough reason on its own to force us to take a hard look at the current design.

## Too complex

If you look at the problem the tlog module is solving, you wouldn't think it's too complex. In the end all it has to do is track all (write) updates done for a vdisk and aggregate them, so it can be stored using the 0-Stor client. I do have to admit that due to several factors the problem is more complex than you would expect from simply reading a high level description.

Despite this I believe that the current implementation is too complex. The code has a lot of moving parts, some connected, others not. There are a lot of goroutines happening, and a lot of mutexes, really a lot of mutexes. Probably too many of those.

## Too general

I think part of the over-complexity mentioned above is due to the fact that the design of the tlog module originates from a perspective where it was still meant to be general. However I think those plans have changed. We now have the 0-stor repository which provides all the generic logic one might need for other projects, which makes me think that the tlog module of the 0-disk repo will only ever be used for 0-Disk/tlog purposes. So while stuff like the tlogclient already leaned towards 0-Disk specialization, we can definitely go a lot further in this, and remove some moving parts because of this as well.

It doesn't explain all over-complexity, and neither will it solve all of it, but it will help.

## Multiple Tlog Servers

Because of the current tlog design we can only ever use 1 tlogserver at the same time. This is a hard nut too crack as the current design makes it very difficult to switch to multiple tlogservers at once, due to the fact that the tlogserver is doing all the sequence ordering, as well as assembling the aggregations. This makes it impossible to have multiple tlogservers at once, but only because of how the design is, not because of what the problem itself is.

This is sad, as we do support providing multiple tlogservers (per nbdserver). For now this is used only because we than can switch to another tlogserver in case the currently used one fails for whatever reason.

What if we could spread the work at all times over all available tlogservers, such that we no longer have to switch to a new tlogserver fails, but we can simply continue to work even if one of the multiple tlogservers which would work in parallel fail.

[...]

PR #448, opened 2017-09-08
tlog : stop vdisk on failed flush.

iwanbk:
So the client could move to other server.
Or reconnect again and then we retry the flush.

Part of #399

### DIFF
--- a/nbd/nbdserver/backend.go
+++ b/nbd/nbdserver/backend.go
@@ -244,7 +244,7 @@ func (ab *backend) GoBackground(ctx context.Context) {
 		case err = <-done:
 			log.Infof("vdisk '%s' finished the flush under SIGTERM handler", ab.vdiskID)
 
-		case <-time.After(2 * time.Minute):
+		case <-time.After(flushWaitRetry * flushWaitRetryNum):
 			// TODO :
 			// - how long is the reasonable waiting time?
 			// - put this value in the config?
--- a/nbd/nbdserver/const.go
+++ b/nbd/nbdserver/const.go
@@ -1,0 +1,10 @@
+package main
+
+import (
+	"time"
+)
+
+const (
+	flushWaitRetry    = time.Minute // retry flush after this timeout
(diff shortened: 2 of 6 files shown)


GlenDC:
Checked out your PR, I'm not the biggest fan of it, but I have nothing to add for now. So if you feel like this fixes your issue feel free to merge it in. I won't block you.

iwanbk:
Which part that you dont like.
I might be able to improve it now or later

GlenDC:
Not sure, the same general bad feeling that I had when creating #435. It just keeps going in the same pattern, but I have nothing extra to add here. And in general I have no energy/time to provide a solution now or soon anyhow, so no reason to block you on this. If it for now does what you think it should do, that's all that matters now i guess.

iwanbk:
> Not sure, the same general bad feeling that I had when creating #435. It just keeps going in the same pattern, but I have nothing extra to add here

Got it, i do understand your point.
As i noted on that issue also, i also have proposal for re-design

iwanbk:
@GlenDC any ETA when you will merge this to master?

GlenDC:
@iwanbk when you and @khaledkbadr confirm that this indeed fixes the open issue #425...

iwanbk:
It never intended to fix #425, see PR description

GlenDC:
Oh yes you're right, I'm sorry, I'm confused with another PR. It's morning you know.

GlenDC:
To be honest I was waiting for a merge to master until all issues were closed. Do you need this PR than right now? Does it block you?

iwanbk:
> To be honest I was waiting for a merge to master until all issues were closed

I think it is still too morning for you because i saw few PRs for alpha-8 already merged to master :D 

Yeah, i need it.

GlenDC:
No that wasn't a mistake. Those were merged a long time ago. When I said I was waiting, I mean after that. But OK I'll merge it in today in that case.

Issue #405, opened 2017-08-21
Rolling back a disk fails after stopping a the machine


ashraffouda:
Rolling back a disk with this scenario fails 
1- Create a disk 
2- create a vm with this disk 
3- stop the machine 
4- resize the disk 
5- do roll back for this disk 
the error is 
```
 File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/actions/vdisk/49fc5356ac3092fc5e7e040537449d5e.py", line 225, in rollback
    raise j.exceptions.RuntimeError("Failed to run zeroctl restore {} {}".format(result.stdout, result.stderr))
JumpScale9.errorhandling.JSExceptions.RuntimeError: ERROR: Failed to run zeroctl restore nil last hash Error: nil last hash
Usage:
  zeroctl restore vdisk id [flags]

Flags:
      --config SourceConfig        config resource: dialstrings (etcd cluster) or path (yaml file) (default config.yml)
      --end-timestamp uint         end timestamp in nanosecond(default 0: until the end)
  -f, --force                      when given delete the vdisk if it already existed
  -h, --help                       help for vdisk
      --k int                      K variable of erasure encoding (default 4)
      --m int                      M variable of erasure encoding (default 2)
      --nonce string               hex nonce used for encryption (default "37b8e8a308c354048d245f6d")
      --priv-key string            private key (default "12345678901234567890123456789012")
      --start-timestamp uint       start timestamp in nanosecond(default 0: since beginning)
      --storage-addresses string   comma seperated list of redis compatible connectionstrings (format: '<ip>:<port>[@<db>]', eg: 'localhost:16379,localhost:6379@2'), if given, these are used for all vdisks, ignoring the given confi
g

Global Flags:
  -v, --verbose   log available information

 ((type:runtime.error)
[Wed16 13:52] - Job.py              :56  :j.core.jobcontroller.job.35e18ec418e72c0f437722c5e1e20e0c - ERROR    - Traceback (most recent call last):
  File "/usr/lib/python3.5/concurrent/futures/thread.py", line 55, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/tmp/actions/vdisk/49fc5356ac3092fc5e7e040537449d5e.py", line 346, in processChange
    j.tools.async.wrappers.sync(service.executeAction('rollback', args={'timestamp': args['timestamp']}, context=job.context))
[...]

PR #453, opened 2017-09-08
Fix zeroctl copy and nbd close

iwanbk:
Fix attempt of #405 and #425: 
- also generates tlog data when copying template which doesn't have tlog data
- tlogStorage : try to `Flush` in nbd `Close`

Tested with same flow as #425 , except that #425 using 0-orch
- copy vdisk
- start vm
- restore vdisk with `-f` option
- restart vm


GlenDC:
Has it been confirmed @iwanbk that this fixes the issue of #425?

nbd/nbdserver/tlog.go
```diff
@@ -206,6 +206,10 @@ func (tls *tlogStorage) Flush() (err error) {
 	tls.mux.Lock()
 	defer tls.mux.Unlock()
 
+	if tls.cache.Empty() {
+		return
```
GlenDC: Cannot do that here, you're forgetting to call `tls.storage.Flush()` in that case, which is needed as the LBA might still have some unflushed data in-memory (meaning it won't be written to the storage metadata server and thus lose data).

nbd/nbdserver/tlog.go
```diff
@@ -254,6 +258,14 @@ func (tls *tlogStorage) Flush() (err error) {
 
 // Close implements BlockStorage.Close
 func (tls *tlogStorage) Close() (err error) {
+	log.Infof("tlogStorage Close with lastSequence = %v, cache empty = %v",
+		tls.getLatestSequence(), tls.cache.Empty())
+
+	err = tls.Flush()
```
GlenDC: Why are you flushing here? That shouldn't be needed. If you need that here it means we are mis-using the API somewhere. Feel free to log/return an error in case the cache isn't empty, but really doing flushing logic here is too much I think.

This function should be about closing any open resources, not flushing.

zeroctl/cmd/copyvdisk/vdisk.go
```diff
@@ -136,4 +165,18 @@ WARNING: when copying nondeduped vdisks,
 	VdiskCmd.Flags().BoolVar(
 		&vdiskCmdCfg.ForceSameStorageCluster, "same", false,
 		"enable flag to force copy within the same nbd servers")
+
+	VdiskCmd.Flags().IntVar(
```
GlenDC: Why are these flags here? That should be configured in the config, no?

zeroctl/cmd/restore/vdisk.go
```diff
@@ -69,7 +69,10 @@ func restoreVdisk(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	_, err = player.Replay(decoder.NewLimitByTimestamp(vdiskCmdCfg.StartTs, vdiskCmdCfg.EndTs))
+	log.Infof("restoring vdisk with start timestamp=%v end timestamp=%v",
+		vdiskCmdCfg.StartTs, vdiskCmdCfg.EndTs)
+	lastSeq, err := player.Replay(decoder.NewLimitByTimestamp(vdiskCmdCfg.StartTs, vdiskCmdCfg.EndTs))
+	log.Infof("restore finished with last sequence = %v and err = %v", lastSeq, err)
```
GlenDC: No need to log this error, it will already be returned by the cmd error. And in case you do want to log the error for some good reason, log it under the Error level not the Info level.

zeroctl/cmd/copyvdisk/vdisk.go
```diff
@@ -89,6 +93,31 @@ func copyVdisk(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf(
+			"couldn't read source vdisk %s's static config: %v", targetVdiskID, err)
+	}
+
+	if targetStaticConfig.Type.TlogSupport() {
+		log.Infof("generating tlog data for target vdisk `%v`", targetVdiskID)
+		generator, err := generator.New(configSource, generator.Config{
+			SourceVdiskID: sourceVdiskID,
+			TargetVdiskID: targetVdiskID,
+			DataShards:    vdiskCmdCfg.DataShards,
```
GlenDC: You're already giving the config source to the generator, I really don't see why we would also have to give the tlog shard/parity/privKey info ourselves here. They should really be received from the source and target vdisks.

tlog/generator/generator.go
```diff
@@ -0,0 +1,103 @@
...
+	configSource  config.Source
+}
+
+// New creates new Generator
+func New(configSource config.Source, conf Config) (*Generator, error) {
+	if err := validator.Validate(conf); err != nil {
+		return nil, err
+	}
+
+	flusher, err := flusher.New(configSource, conf.DataShards, conf.ParityShards, conf.TargetVdiskID, conf.PrivKey)
```
GlenDC: again no need to pass in data shards, parity shards and priv Key here, as the flusher constructor (`New`), can get this using the passed in vdiskID and the configSource.

tlog/generator/generator.go
```diff
@@ -0,0 +1,103 @@
...
+		BlockSize:       int64(staticConf.BlockSize),
+	}, ardbProv)
+	if err != nil {
+		return err
+	}
+	defer sourceStorage.Close()
+
+	var seq uint64
+	for _, idx := range indices {
+		content, err := sourceStorage.GetBlock(idx)
```
GlenDC: Why not spawn a few worker goroutines for this? I know you need to add them in order, but you could than easily force that order right, by keeping an out-of-order buffer, before you add it to the flusher, no?

nbd/nbdserver/tlog.go
iwanbk: ok, removed

nbd/nbdserver/tlog.go
iwanbk: you're right.
It was because i saw the `Close` being called when it still under `Flush`.
I added `tls.mux.Lock` as a fix

zeroctl/cmd/copyvdisk/vdisk.go
iwanbk: I was sure we don't have `shard/parity/privKey` in configSource, i also see that it  become command  line options in `zeroctl restore`

zeroctl/cmd/copyvdisk/vdisk.go
iwanbk: I was sure we don't have shard/parity/privKey in configSource, i also see that it become command line options in zeroctl restore

tlog/generator/generator.go
iwanbk: I was sure we don't have shard/parity/privKey in configSource, i also see that it become command line options in zeroctl restore

tlog/generator/generator.go
iwanbk: we don't need to order it.
Just don't feel the need to optimize it now in this stage.

zeroctl/cmd/copyvdisk/vdisk.go
GlenDC: Ah yes, sorry. I don't know what I was thinking.

tlog/generator/generator.go
```diff
@@ -80,20 +83,98 @@ func (g *Generator) GenerateFromStorage() error {
...
+		content []byte
+	}
+	var (
+		wg              sync.WaitGroup
+		numProcess      = runtime.NumCPU()
+		indicesCh       = make(chan int64, numProcess)
+		idxContentCh    = make(chan idxContent, numProcess)
+		errCh           = make(chan error)
+		doneCh          = make(chan struct{})
+		ctx, cancelFunc = context.WithCancel(context.Background())
```
GlenDC: I think it would be better if `context.Background()` is replaced by a context given as a parameter to this function.

tlog/generator/generator.go
```diff
@@ -80,20 +83,98 @@ func (g *Generator) GenerateFromStorage() error {
...
-		content, err := sourceStorage.GetBlock(idx)
-		if err != nil {
-			return err
+	type idxContent struct {
+		idx     int64
+		content []byte
+	}
+	var (
+		wg              sync.WaitGroup
+		numProcess      = runtime.NumCPU()
```
GlenDC: While this is a good default, I think it should also be available as a parameter (or config)  for this function. This way the user can also have less or more parallel tasks, depending on her needs. Even better it allows us to expose an `-j, --jobs` flag, to the `zeroctl` commands using this function, so that users of the tool can also choose what they want. This is also available for the `export-/import vdisk` commands.

tlog/generator/generator.go
```diff
@@ -80,20 +83,98 @@ func (g *Generator) GenerateFromStorage() error {
...
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		for ic := range idxContentCh {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				err = g.flusher.AddTransaction(schema.OpSet, seq, ic.content, ic.idx, tlog.TimeNowTimestamp())
```
GlenDC: I think it makes more sense to have the same timestamps for all blocks, as we are creating them all at once. Because having different timestamps makes me think that some blocks are newer than others, while in fact they are all equally old (the time of copying). So I think it would be better if you just at the start of a function generate the timestamp as `timestamp := tlog.TimeNowTimestamp()` and use that for all `g.flusher.AddTransaction` calls.

tlog/generator/generator.go
iwanbk: done

tlog/generator/generator.go
iwanbk: done

tlog/generator/generator.go
iwanbk: fixed

zeroctl/cmd/copyvdisk/vdisk.go
```diff
@@ -179,4 +183,9 @@ WARNING: when copying nondeduped vdisks,
 		"priv-key", "12345678901234567890123456789012",
 		"private key")
 
+	VdiskCmd.Flags().IntVar(
```
GlenDC: Better use `IntVarP` here, so we can also provide a shorthand, as `j` would make sense here as a short-hand alternative. Will do it myself while merging in your changes.

iwanbk: OK., thanks

PR #472, opened 2017-09-14
remove need of IYO environment variable to run tests

### LINKED ISSUE #457: make tlog tests work without need for iyo env variables
Currently one cannot run any tests which create a 0-stor client, without defining _valid_ values for the three `iyo` variables: `iyo_organization`, `iyo_client_id` and `iyo_secret`.

0-Stor seems to use a PEM certificate and public key for this, as can be seen in:
+ https://github.com/zero-os/0-stor/blob/master/client/client_test.go#L27
+ https://github.com/zero-os/0-stor/blob/master/client/client_test.go#L68

That could be a solution. Either way, we need something so travis and developers can run tests again without requiring to set environment variables.

zaibon:
fixes #457

### DIFF
--- a/config/format.go
+++ b/config/format.go
@@ -420,8 +420,8 @@ func (cfg *ZeroStorClusterConfig) Equal(other *ZeroStorClusterConfig) bool {
 type IYOCredentials struct {
 	Org       string `yaml:"org" valid:"required"`
 	Namespace string `yaml:"namespace" valid:"required"`
-	ClientID  string `yaml:"clientID" valid:"required"`
-	Secret    string `yaml:"secret" valid:"required"`
+	ClientID  string `yaml:"clientID" valid:"optional"`
+	Secret    string `yaml:"secret" valid:"optional"`
 }
 
 // NewTlogClusterConfig creates a new TlogClusterConfig from a given YAML slice.
--- a/config/format_data_test.go
+++ b/config/format_data_test.go
@@ -330,26 +330,6 @@ servers:
   - address: "1.1.1.1:11"
 metadataServers:
   - address: "3.3.3.3:33"
-`, // missing clienid
-	`
-iyo:
-  org: "foo org"
-  namespace: "foo namespace"
-  secret: "foo secret"
-servers:
-  - address: "1.1.1.1:11"
-metadataServers:
-  - address: "3.3.3.3:33"
-`, // missing secret
-	`
-iyo:
-  org: "foo org"
-  namespace: "foo namespace"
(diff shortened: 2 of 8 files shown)


zaibon:
It's a actually more tricky then that cause after removing the IYO credentials, I have other tests failing. Mostly because of the validation of the tlog config. I need to spent bit more time knowing the code

GlenDC:
@zaibon can you please also prune the vendor folder after updating it (`dev prune`), just make sure that you undo the deletion of `vendor/zombiezen.com/go/capnproto2/capnpc-go` and `vendor/zombiezen.com/go/capnproto2/std/go.capnp`, as we need those, even though they get pruned automatically.

Because right now you added a lot of extra files in the vendor folder which we don't require at all.

GlenDC:
Also please rebase master into this branch and resolve the conflicts.

GlenDC:
> It's a actually more tricky then that cause after removing the IYO credentials, I have other tests failing. Mostly because of the validation of the tlog config. I need to spent bit more time knowing the code

No worries, let me know when it's ready for another review.

iwanbk:
> It's a actually more tricky then that cause after removing the IYO credentials, I have other tests failing. Mostly because of the validation of the tlog config. I need to spent bit more time knowing the code

It seems because you replace below code (and other similar code) to empty string.
```go
os.Getenv("iyo_organization")
```
random value should work i guess

GlenDC:
Why you do a merge? This is a feature branch, would have been cleaner to simply do a rebase & force-push. You're the only one on this branch so it's possible.

zaibon:
The only conflict I had was in the vendor. I tried to rebase, it made a huge mess.

GlenDC:
> The only conflict I had was in the vendor. I tried to rebase, it made a huge mess.

OK... but either way, as part of that rebase you could have dropped that vendor update, as the update to 0-Stor on master is newer anyhow.

zaibon:
no that was the problem, the one from master was not newer, that's why I had to do it here too.
Let me remove this merge commit, and clean it up

zaibon:
@GlenDC have a look again ;-)

GlenDC:
Unit tests pass as well, does it mean that it's ready for a final review @zaibon?

zaibon:
yep

GlenDC:
@zaibon can you also delete the environment variables from travis?

zaibon:
@GlenDC yep no problem

zaibon:
> @zaibon can you also delete the environment variables from travis?

Done

PR #480, opened 2017-09-19
Fix zeroctl copy : vdisk has tlog support if it has non empty tlogServerClusterId

iwanbk:
Basically do the same thing as #479 for alpha-8, but `copy tlog` code has been changed a lot in master, so need different PR

### DIFF
--- a/tlog/copy/copy.go
+++ b/tlog/copy/copy.go
@@ -7,6 +7,7 @@ import (
 	"gopkg.in/validator.v2"
 
 	"github.com/zero-os/0-Disk/config"
+	"github.com/zero-os/0-Disk/log"
 )
 
 // Config represents config for copy operation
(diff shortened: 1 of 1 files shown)


tlog/copy/copy.go
```diff
@@ -47,15 +60,22 @@ func Copy(ctx context.Context, confSource config.Source, conf Config) error {
...
 	}
 
+	sourceVdiskNBDConfig, err := config.ReadVdiskNBDConfig(confSource, conf.SourceVdiskID)
+	if err != nil {
+		return fmt.Errorf(
+			"couldn't read source vdisk %s's storage config: %v", conf.SourceVdiskID, err)
+	}
 	// if source vdisk has no tlog, generate it
-	if !sourceStaticConf.Type.TlogSupport() {
+	if !sourceStaticConf.Type.TlogSupport() || sourceVdiskNBDConfig.TlogServerClusterID == "" {
```
zaibon: shouldn't this be the opposite? If not tlog configured, we don't do anything?

zaibon: But this should only happens if the source vdisk doesn't have tlog data, but the target vdisk has a tlog configured right ?

iwanbk: Yeah, check for target vdisk also in this PR https://github.com/zero-os/0-Disk/pull/480/files#diff-0e86d341a795fa64360e611d587a8170R49

zaibon: ok LGTM

GlenDC:
This PR is no longer required due to the merging of the original PR (#479).

PR #484, opened 2017-09-22
Tlog : remove incomplete support for multiple connections for a vdisk

### LINKED ISSUE #451: tlog : remove incomplete support for multiple connections for single vdisk
We had a plan to add support for multiple TCP connections for vdisk at #128.
Some code has been added to the server to handle it.
And maybe in the client also, not sure.

I think it is better to remove the code because:
- it is not usefull
- just added unnecessary complexity
- easy to re-add
- cleaner code might help to re-design the tlog

It might also closes #128 if we decide to not add it at all

iwanbk:
Fixes #451


tlog/schema/tlog_schema.capnp
```diff
@@ -34,6 +34,7 @@ struct TlogClientMessage {
 		block @0 :TlogBlock;        # block message
 		forceFlushAtSeq @1 :UInt64; # force flush at seq message
 		waitNBDSlaveSync @2 :Void;  # Wait NBD Slave Sync message
+        disconnect @3 :Void;        # disconnect from server
```
GlenDC: You forgot the indentation.

PR #489, opened 2017-09-26
tlog : remove reset first sequence

### LINKED ISSUE #319: tlogserver : remove ResetFirstSequence feature
It was added because nbdserver unable to get correct sequence number upon vdisk restart.
Once it has correct number, either by solving #273 or using `0-stor-lib`, we can remove this feature.

Correct way can be found on https://github.com/zero-os/0-Disk/issues/273#issuecomment-313401453

> modify the HandshakeResponse of the tlog handhsake protocol, so that it also returns the sequence it will use (it should be the firstSequence given by the client in case resetFirstSequence is true, but it can also be a higher value in case that bool is false), obviously we than also no longer want to reset that first sequence in our tlogStorage;
> in case the returned sequence of the server is higher than the one we had stored, you will need to use the tlog replay feature to get the vdisk up to date;
>

iwanbk:
Fixes #319 
Part of #230


nbd/nbdserver/tlog/tlog.go
```diff
@@ -49,17 +49,18 @@ func Storage(ctx context.Context, vdiskID, clusterID string, configSource config
...
 		if err != nil {
 			cancel()
 			return nil, errors.New("tlogStorage requires valid tlogclient: " + err.Error())
 		}
+
+		if metadata.lastFlushedSequence < client.LastFlushedSequence() {
+			// TODO call tlog player if last flushed sequence from tlog server is
+			// higher than us.
+			// see: https://github.com/zero-os/0-Disk/issues/230
+			log.Fatalf("couldn't stat vdisk `%v`: need to sync with tlog", vdiskID)
```
GlenDC: You cannot do a fatal log, as it will exit the entire nbdserver. That's not something you'll want. Instead simply return an error and you'll stop the mounting of _this_ vdisk only, which is what you do want.

tlog/stor/client.go
```diff
@@ -157,7 +162,20 @@ func (c *Client) ProcessStore(blocks []*schema.TlogBlock) ([]byte, error) {
 
 	c.lastMd = lastMd
 	c.lastMetaKey = key
-	return data, c.saveLastMetaKey()
+	c.lastSequence = blocks[len(blocks)-1].Sequence()
+
+	c.storeNum++
```
GlenDC: Why store it at all in external storage before some kind of official flush call? I suppose you want to be sure it is stored, but if the flush func is not called before exit (even in case of an error), than perhaps that indicates a bug?

nbd/nbdserver/tlog/tlog.go
iwanbk: oh, i just changed it to `log.Errorf` before reading your comment.
I didn't return error because it is the old behavior and on our todo.
But yeah, returning error seems a better way.

tlog/stor/client.go
iwanbk: No, it is called after the flush

PR #494, opened 2017-09-27
nbd : don't return error in case it needs to sync with tlog.

iwanbk:
Some reasons:
- the sync feature still in todo
- the check (whether we need to sync or not) itself still not tested
- it also the old behaviour

### DIFF
--- a/nbd/nbdserver/tlog/tlog.go
+++ b/nbd/nbdserver/tlog/tlog.go
@@ -59,8 +59,7 @@ func Storage(ctx context.Context, vdiskID, clusterID string, configSource config
 			// TODO call tlog player if last flushed sequence from tlog server is
 			// higher than us.
 			// see: https://github.com/zero-os/0-Disk/issues/230
-			cancel()
-			return nil, fmt.Errorf("error when starting vdisk `%v`: need to sync with tlog", vdiskID)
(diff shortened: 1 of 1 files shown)


nbd/nbdserver/tlog/tlog.go
```diff
@@ -59,8 +59,7 @@ func Storage(ctx context.Context, vdiskID, clusterID string, configSource config
 			// TODO call tlog player if last flushed sequence from tlog server is
 			// higher than us.
 			// see: https://github.com/zero-os/0-Disk/issues/230
-			cancel()
-			return nil, fmt.Errorf("error when starting vdisk `%v`: need to sync with tlog", vdiskID)
+			log.Infof("possile error when starting vdisk `%v`: might need to sync with tlog", vdiskID)
```
GlenDC: This should definitely be logged as an error _and_ also returned as an error. As this is a critical issue. We should probably also warn AYS using an STDERR log. Because if you just continue now, you'll keep with a corrupted tlog.

iwanbk:
Please read the commit message

GlenDC:
I read the commit message. Perhaps I'm a bit too early, I'm just afraid this will be forgotten and stay just an info log forever... But OK, fine for now.

iwanbk:
> I'm just afraid this will be forgotten and stay just an info log forever... But OK, fine for now.

No, the issue still open at #230

PR #495, opened 2017-09-27
tlog : prepend vdisk ID to the data to generate unique hash.

iwanbk:
It was relied on the fact that aggregation contains vdiskID.
But vdiskID has been removed at 0809ea4db380df30edb382e962d1d8c2bbfab4ec

### DIFF
--- a/tlog/stor/client.go
+++ b/tlog/stor/client.go
@@ -135,7 +135,7 @@ func (c *Client) ProcessStore(blocks []*schema.TlogBlock) ([]byte, error) {
 		return nil, err
 	}
 
-	key := c.hasher.Hash(data)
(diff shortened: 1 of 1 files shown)


tlog/stor/client.go
```diff
@@ -135,7 +135,7 @@ func (c *Client) ProcessStore(blocks []*schema.TlogBlock) ([]byte, error) {
 		return nil, err
 	}
 
-	key := c.hasher.Hash(data)
+	key := c.hasher.Hash(append([]byte(c.vdiskID), data...))
```
GlenDC: And why exactly does this hash have to be unique, even though we save it under a unique key already?

PR #506, opened 2017-09-29
Generates tlog data when importing a vdisk.

### LINKED ISSUE #466: generate tlog data when importing a vdisk (from a backup)
When importing a vdisk from a (FTP) backup currently we only generate storage data.

However in case that vdisk is configured with tlog support, and also has a tlogclustered configured for it, we should also generate tlog data, I think.

This will make importing a vdisk slower, but I think we have no other choice, as otherwise the tlog feature for that vdisk will be useless, as a restore would remain with partial results until all indices that were available since the import were overridden.

This might be able to utilize the tlog generator written as part of the PR #453, but I'm not too sure.

iwanbk:
Fixes #466

### DIFF
--- a/docs/zeroctl/commands/import.md
+++ b/docs/zeroctl/commands/import.md
@@ -4,6 +4,7 @@
 
 Import a [vdisk][vdisk] from a storage (FTP) server,
 where the backup is stored in a secure and efficient manner.
+Tlog data will be generated if the vdisk has configured tlog cluster.
 
 > (!) Remember to use the same (snapshot) name,
 crypto (private) key and the compression type,
@@ -39,11 +40,14 @@ Flags:
   -b, --blocksize int                 the size of the exported (deduped) blocks (default 131072)
   -c, --compression CompressionType   the compression type to use, options { lz4, xz } (default lz4)
       --config SourceConfig           config resource: dialstrings (etcd cluster) or path (yaml file) (default config.yml)
+      --data-shards int               data shards (K) variable of erasure encoding (default 4)
   -f, --force                         when given, delete the vdisk if it already existed
   -h, --help                          help for vdisk
   -j, --jobs int                      the amount of parallel jobs to run (default $NUMBER_OF_CPUS)
   -k, --key AESCryptoKey              an optional 32 byte fixed-size private key used for decryption when given
-  -s, --storage StorageConfig         ftp server url or local dir path to import the backup from (default $HOME/.zero-os/nbd/vdisks)
+      --parity-shards int             parity shards (M) variable of erasure encoding (default 2)
+  -s, --storage StorageConfig         ftp server url or local dir path to import the backup from (default /$HOME/.zero-os/nbd/vdisks)
+      --tlog-priv-key string          tlog private key (default "12345678901234567890123456789012")
 
 Global Flags:
   -v, --verbose   log available information
--- a/tlog/stor/config_source.go
+++ b/tlog/stor/config_source.go
@@ -2,6 +2,7 @@ package stor
 
 import (
 	"github.com/zero-os/0-Disk/config"
+	"github.com/zero-os/0-Disk/log"
 )
 
 // ConfigFromConfigSource creates tlog stor client config from config.Source
@@ -9,12 +10,14 @@ func ConfigFromConfigSource(source config.Source, vdiskID, privKey string, dataS
 	// read vdisk config
 	vdiskConf, err := config.ReadVdiskTlogConfig(source, vdiskID)
 	if err != nil {
+		log.Errorf("failed to read vdisk tlog config for vdisk `%v`: %v", vdiskID, err)
 		return
 	}
 
 	// read zerostor config of this vdisk
 	zsc, err := config.ReadZeroStoreClusterConfig(source, vdiskConf.ZeroStorClusterID)
 	if err != nil {
+		log.Errorf("failed to read ZeroStorCluster config for vdisk `%v`: %v", vdiskID, err)
 		return
 	}
 
--- a/zeroctl/cmd/backup/import.go
+++ b/zeroctl/cmd/backup/import.go
@@ -11,7 +11,9 @@ import (
 	"github.com/zero-os/0-Disk/log"
 	"github.com/zero-os/0-Disk/nbd/ardb/backup"
 	"github.com/zero-os/0-Disk/nbd/ardb/storage"
-	"github.com/zero-os/0-Disk/nbd/nbdserver/tlog"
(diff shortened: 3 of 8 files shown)


docs/zeroctl/commands/import.md
```diff
@@ -39,14 +40,18 @@ Flags:
   -b, --blocksize int                 the size of the exported (deduped) blocks (default 131072)
   -c, --compression CompressionType   the compression type to use, options { lz4, xz } (default lz4)
       --config SourceConfig           config resource: dialstrings (etcd cluster) or path (yaml file) (default config.yml)
+      --data-shards int               data shards (K) variable of erasure encoding (default 4)
   -f, --force                         when given, delete the vdisk if it already existed
   -h, --help                          help for vdisk
-  -j, --jobs int                      the amount of parallel jobs to run (default $NUMBER_OF_CPUS)
+  -j, --jobs int                      the amount of parallel jobs to run (default 8)
```
GlenDC: copy-paste error: should be `$NUMBER_OF_CPUS`, not `8`.

docs/zeroctl/commands/import.md
```diff
@@ -39,14 +40,18 @@ Flags:
...
   -k, --key AESCryptoKey              an optional 32 byte fixed-size private key used for decryption when given
-  -s, --storage StorageConfig         ftp server url or local dir path to import the backup from (default $HOME/.zero-os/nbd/vdisks)
+      --parity-shards int             parity shards (M) variable of erasure encoding (default 2)
+  -s, --storage StorageConfig         ftp server url or local dir path to import the backup from (default /home/ibk/.zero-os/nbd/vdisks)
```
GlenDC: copy-paste error: should be `$HOME/.zero-os/nbd/vdisks`, not `/home/ibk/.zero-os/nbd/vdisks`.

docs/zeroctl/commands/import.md
```diff
@@ -39,14 +40,18 @@ Flags:
...
+      --tlog-priv-key string          tlog private key (default "12345678901234567890123456789012")
 
 Global Flags:
   -v, --verbose   log available information
+
```
GlenDC: please remove this added line

tlog/stor/config_source.go
```diff
@@ -9,12 +11,14 @@ func ConfigFromConfigSource(source config.Source, vdiskID, privKey string, dataS
 	// read vdisk config
 	vdiskConf, err := config.ReadVdiskTlogConfig(source, vdiskID)
 	if err != nil {
+		err = fmt.Errorf("failed to read vdisk tlog config for vdisk `%v`: %v", vdiskID, err)
```
GlenDC: Might be better to log this error message instead of wrapping it as the new error message. The problem with wrapping error messages is that you can't any longer check if a returned error is equal to a certain public error variable, or of a certain type or w/e.

Perhaps it's not important now, but as it's a function to be used in many different places, it might become important.

tlog/stor/config_source.go
```diff
@@ -9,12 +11,14 @@ func ConfigFromConfigSource(source config.Source, vdiskID, privKey string, dataS
...
 		return
 	}
 
 	// read zerostor config of this vdisk
 	zsc, err := config.ReadZeroStoreClusterConfig(source, vdiskConf.ZeroStorClusterID)
 	if err != nil {
+		err = fmt.Errorf("failed to read ZeroStorCluster config for vdisk `%v`: %v", vdiskID, err)
```
GlenDC: Might be better to log this error message instead of wrapping it as the new error message. The problem with wrapping error messages is that you can't any longer check if a returned error is equal to a certain public error variable, or of a certain type or w/e.

Perhaps it's not important now, but as it's a function to be used in many different places, it might become important.

zeroctl/cmd/backup/backup.go
```diff
@@ -17,4 +17,7 @@ var vdiskCmdCfg struct {
 	CompressionType     backup.CompressionType // optional
 	JobCount            int                    // optional
 	Force               bool                   //optional
+	DataShards          int
```
GlenDC: Please add those into a separate struct meant only for the import command and defined within the `import.go` file. This config struct is meant _only_ for shared config variables between the import and export command.

zeroctl/cmd/backup/import.go
```diff
@@ -56,7 +63,36 @@ func importVdisk(cmd *cobra.Command, args []string) error {
...
+	log.Info("Importing the vdisk")
+	err = backup.Import(ctx, cfg)
+	if err != nil {
+		return err
+	}
+
+	// check if this vdisk has tlog cluster
+	hasTlogCluster, err := tlog.HasTlogCluster(configSource, vdiskCmdCfg.VdiskID)
+	if err != nil || !hasTlogCluster {
+		return err
```
zaibon: This line could return a nil error

GlenDC: What do you mean nil error? It returns either nil, or an error on this line. I think that's fine, no?

zaibon: if `err` is nil but `hasTlogCluster` is false, we return the nil `err` which is wrong

GlenDC: Why would that be wrong... At that point there is nothing to do anymore as there is no tlog cluster defined, which is fine... If you do want tlog data you define a tlog cluster. I don't get what kind of error you would expect here?

zaibon: indeed, forget it

GlenDC: Ok in that case I guess you want to turn your review into an approval. What did you change by the way in your re-commit of the original commit. I couldn't see a difference with the naked eye with the original commit.

zaibon: just rebase to get the recover removed.

docs/zeroctl/commands/import.md
iwanbk: oops

iwanbk: fixed

docs/zeroctl/commands/import.md
iwanbk: fixed

docs/zeroctl/commands/import.md
iwanbk: fixed

iwanbk:
@GlenDC i've addressed all your remarks

GlenDC:
@iwanbk make sure to resolve #507 ASAP, as it might cause some confusion with the users if those config-defined variables aren't validated.

PR #508, opened 2017-09-29
remove panic recovery

GlenDC:


iwanbk:
curious, what issue this PR solved?

It is not that i'm not agree.

GlenDC:
It is not linked to a PR. Was discussed on telegram with @robvanmieghem to remove.

+ not all panics can be recovered at this high level anyhow (think of panics coming from deep goroutines);
+ we never throw a panic, thus a panic is really only because of a programmer error, so that should be fixed, no reason to recover it;

AFAIK recovering is only ever a good idea in Golang in case we throw a panic explicitly (e.g. somewhere in some deep parsing logic), so we can easily go back up in the stack without having to go via a complicated error path.

PR #514, opened 2017-10-03
add snapshot header feature

### LINKED ISSUE #509: Metadata of an exported vdisk should contain vdisk blocksize and size properties
Currently the vdisk blocksize and size properties must be supplied at the import from the ftp server. If eg an incorrect blocksize is passed, or a size that is smaller than the partitions / filesystems in the vdisk, that will lead to errors.

### LINKED ISSUE #510: validate that snapshot block data does not exceed vdisk size at import
Currently when importing a snapshot into/as a vdisk, we do not validate if there is enough size (as defined by the vdisk staticConfig `size property) to import all the blocks the snapshot contains. This should be validated and return an error in case the given vdisk config defines a too small size.

This issue is part of issue #509.

### LINKED ISSUE #511: add `zeroctl describe snapshot` command
The `zeroctl describe snapshot` command will be used to print out (to the STDOUT) all the information that can be given for a given snapshot, and will thus be heavily related to the import/export vdisk commands.

The output would look something like this:

```javascript
{
     snapshotID: "mybackup",  // required : ID of the snapshot (vdisk backup)
     blockSize: 4096, // required: block size of the snapshot in bytes
     size: 32768, // required: size of the snapshot in bytes (computed in runtime)
     created: "2006-01-02T15:04:05Z07:00", // optional: creation (RFC3339) data, time and timezone
     source: { // optional: info about the source vdisk
            vdiskID: "foo", // optional: ID of the source vdisk
            blockSize: 32768,  // optional: block size of the source vdisk in bytes
            size: 32768,  // optional: block size of the source vdisk in bytes
     },
}
```

This issue is part of issue #509.

### LINKED ISSUE #512: remove `--blockSize` flag from `zeroctl import vdisk` command
Currently the `--blockSize` flag (or shorthand `-b`) has to be specified when importing a backup using the `zeroctl import vdisk`  command. This is the size of the deduped blocks of a snapshot (_not_ the size of the target or source vdisk).

However once we have the more complete Header, to be implemented in #509, which includes both required metadata and optional metadata, we no longer need this information given by the user, as we'll have the information ready in our header as part of the required metadata.

This is great as it will no longer require the user to know or care about the snapshot block size. Making it less error prone to use and thus more user-friendly.

GlenDC:
This PR adds the feature to identify a snapshot (vdisk backup) using a header, rather than storing/loading the deduped map directly. This allows us to add some useful metadata as well as some extra other metadata. While adding this feature and all what goes with it, backwards compatibility with backups made in the past is guaranteed.

+ closes #509;
+ closes #510 (0159798);
+ closes #511 (67af0b5);
+ closes #512 (f223ae6);

While most of the work for this PR is perhaps a bit invisible, it does also add the new `zeroctl describe snapshot` command, which allows you to describe a snapshot for both humans and machines in JSON (pretty) format (and thus make this PR a bit more visible):

```
$ zeroctl describe snapshot mybackup --pretty
```

outputs:

```javascript
{
  	"snapshotID": "mybackup",
  	"blockSize": 131072,
  	"size": 14024704,
  	"created": "2017-10-02T23:48:51-05:00",
  	"source": {
  	  	"vdiskID": "foo",
  	  	"blockSize": 4096,
  	  	"size": 10737418240
  	}
}
```

For machines you would basically omit the `--pretty` flag and get the following output instead: 

```javascript
{"snapshotID":"mybackup","blockSize":131072,"size":14024704,"created":"2017-10-02T23:48:51-05:00","source":{"vdiskID":"foo","blockSize":4096,"size":10737418240}}
```


nbd/ardb/backup/header_test.go
```diff
@@ -0,0 +1,292 @@
...
+// ====
+// CODE COPIED FROM REMOVED (OLD) CODE!!
+// IT IS ONLY HERE TO EMULATE A SERIALIZED AND DIRECTLY STORED DEDUPED MAP!!
+// ====
+
+// Serialize allows you to write all data of this map in a binary encoded manner,
+// to the given writer. The encoded data will be compressed and encrypted before being
+// writen to the given writer.
+// You can re-load this map in memory using the `DeserializeDedupedMap` function.
+func (dm *DedupedMap) Serialize(key *CryptoKey, ct CompressionType, dst io.Writer) error {
```
zaibon: Does this has to be exported ?

GlenDC: It doesn't matter, it's a test file o.O Test files can't export anything, ever.

GlenDC: So what I mean is, there is nothing being exported here. You cannot use it in any other package, not even for testing purposes. As test files cannot export stuff.

zaibon: Hey so that's what I meant, it's never going to be used anywhere else, so for consistency isn't it better to give it a lower case?

GlenDC: It can never be used anywhere else, even if you wanted to use it ever.

It could be lower case, sure... but it's just a direct copy paste of the original code, so I can reproduce serializing a deduped map as we used to do. I really don't think it matters. If it was in a non-test file, sure I would agree with you. We shouldn't expose stuff that isn't exposed. But as test files can't expose AND as this is meant to be a direct copy as possible of the original code, I don't see the harm.

If you want I can change it, yes.

Is this your only comment? Because if so, I'll just merge it straight after that change.

zaibon: It just a styling matter, not that this impact on anything. Just I found easier to understand that this is no going to used anywhere else when we keep same convention, that is for test or real code.

No need to change it if it makes you modify copied/pasted code.

zaibon:
LGTM

nbd/ardb/backup/header_test.go
GlenDC: no, no it's fine. I'll do it. Not a problem.

GlenDC: The goal was more to have the same logic, so we ensure that we can test backwards compatibility, changing a capital doesn't impact that, so that's ok.

PR #515, opened 2017-10-03
Add option to set number of tlog blocks in one flush during zeroctl copy & import

iwanbk:


docs/zeroctl/commands/copy.md
```diff
@@ -26,8 +26,15 @@ Usage:
   zeroctl copy vdisk source_vdiskid target_vdiskid [target_cluster] [flags]
 
 Flags:
-      --config string   zeroctl config file (default "config.yml")
-  -h, --help            help for vdisk
+Flags:
```
GlenDC: typo, you have two times `Flag:`...

docs/zeroctl/commands/copy.md
```diff
@@ -26,8 +26,15 @@ Usage:
...
+      --config SourceConfig   config resource: dialstrings (etcd cluster) or path (yaml file) (default config.yml)
+      --data-shards int       data shards (K) variable of erasure encoding (default 4)
+      --flush-size int        number of tlog blocks in one flush (default 25)
+  -h, --help                  help for vdisk
+  -j, --jobs int              the amount of parallel jobs to run the tlog generator (default $NUMBER_OF_CPUS)
+      --parity-shards int     parity shards (M) variable of erasure encoding (default 2)
+      --priv-key string       private key (default "12345678901234567890123456789012")
+      --same                  enable flag to force copy within the same nbd servers
```
GlenDC: This flag (--same) doesn't exist, so not sure what you are documenting here.

tlog/flusher/flusher.go
```diff
@@ -40,11 +40,13 @@ func New(confSource config.Source, dataShards, parityShards int, vdiskID, privKe
 		return nil, err
 	}
 
-	tlogConf := server.DefaultConfig()
+	if flushSize == 0 {
+		flushSize = server.DefaultConfig().FlushSize
```
GlenDC: I prefer you would make a constant for the `FlushSize` and use that one, rather than generating an entire config variable to just use one of it...

docs/zeroctl/commands/copy.md
iwanbk: i believe it didn't come from nowhere.
```go
https://github.com/zero-os/0-Disk/blob/03c39883892672970942504994fdddeddaf091a0/zeroctl/cmd/copyvdisk/vdisk.go#L175-L177
```
And i'm sure it is not mine

GlenDC: Ahaha, you are right, that is defintely my flag. I remember it now. I was grepping for `--same` but obviously that is not how it is defined in Go code. Ok silly me, so sorry! 👯

PR #516, opened 2017-10-03
nbd : import vdisk also writes tlogStorage metadata

iwanbk:
Parts of #230


iwanbk:
i;ve addressed your remarks

PR #517, opened 2017-10-03
update 0-stor dependency

zaibon:
Cause we need this commit in: https://github.com/zero-os/0-stor/commit/1bb4a2d96ceb1d85a869044079065145db6a336a


GlenDC:
This PR has been replaced by PR #518, due to weirdness in the CI...

PR #518, opened 2017-10-03
update 0-stor dependency

zaibon:


tlog/stor/embeddedserver/stor.go
```diff
@@ -28,7 +28,7 @@ func newZeroStorServer() (*zeroStorServer, error) {
 		return nil, err
 	}
 
-	server, err := server.NewWithDB(db, false)
+	server, err := server.NewWithDB(db, false, 16)
```
GlenDC: Please turn those magical values into constant (can be even local constants just above this function), as it's really not clear to me at all what those values are supposed to do. I had to check the godocs to know for sure.

On top of that you really should document that function better:

https://github.com/zero-os/0-stor/blob/1bb4a2d96ceb1d85a869044079065145db6a336a/server/server.go#L39-L41

As right now it is also not that clear in what unit that size is given. Only after checking the code itself would one know that it is about MiB (note the capital `B` btw).

PR #520, opened 2017-10-04
tlog : validate data+parity against given servers.

### LINKED ISSUE #507: ConfigFromConfigSource should validate data/parity shards
Currently `ConfigFromConfigSource` is not validating if the given data/parity shards actually matches with the amount of specified servers:

https://github.com/zero-os/0-Disk/blob/4f117033758df6e983d9e3b3704dc312c937409f/tlog/stor/config_source.go#L7-L48

It should be validated such that this function will return an error in case the total given data and parity shards don't make sense for the specified amount of servers.

iwanbk:
Fixes #507

### DIFF
--- a/tlog/stor/config_source.go
+++ b/tlog/stor/config_source.go
@@ -1,6 +1,8 @@
 package stor
 
 import (
+	"fmt"
+
 	"github.com/zero-os/0-Disk/config"
 	"github.com/zero-os/0-Disk/log"
 )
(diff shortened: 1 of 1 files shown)


tlog/stor/config_source.go
```diff
@@ -21,6 +23,12 @@ func ConfigFromConfigSource(source config.Source, vdiskID, privKey string, dataS
 		return
 	}
 
+	minServerNum := dataShards + parityShards
+	if minServerNum > len(zsc.Servers) {
```
GlenDC: small nit pick, but I would store `len(zsc.Servers)` in a local variable here, as you're going to reuse it. Not that getting the length is expensive, but still...

iwanbk:
> (thus with 4/2 config)
> 
> I still don't get a complain, even though it should clearly exit with error on startup (of a vdisk), as the amount of servers I specify does not match the 4/2 data/parity config?

Fixed, forgot to `return`

> small nit pick, but I would store len(zsc.Servers) in a local variable here, as you're going to reuse it. Not that getting the length is expensive, but still...

fixed

GlenDC:
> Fixed, forgot to return

I can confirm it does indeed now return an error when using the wrong config.

My config however was correct as expected, so that wasn't my issue. However I don't have the issue anymore of yesterday, it somehow just went away so for now can't say much about what the issue might have been, as I have no clue.

---

Either way, your PR fixes the issue, so merged.

PR #521, opened 2017-10-04
nbdserver : Do not 'panic' in case of error happens.

### LINKED ISSUE #491: a vdisk should never be able to crash its nbdserver
Never should there be a situation where due to an (un)intentional a panic gets triggered which is not caught. It's ok for a single vdisk to crash (as in its connection), but it should never ever crash the entire system (the nbdserver and all its other open vdisk (see: nbd) connections.

Right now there is recover code which aims to catch an uncaught panic originating from an nbd connection (see: a vdisk): https://github.com/zero-os/0-Disk/blob/master/nbd/gonbdserver/nbd/listener.go#L170-L186

This code however is flawed as it turns out that you can't recover (like that) from a panic originating in a spawned goroutine of a spawned goroutine of a...: https://play.golang.org/p/-hrnfaaRFi

This bug was found due to a panic triggered in the tlog storage code: https://github.com/zero-os/0-Disk/blob/master/nbd/nbdserver/tlog/tlog.go#L419-L436, which for now is a critical error as there is no self-healing.

iwanbk:
Fixes #491 
error will be propagated to `Flush()` as proposed at https://github.com/zero-os/0-Disk/issues/491#issuecomment-332424038 and the vdisk exited


nbd/gonbdserver/nbd/connection.go
```diff
@@ -433,6 +433,7 @@ func (c *Connection) receive(ctx context.Context) {
 				if err := c.backend.Flush(ctx); err != nil {
 					c.logger.Infof("Client %s got flush I/O error: %s", c.name, err)
 					nbdRep.NbdError = errorCodeFromGolangError(err)
+					return
```
GlenDC: you can't just return here, that's wrong. We need to send the Reply even in an error case.

nbd/gonbdserver/nbd/connection.go
```diff
@@ -469,13 +470,15 @@ func (c *Connection) receive(ctx context.Context) {
...
```
GlenDC: you can't just return here, that's wrong. We need to send the Reply even in an error case.

nbd/gonbdserver/nbd/connection.go
```diff
@@ -469,13 +470,15 @@ func (c *Connection) receive(ctx context.Context) {
...
 				}
 			}
 
 		case NBD_CMD_FLUSH:
 			if err := c.backend.Flush(ctx); err != nil {
 				c.logger.Infof("Client %s got flush I/O error: %s", c.name, err)
 				nbdRep.NbdError = errorCodeFromGolangError(err)
+				return
```
GlenDC: you can't just return here, that's wrong. We need to send the Reply even in an error case.

nbd/nbdserver/tlog/tlog.go
```diff
@@ -396,15 +408,13 @@ func (tls *tlogStorage) transactionSender(ctx context.Context) {
...
-				}
-				panic(fmt.Errorf(
+				tls.flushErrCh <- fmt.Errorf(
 					"tlogStorage couldn't send block %d (seq: %d): %s",
 					transaction.Index,
 					transaction.Sequence,
 					err,
-				))
+				)
+				return
```
GlenDC: Why is this the only location where you exit out of a function/goroutine when sending an error over the `flushErrCh`?

nbd/nbdserver/tlog/tlog.go
```diff
@@ -365,9 +377,9 @@ func (tls *tlogStorage) spawnBackgroundGoroutine(ctx context.Context) error {
 				case tlog.BlockStatusDisconnected:
 					log.Info("tlog connection disconnected")
 				default:
-					panic(fmt.Errorf(
-						"tlog server had fatal failure for vdisk %s: %s",
-						tls.vdiskID, res.Resp.Status))
+					log.Errorf(
+						"got unknown response type for vdisk %s: %s",
```
GlenDC: that's a much better error message indeed

nbd/gonbdserver/nbd/connection.go
iwanbk: Hmm...i thought it already sent the reply because there is this line above
```go
nbdRep.NbdError = errorCodeFromGolangError(err)
```
What reply we should sent?

GlenDC: If you return that reply isn't sent he... Look down below the code, it will send a reply there:

https://github.com/zero-os/0-Disk/blob/328c5ea5e6d65b25f941bae2eb2ca08eed70121d/nbd/gonbdserver/nbd/connection.go#L523-L527

The line you copy is just setting the reply, but if you return before, that reply will actually not be send AFAIK.

nbd/nbdserver/tlog/tlog.go
iwanbk: was a mistake

iwanbk:
i addressed your comment about not sending reply on flush error

iwanbk:
Any other thing to do regarding this PR?

GlenDC:
> Any other thing to do regarding this PR?

I think it looks good, I just need to confirm something with regarding your change to the gonbdserver, as I want to make sure that still confirms to what the specs says.

Will do that one of the first work things I do tomorrow.

GlenDC:
@iwanbk what you did in general in the gonbdserver seems OK-ish. However, I would like to know where your desire to suddenly change this file come from? Can you please backup with some comments on why you did this?

On top of that:

+ Why you set the `run` variable at `false` when it's a flush error, but not when it's a `WriteAt` error (just to give an example), in general you seem to be only using your new logic for `flushing`, why is this?

Not sure if it's a good idea to have this new run variable, if you have a good reason to want to stop receiving after receiving an error (and I'm not sure if is), it might be a better idea to cancel the `run` variable usage and thus go back to a plain `for {}`, and instead change https://github.com/zero-os/0-Disk/blob/9499d13ab2e8ff87ef5ad0478f1b61fa9ab63286/nbd/gonbdserver/nbd/connection.go#L524 to:

```go
!c.sendHeader(ctx, nbdRep) || nbdRep.NbdError != 0 
```

This way you would catch all errors, without forgetting now one, or forgetting it when adding such a new error path in the future.

But again, first I would love to know where this desire to stop receiving after replying an error comes from. There is no official spec, however in https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md#request-types, I don't read anywhere that we need to hard-exit after an error.

iwanbk:
> But again, first I would love to know where this desire to stop receiving after replying an error comes from. There is no official spec, however in https://github.com/NetworkBlockDevice/nbd/blob/master/doc/proto.md#request-types, I don't read anywhere that we need to hard-exit after an error.

Well, it actually came from our discussion as your suggestion at https://github.com/zero-os/0-Disk/issues/491#issuecomment-332837822

> No we don't. For now there is no self-healing, so no point in not exiting yet. Once self-healing is in-place we won't have to exit as there will be plenty of things we can try before doing so. But right now there isn't anything we can do in these critical error cases, and thus exiting is all we can do.

Except that i misunderstood you

GlenDC:
> Except that i misunderstood you

Oh, if that's the case and your motivation for these changes than I would just suggest to revert your changes to gonbdserver.

iwanbk:
> Oh, if that's the case and your motivation for these changes than I would just suggest to revert your changes to gonbdserver.

done

nbd/nbdserver/tlog/tlog.go
```diff
@@ -396,15 +408,12 @@ func (tls *tlogStorage) transactionSender(ctx context.Context) {
 				transaction.Content,
 			)
 			if err != nil {
-				if err == tlogclient.ErrClientClosed {
-					return
-				}
-				panic(fmt.Errorf(
+				tls.flushErrCh <- fmt.Errorf(
```
GlenDC: as we only read from this `tls.flushErrCh` during a `Flush` call, and it might take quite a while before the `Flush` call gets called, it could be fairly possible that this goroutine becomes blocked on this channel in case enough flush errors occur at this stage before a `Flush` call gets called.

If that's so than also the `(tls *tlogStorage) set` could become blocked, as that channel would get full if the transaction sender goroutine is no longer reading from that channel.

Did you take this all into account? As all that would be pretty bad, and could give the program a halting problem, or am I missing something here?

iwanbk: Yeah, we already discussed this on the related GH issue.
In short, we have 3 options:
- exit 
- this way (that you commented above)
- wait for self healing  (you said that we don't have this mechanism yet)

I prefer to exit, it is cleaner.
All of them are not ideal, but are better than panic

GlenDC: Yeah sure. For now I'm ok to have it return an error only in Flush, but that doesn't mean you can simply program it so that it can lock forever on some goroutines...

If you have a clean way to exit, with an error propagated all the way down to the gonbdserver module, than I'm all ears. A panic is certainly a no-go.

> wait for self healing (you said that we don't have this mechanism yet)

There is no reason to wait. Because even when we have self-healing, there are still scenarios where we can't self-heal (e.g. all layers are having troubles), and thus we still need to simply exit with an error, just like we're doing here. So basically the work you're doing here, handles the worst case scenario, for when self-healing did not work (which is similar to not being implemented at all).

iwanbk: > A panic is certainly a no-go.

Yeah, it affects all vdisk, not only this one

GlenDC: That and the fact that we should simply not use panic, it's not go-idiomatic either. That's more a java and C# paradigm, where throw-catch approaches are common. In Golang a panic should only be used for something really exceptional.

GlenDC:
@iwanbk you're still working on this PR?

iwanbk:
> @iwanbk you're still working on this PR?

I actually waits for your proposal which improve/replace my proposal, as the issue is assigned to both of us.

GlenDC:
Didn't know you were waiting for me. If so I'll give it some taught this week, to see if I can come up with something that fits in the current tlog design.

In the meanwhile I've merged current master into this branch.

GlenDC:
I was thinking, and I do not think we want to trigger a critical error because tlogservers are no longer reachable. Which is also another reason why we shouldn't panic here.

The tlog module is only the third (and final) layer, used in 0-Disk to provide data storage and redundancy. The primary cluster being the first layer, and the slave cluster the second layer. So I think it's a bit crazy to shut down an entire vdisk just because a final layer is no longer (temporarily) running, even though it still might have 2 fully operational layers before that, and might not need this third layer any time soon.

While it is true that it's bad that we might have a gap in our transaction data, it doesn't have to mean that the tlog for this vdisk remains corrupt. I think we could follow an approach as follows:

+ Somehow register that the tlog for a vdisk has stopped due to an error (and thus this is the start of the gap);
+ When a tlogserver comes back online (or we add a new tlogserver), we will not be able to fill that gap exactly as it happened, but we'll still be able to track any information that should have been tracked while we couldn't (all this information could simply share a timestamp, the timestamp where the tlogserver started working again):
   1. Gather all block indices for a given vdisk;
   2. For each block index:
      a. if the block index doesn't exist yet in the tlog transactions, than simply add it (it's a new one):
      b. if it does exist:
         1. fetch the data from the storage (using the block storage);
         2. compare the data with the last data we tracked on the tlog transactions, if it's different, track this modification;
    3. While we do all this we should also track any block indices that are no longer existing in the block storage, but do still exist in the latest snapshot received from the tlog transactions. Those should all be deleted now;

By doing this, we can ensure the tlog transactions is back up to date with the latest vdisk state, as soon as we recovered the tlog processes. Obviously there is a gap of missing information, so sadly you can't roll back to a specific point within that gap. But at the same time, this was never a critical point of failure.

The only reason when this should become a critical point of failure, is if we suddenly require a valid layer 3 (tlog data), because both layer 1 and layer 2 have failed, while layer 3 is also in fail-mode (as described above). If that happens, than we are truly fucked, and only than it should count as a critical error.

iwanbk:
interesting point @GlenDC, will thin about it.

But please remember, this issue was reported by azmy when the nbdserver crashed because of `ardb` crashed/dead.

So, there is other things than tlog here.

iwanbk:
> By doing this, we can ensure the tlog transactions is back up to date with the latest vdisk state, as soon as we recovered the tlog processes.

not reallu sure about it, i think it deserves new issue because it seems quite complex and not really related with this issue.


> Obviously there is a gap of missing information, so sadly you can't roll back to a specific point within that gap. But at the same time, this was never a critical point of failure.

I don't think that it is the right way for tlog to not synced with primary cluster.
What if the missing transactions is a very important info, like some metadata? We might corrupted a lot of data.

i can only think for us to:
- prolong the time for tlog to be considered failed, create a proposal for it at #605
- regenerate tlog data, as what we do in `zeroctl copy`

iwanbk:
In the meanwhile, i think we can focus on the ardb think first.
Because it has quite clear spec : should not panic, should not become critical error. I will send another PR for it.

GlenDC:
Closing this PR, as it's not a solution we desire for the current master branch.

On top of that it won't be needed anymore once we switch over to the standalone tlog repo. We'll keep the branch alive for now, for reference reasons, will delete it in the future once we really know for sure we won't need it anymore.

PR #528, opened 2017-10-05
Adds watching updates on Tlog cluster ID

### LINKED ISSUE #526: config VdiskNBDConfig is not watched for updates in tlog storage
Steps:
- start nbdserver with file config
- start tlog server
- start a VM
- change `tlogServerClusterID`, to point to new tlog cluster

expected:
 tlog client connect to new cluster. an easy way to check this: if happens, it will print this kind of log, with new server address 
```
tlogclient vdisk 'myimg' change server addrs to '[localhost:11233]' module=global caller=client.go:270
```

It seems to be `config` issue because it worked when i change server address directly and not changing `tlogServerClusterID`

chrisvdg:
fixes #526

### DIFF
--- a/nbd/nbdserver/tlog/tlog.go
+++ b/nbd/nbdserver/tlog/tlog.go
@@ -433,14 +433,8 @@ func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, sou
 		return nil, err
 	}
 
-	var vdiskNBDConfig config.VdiskNBDConfig
-	select {
-	case vdiskNBDConfig = <-vdiskNBDUpdate:
-		tls.tlogClusterID = vdiskNBDConfig.TlogServerClusterID
-	case <-ctx.Done():
-		cancel()
-		return nil, errors.New("tlogStorage requires active context")
-	}
+	vdiskNBDConfig := <-vdiskNBDUpdate
+	tls.tlogClusterID = vdiskNBDConfig.TlogServerClusterID
 
 	tlogCfgCtx, tlogCfgCancel := context.WithCancel(ctx)
 	tlogClusterCfgUpdate, err := config.WatchTlogClusterConfig(tlogCfgCtx, source, vdiskNBDConfig.TlogServerClusterID)
@@ -460,7 +454,13 @@ func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, sou
 	go func() {
 		for {
 			select {
-			case vdiskNBDConfig := <-vdiskNBDUpdate:
+			case vdiskNBDConfig, ok := <-vdiskNBDUpdate:
+				// in case channel is closed
+				// but ctx.Done() wasn't triggered yet
+				if !ok {
+					return
+				}
+
 				if tls.tlogClusterID != vdiskNBDConfig.TlogServerClusterID {
 					if vdiskNBDConfig.TlogServerClusterID == "" {
 						source.MarkInvalidKey(config.Key{ID: vdiskID, Type: config.KeyVdiskNBD}, vdiskID)
@@ -471,8 +471,8 @@ func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, sou
 						continue
 					}
 
-					newTlogCfgCtx, newTlogCfgCancel := context.WithCancel(ctx)
-					newTlogClusterCfgUpdate, err := config.WatchTlogClusterConfig(newTlogCfgCtx, source, vdiskNBDConfig.TlogServerClusterID)
+					tlogCfgCtx, newTlogCfgCancel := context.WithCancel(ctx)
+					newTlogClusterCfgUpdate, err := config.WatchTlogClusterConfig(tlogCfgCtx, source, vdiskNBDConfig.TlogServerClusterID)
 					if err != nil {
 						log.Errorf(
 							"vdisk %s encountered an error switching to Tlog cluster ID '%s': %s",
@@ -484,7 +484,6 @@ func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, sou
 
 					log.Debugf("Setting Tlog cluster ID to: %s", vdiskNBDConfig.TlogServerClusterID)
 					tlogCfgCancel()
-					tlogCfgCtx = newTlogCfgCtx
 					tlogCfgCancel = newTlogCfgCancel
 					tls.tlogClusterID = vdiskNBDConfig.TlogServerClusterID
 					tlogClusterCfgUpdate = newTlogClusterCfgUpdate
--- a/nbd/nbdserver/tlog/tlog.go
+++ b/nbd/nbdserver/tlog/tlog.go
@@ -31,6 +31,7 @@ func Storage(ctx context.Context, vdiskID string, configSource config.Source, bl
 		return nil, errors.New("tlogStorage requires a valid metadata ARDB provider: " + err.Error())
 	}
 
+	ctx, cancel := context.WithCancel(ctx)
 	tlogStorage := &tlogStorage{
 		vdiskID:        vdiskID,
 		storage:        storage,
@@ -40,9 +41,9 @@ func Storage(ctx context.Context, vdiskID string, configSource config.Source, bl
 		transactionCh:  make(chan *transaction, transactionChCapacity),
 		toFlushCh:      make(chan []uint64, toFlushChCapacity),
 		cacheEmptyCond: sync.NewCond(&sync.Mutex{}),
-		done:           make(chan struct{}, 1),
+		cancel:         cancel,
 	}
-	ctx, cancel := context.WithCancel(ctx)
+
 	tlogClusterConfig, err := tlogStorage.tlogRPCReloader(ctx, vdiskID, configSource)
 	if err != nil {
 		cancel()
@@ -103,7 +104,7 @@ type tlogStorage struct {
 	// condition variable used to signal when the cache is empty
 	cacheEmptyCond *sync.Cond
 
-	done chan struct{}
+	cancel context.CancelFunc
 }
 
 type transaction struct {
@@ -280,7 +281,7 @@ func (tls *tlogStorage) Close() (err error) {
 	defer tls.storageMux.Unlock()
 
 	log.Infof("tlog storage closed with cache empty = %v", tls.cache.Empty())
-	close(tls.done)
+	tls.cancel()
 
 	err = tls.storage.Close()
 	if err != nil {
@@ -298,8 +299,8 @@ func (tls *tlogStorage) Close() (err error) {
 	return
 }
 
-func (tls *tlogStorage) spawnBackgroundGoroutine(ctx context.Context) error {
-	ctx, cancelFunc := context.WithCancel(ctx)
+func (tls *tlogStorage) spawnBackgroundGoroutine(parentCtx context.Context) error {
+	ctx, cancelFunc := context.WithCancel(parentCtx)
 	recvCh := tls.tlog.Recv()
 
 	// used for sending our transactions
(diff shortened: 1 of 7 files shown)


nbd/nbdserver/tlog/tlog.go
```diff
@@ -115,6 +104,9 @@ type tlogStorage struct {
 	cacheEmptyCond *sync.Cond
 
 	done chan struct{}
+
+	// for testing
+	tlogClusterIDUpdate chan struct{}
```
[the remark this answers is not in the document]
chrisvdg: 5ms already seemed plenty of time on local machine and Travis for the update to propagate so set the polling rate to 5ms

nbd/nbdserver/tlog/tlog.go
```diff
@@ -436,25 +428,79 @@ func (tls *tlogStorage) flusher(ctx context.Context) {
 
 }
 
-func (tls *tlogStorage) goTlogRPCReloader(ctx context.Context, ch <-chan config.TlogClusterConfig, cancel func()) {
-	defer cancel()
+func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, source config.Source) (*config.TlogClusterConfig, error) {
+	reloaderCTX, reloaderCancel := context.WithCancel(ctx)
```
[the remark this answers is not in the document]
chrisvdg: I can't defer cancel as that would close the config updating goroutine created by this function as soon as it returns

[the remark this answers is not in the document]
chrisvdg: And yes I want to cancel it separately as that will close the watch for the old tlog cluster config  when the tlog cluster ID has changed

nbd/nbdserver/tlog/tlog.go
```diff
@@ -436,25 +428,79 @@ func (tls *tlogStorage) flusher(ctx context.Context) {
...
+
+	go func() {
+		for {
+			select {
+			case vdiskNBDConfig := <-vdiskNBDUpdate:
+				if tls.clusterID != vdiskNBDConfig.TlogServerClusterID {
+					// shut down current TlogClusterConfig watcher and start a new one with new Tlog cluster ID
+					tlogCfgCancel()
+					tlogCfgCtx, tlogCfgCancel = context.WithCancel(reloaderCTX)
+					newTlogClusterCfgUpdate, err := config.WatchTlogClusterConfig(tlogCfgCtx, source, vdiskNBDConfig.TlogServerClusterID)
```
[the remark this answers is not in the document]
chrisvdg: Any idea for the error code? and what about the data for the error broadcast? add the vdiskID?

nbd/nbdserver/tlog/tlog.go
```diff
@@ -462,9 +462,17 @@ func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, sou
 			select {
 			case vdiskNBDConfig := <-vdiskNBDUpdate:
 				if tls.tlogClusterID != vdiskNBDConfig.TlogServerClusterID {
+					// if new TlogServerClusterID broadcast error
+					if vdiskNBDConfig.TlogServerClusterID == "" {
+						// TODO: log to AYS
```
GlenDC: In a way you could just use 403 (InvalidConfig), but the config isn't valid so that doesn't seem right, even though in our eyes it's invalid. Perhaps you can make a new status code 405 (UnexpectedConfig)? And yes you should also add the vdiskID to the data of that broadcast, as it has everything to do that for that vdisk we were using tlog and now suddenly no tlogClusterID is specified.

nbd/nbdserver/tlog/tlog.go
```diff
@@ -462,9 +462,17 @@ func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, sou
 			select {
 			case vdiskNBDConfig := <-vdiskNBDUpdate:
 				if tls.tlogClusterID != vdiskNBDConfig.TlogServerClusterID {
+					// if new TlogServerClusterID broadcast error
```
GlenDC: this comment doesn't seems to match what the code is doing

nbd/nbdserver/tlog/tlog.go
```diff
@@ -474,20 +482,17 @@ func (tls *tlogStorage) tlogRPCReloader(ctx context.Context, vdiskID string, sou
...
 				}
 			case tlogClusterCfg := <-tlogClusterCfgUpdate:
-				if len(tlogClusterCfg.Servers) == 0 {
-					// TODO: notify 0-orchestrator
-					log.Errorf(
-						"nbd config for vdisk %s no longer specifies tlog rpcs", tls.vdiskID)
-					continue
-				}
-
+				log.Debug("tlog update")
```
GlenDC: if you want to add a debug statement, please make it more meaningful, as this one doesn't say much. It could add some more context about the vdiskID it is doing this for, and so forth.

nbd/nbdserver/tlog/tlog.go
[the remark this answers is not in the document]
GlenDC: https://github.com/zero-os/0-Disk/pull/528#discussion_r143503269

nbd/nbdserver/tlog/tlog.go
```diff
@@ -436,25 +425,81 @@ func (tls *tlogStorage) flusher(ctx context.Context) {
...
+							"vdisk %s encountered an error switching to Tlog cluster ID '%s': %s",
+							tls.vdiskID,
+							vdiskNBDConfig.TlogServerClusterID,
+							err)
+						continue
+					}
+
+					log.Debugf("Setting Tlog cluster ID to: %s", vdiskNBDConfig.TlogServerClusterID)
+					tlogCfgCancel()
+					tlogCfgCtx = newTlogCfgCtx
```
GlenDC: no reason to store the context, you can and probably should shadow it, that is not a problem. The context is only ever used once, to create this cluster watcher, so no reason to keep it in memory. Overshadowing it will make it possible for the runtime to release this variable somewhere after the case body has reached its ending.

nbd/nbdserver/tlog/tlog.go
```diff
@@ -436,25 +425,81 @@ func (tls *tlogStorage) flusher(ctx context.Context) {
...
+					tlogCfgCancel = newTlogCfgCancel
+					tls.mux.Lock()
+					tls.tlogClusterID = vdiskNBDConfig.TlogServerClusterID
```
GlenDC: Please remove the `tls.mux.Lock()` and `tls.mux.Unlock()` statements. The only thing you are updating here is the `tls.tlogClusterID`, which is only ever used within this goroutine, and only one case at the same time can be active. So there is really no possibility of a race here.

GlenDC: The real update happens later, and is protected by its own lock inside the `tls.tlog.ChangeServerAddresses(tlogClusterCfg.Servers)` call.

nbd/nbdserver/tlog/tlog.go
GlenDC: The only variables you want to re-assign is `tlogCfgCancel`, `tls.tlogClusterID`, and `tlogClusterCfgUpdate`.

nbd/nbdserver/tlog/tlog_test.go
```diff
@@ -941,3 +946,120 @@ func newZeroStorConfig(t *testing.T, vdiskID string, tlogConf *server.Config) (*
...
+	})
+
+	// wait for update to propagate
+	timeoutTicker := time.NewTicker(30 * time.Second)
+	pollTicker := time.NewTicker(5 * time.Millisecond)
+	changed := false
+	for !changed {
+		select {
+		case <-pollTicker.C:
+			tlogStorage.mux.Lock()
```
GlenDC: Aha, now I get why you are adding a mutex lock earlier around `tls.tlogClusterID = vdiskNBDConfig.TlogServerClusterID`, haha. This is bad, you are only locking probably because you received a race warning triggered by your test code here.

Production code should really not suffer in performance because of a unit test, especially a non-critical test such as this one.

I think you can prevent it by not checking for the tlogClusterID. Instead you could in your polling check directly for `logClient.getServers()`, as the client (`stubTlogClient`) methods are anyhow protected by their own locks and are thus goroutine safe as it is.

GlenDC: It means you are not directly testing if `tlogClusterID` is updated, but that's OK I think. The only thing you really care about right now is that the servers get updated multiple times. You probably do want to do not just one update, but a couple of updates, such that you're sure it keeps working, e.g.:

+ switch `foo` to -> `bar` and test;
+ switch to empty string (should not do anything), no reason to test here;
+ switch from `bar` to -> `baz` and test;
+ switch from `baz` to -> `foo` and test;

Something like that. Obviously you can change the names if you want, or do something else, but the important thing is that you have multiple (sequential) updates, and not just one.

GlenDC: There are certainly ways to also test more detailed and local for your tlogClusterID, but that would require you to refactor some of your code into its own structure and separate out the logic that way, but for now the code is really OK enough as it is. No reason to spend too much time on this or now. So just ignore this comment for any practical purposes.

nbd/nbdserver/tlog/tlog.go
chrisvdg: Not completely sure what you mean by overshadowing

nbd/nbdserver/tlog/tlog_test.go
```diff
@@ -941,3 +946,138 @@ func newZeroStorConfig(t *testing.T, vdiskID string, tlogConf *server.Config) (*
...
+						changed = true
+					}
+				} else if !cluster.invalid {
+					if len(s) > 0 && s[0] != clusters[i-1].servers[0] {
+						changed = true
+					}
+				} else {
+					// wait to make sure the update has propagated
+					// invalid configs don't update so no way to check if propagated
+					time.Sleep(10 * time.Millisecond)
```
GlenDC: this sleep is really not needed. I mean after all, that's why you have the `pollTicker.C`, no?

nbd/nbdserver/tlog/tlog_test.go
```diff
@@ -941,3 +946,138 @@ func newZeroStorConfig(t *testing.T, vdiskID string, tlogConf *server.Config) (*
...
+		// change tlog server cluster
+		source.SetTlogServerCluster(vdiskID, cluster.clusterID, &config.TlogClusterConfig{
+			Servers: cluster.servers,
+		})
+
+		// wait for update to propagate
+		timeoutTicker := time.NewTicker(30 * time.Second)
+		pollTicker := time.NewTicker(5 * time.Millisecond)
+		changed := false
+		for !changed {
```
GlenDC: No reason to use a `changed` bool variable here, in fact it's really not something you should do at all. Just do a regular `for {}` loop and break.

I suppose you do this because either you knew, or you knew by failing this test, that you cannot just use break within your select body, as it would break out only out that select body? Well for that purpose you can make something labeled in go. For this exact situation there is even a blog explaining it: https://www.goinggo.net/2013/11/label-breaks-in-go.html

nbd/nbdserver/tlog/tlog_test.go
```diff
@@ -941,3 +946,138 @@ func newZeroStorConfig(t *testing.T, vdiskID string, tlogConf *server.Config) (*
...
+func TestTlogSwitchClusterID(t *testing.T) {
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	defer cancelFunc()
+
+	const (
+		vdiskID   = "a"
+		blockSize = 8
+	)
+
+	clusters := []switchCluster{
```
GlenDC: The only thing you didn't test if it also still works if you update just a tlog cluster config in use, and not also the clusterID (the functionality which was already supported).

When adding a feature to an already existing codebase, it's good to ensure that the old features still work. Normally there are already written unit tests for this. But I don't think there was code yet that was testing the tlogStorage's config hotreloading feature. So now that you are writing that test anyhow, it would be good to also guarantee that the original hot reload feature also still works.

PR #530, opened 2017-10-06
tlog : single flusher

### LINKED ISSUE #452: tlog : vdisk flusher should use same flusher as tlog generator
A WIP tlog generator has been added as part of #425.
I have created a simple flusher for it.

Both tlog generator and vdisk flusher should share this flusher

iwanbk:
Fixes #452 
Use single flusher for both tlog generator and tlog server

### DIFF
--- a/tlog/aggregation.go
+++ b/tlog/aggregation.go
@@ -1,0 +1,165 @@
+package tlog
+
+import (
+	"bytes"
+	"errors"
+
+	"zombiezen.com/go/capnproto2"
+
+	"github.com/zero-os/0-Disk/tlog/schema"
+)
+
+var (
+	// ErrExceedBlockNumber returned in case user tries to
+	// add block to a full aggregation
+	ErrExceedBlockNumber = errors.New("exceed blocks number")
+
+	// ErrEmptyAggregation returned in case the aggregation is empty
+	// but user tries to do operation for non-empty aggregation
+	ErrEmptyAggregation = errors.New("empty aggregation")
+)
+
+// Aggregation defines a tlog aggregation
+type Aggregation struct {
+	agg          schema.TlogAggregation
+	msg          *capnp.Message
+	blockList    schema.TlogBlock_List
+	lastSequence uint64
+	size         int
+	maxBlockNum  int
+}
+
+// NewAggregation creates an aggregation with given capnp buffer
+// and maximal block number
+func NewAggregation(capnpBuf []byte, maxBlockNum int) (*Aggregation, error) {
+	// create the aggregation object
+	msg, seg, err := capnp.NewMessage(capnp.SingleSegment(capnpBuf))
+	if err != nil {
+		return nil, err
+	}
+	agg, err := schema.NewRootTlogAggregation(seg)
+	if err != nil {
+		return nil, err
+	}
+	blockList, err := agg.NewBlocks(int32(maxBlockNum))
+	if err != nil {
+		return nil, err
+	}
+	return &Aggregation{
+		agg:         agg,
+		msg:         msg,
+		blockList:   blockList,
+		maxBlockNum: maxBlockNum,
+	}, nil
+}
+
+// Empty returns true if the aggregation is empty
+func (a *Aggregation) Empty() bool {
+	return a.size == 0
+}
+
+// Full returns true if the aggregation is full
+func (a *Aggregation) Full() bool {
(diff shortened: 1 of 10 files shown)


tlog/copy/copier_test.go
```diff
@@ -178,10 +179,22 @@ func testCopy(t *testing.T, confSource config.Source, dataShards, parityShards,
...
+			Operation: schema.OpSet,
+			Sequence:  seq,
+			Content:   data,
+			Index:     idx,
+			Timestamp: tlog.TimeNowTimestamp(),
+			Hash:      zerodisk.Hash(data),
+		})
 		require.NoError(t, err)
+
+		if flusher.Full() {
```
GlenDC: Is there a good reason why the `AddTransaction is not doing the `If-Full-Then-Flush` cycle implicitly on each call? Seems a bit cumbersome to have to do this ourselves as a user at all times.

tlog/aggregation.go
```diff
@@ -0,0 +1,165 @@
...
+	for i := 0; i < a.size; i++ {
+		if a.blockList.At(i).Sequence() >= seq {
+			break
+		}
+		idx = i
+	}
+	if idx < 0 {
+		return nil
+	}
+	a.size = a.size - (idx + 1)
```
GlenDC: Is this is a style you guys prefer? I saw @zaibon doing something similar not long ago.

I guess you know you could do just `a.size -= idx + 1`, so what's up with these lines, why you like it to do like this? Just a curious question...

tlog/aggregation.go
```diff
@@ -0,0 +1,165 @@
...
+	// add block to a full aggregation
+	ErrExceedBlockNumber = errors.New("exceed blocks number")
+
+	// ErrEmptyAggregation returned in case the aggregation is empty
+	// but user tries to do operation for non-empty aggregation
+	ErrEmptyAggregation = errors.New("empty aggregation")
+)
+
+// Aggregation defines a tlog aggregation
+type Aggregation struct {
```
GlenDC: I don't fully understand why you keep your data in the schema/capnp given objects, while they might not be really ideal for your purpose.

For example, it seems that for the blockList it would make more sense to keep it in a linked list.

It would also mean you don't really have to keep stuff like `msg` in memory.

But perhaps you have very good reasons to keep it in these structures?

tlog/aggregation.go
```diff
@@ -0,0 +1,165 @@
...
+		agg:         agg,
+		msg:         msg,
+		blockList:   blockList,
+		maxBlockNum: maxBlockNum,
+	}, nil
+}
+
+// Empty returns true if the aggregation is empty
+func (a *Aggregation) Empty() bool {
+	return a.size == 0
```
GlenDC: are you 100% sure size is will never become negative at any point?

tlog/aggregation.go
GlenDC: Well even a linked list might not be ideal.

My bigger point was more that it might be a better idea to keep those things in "specialized" structures, considering its in-memory use case.

tlog/transaction.go
```diff
@@ -0,0 +1,11 @@
+package tlog
+
+// Transaction defines a tlog transaction
+type Transaction struct {
+	Operation uint8
```
GlenDC: Not just this case in specific (but definitely also this case), but I would definitely recommend to start giving explicit types to these kind of enum-based values. Such that you can define a typed enum, which makes it much clearer what exactly this property is, as well as easier and less error-prone to use.

tlog/copy/copier_test.go
iwanbk: It was,but current `Flush` has different return values `AddTransaction`

tlog/aggregation.go
iwanbk: there is no specific reason. it just looks more clear for me

tlog/aggregation.go
iwanbk: By keeping it in capnp given object, i only allocate memory to hold the full aggregation once.

If we keep it in our own data structures, we need to allocate it twice:
- keep transactions in our memory
- keep transactions in capnp

tlog/aggregation.go
iwanbk: yep

tlog/transaction.go
iwanbk: you're right

PR #536, opened 2017-10-07
Wait for tlog  to be ready

### LINKED ISSUE #527: tlog : sync with other active tlog server
We already have support for multiple tlog servers, but only one tlog server active at a time.

We have situation where we have two active tlog servers for same vdisk, in limited amount of time. 

The scenario is Live Migration:

```

nbdserver1, tlogserver1 = before migration starts, they will go away after migration
nbdserver2, tlogserver2 = after migration, they will continue after migration

step                    | host w                | host x      | host y            | host z
------------------------+-----------------------+-----------------------------------------
before migration status | qemu1, nbdserver1     | tlogserver1 | nbdserver2        | tlogserver2 (idle) listening to tlogserver1
                        |                       |             |                   | waiting for taking over 
------------------------+-----------------------+-------------+-------------------+-------
migrating status        | qemu1 waits for sync  | tlogserver1 | qemu2, nbdserver2 | tlogserver2 (idle) listening to tlogserver1
                        | from nbdserver1, then |             |                   | waiting for taking over 
                        | switches to qemu2 on  |             |                   |
                        | host y                |             |                   |
------------------------+-----------------------+-------------+-------------------+-------
after migration         |                       | tlogserver1 | qemu2, nbdserver2 | tlogserver2 takes over                  
                        |                       | flushes last|
                        |                       | block and   |
                        |                       | notifies    |
                        |                       | tlogserver2 |
                        |                       | to take over|
```

In this scenario, both tlog servers could be active but there is **only one** that can do the flush at a time.

### coordination between tlog servers
TSX = tlog server X
NBDX = NBD server X

**migrating status**
TS2 connect to TS1.
TS2 can safely take over and do flush if failed to connect to TS1

**after migration**
0-orch send SIGTERM to NBD1 and then TS1.
TS2 detect that TS1 is down, TS2 can take over and do the flush

## what TS1 need to do
simply open a TCP listener (for TS2 to open TCP socket) or HTTP server (for TS2 to poll).
tcp socket seems to be more efficient but HTTP is easier to use and has more builtin error handling.
polling in this case also not really bad:
- for limited amount of time
- only one connection from TS2

## what TS2 need to do
- open tcp socket or do http poll to TS1, as described in "coordination" section
- hold all flush until meet one of these conditions:
     - detect that TS1 dies
     - ~~reach timeout  (configurable)~~ (see comment from @FastGeert below)
     - ~~reach X number of transactions (configurable)~~ (see comment from @FastGeert below)

- load last sequence from 0-stor
- if needed, rewrite the sequence number of all received blocks
- after take over, inform NBD2 to use new "last sequence". As temporary solution, TS2 could use response from TS2 to NBD2 to inform the new last sequence

iwanbk:
Fixes #527 

Solution:
- on startup, tlogserver 1 (t1) open TCP port specified by `--wait-listen-addr`
- on startup, t2 connect to port `--wait-connect-addr` of t1. This port must be the same value as `--wait-listen-addr` of t1:
       - if t2 can't connect, t2 mark itself as ready
       - if t2 can connect:

                   -  t2 wait for t1 to dies. As indicated by closed TCP connection
                   - **TODO**  add connect retry mechanism
                   - **TODO** add mechanism for t1 to tell t2 about it's last sequence
- nbd2 connect to t2
- nbd2 use tlogclient 2 (c2)  to get readyness status of t2. If it not ready:
         - start goroutine to wait for t2 to be ready
         - hold all write&delete IO in a slice buffer 
         - after t2 ready, nbd2 do the on hold operation


GlenDC:
The branch of this PR also has conflicts with master which will have to be resolved as well.

nbd/nbdserver/tlog/tlog.go
```diff
@@ -125,11 +132,27 @@ type transaction struct {
...
+	content    []byte
+}
+
 // SetBlock implements BlockStorage.SetBlock
 func (tls *tlogStorage) SetBlock(blockIndex int64, content []byte) error {
 	tls.mux.Lock()
 	defer tls.mux.Unlock()
 
+	// insert to buffer if tlog not ready yet
+	tls.tlogReadyMux.Lock()
```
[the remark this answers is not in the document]
iwanbk: > I do not think that your approach is good here

Yeah, indeed. I couldn't find better approach back then.
Let me know if you know better way to do this

> You are skipping the call to tls.set, which means the content is not added to the cache, and thus > isn't available in a sequentialGetBlock call. Which is pretty bad...


you're right

nbd/nbdserver/tlog/tlog.go
```diff
@@ -215,18 +238,40 @@ func (tls *tlogStorage) DeleteBlock(blockIndex int64) (err error) {
...
+
 	err = tls.set(blockIndex, nil)
 	return
 }
 
 // Flush implements BlockStorage.Flush
 func (tls *tlogStorage) Flush() error {
+	// wait for tlog to be ready
+	// TODO : find a way without waiting
+	for {
```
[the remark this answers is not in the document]
iwanbk: > Because as far as I understand the code, you'll be stuck in the loop forever, if for whatever reason > the tls.tlogReady never becomes true.

While your statement is true, @FastGeert doesn't want to put any timeout in this readyness thing.
On the other hand, does the gonbdserver doesn't catch when Flush is too slow?

But yeah, it seems that this part might need timeout

nbd/nbdserver/tlog/tlog.go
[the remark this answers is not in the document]
iwanbk: > Yeah, indeed. I couldn't find better approach back then.
> Let me know if you know better way to do this

Found better way, without needing additional mutex, will push it soon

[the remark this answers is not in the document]
iwanbk: pushed

nbd/nbdserver/tlog/tlog.go
```diff
@@ -49,11 +49,13 @@ func Storage(ctx context.Context, vdiskID, clusterID string, configSource config
 	}
 
 	if client == nil {
+		log.Infof("creating tlogclient for vdisk `%v`", vdiskID)
 		client, err = tlogclient.New(tlogClusterConfig.Servers, vdiskID)
 		if err != nil {
 			cancel()
 			return nil, errors.New("tlogStorage requires valid tlogclient: " + err.Error())
 		}
+		log.Infof("tlogclient for vdisk `%v` created", vdiskID)
```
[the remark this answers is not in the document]
iwanbk: fixed

nbd/nbdserver/tlog/tlog.go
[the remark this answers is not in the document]
iwanbk: done, added timeout

nbd/nbdserver/tlog/wait_tlog_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/zero-os/0-Disk/config"
+	"github.com/zero-os/0-Disk/nbd/ardb/storage"
+	"github.com/zero-os/0-Disk/tlog"
+	"github.com/zero-os/0-Disk/tlog/stor"
+	"github.com/zero-os/0-Disk/tlog/tlogserver/server"
+)
+
+// Test that include waiting for other tlog to be finished
+// see : https://github.com/zero-os/0-Disk/issues/527
+// it tests the idea described at  https://github.com/zero-os/0-Disk/issues/527#issuecomment-334711480
```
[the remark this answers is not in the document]
iwanbk: done

tlog/status.go
```diff
@@ -27,6 +27,8 @@ func (status BlockStatus) String() string {
 		return "RecvFailed"
 	case BlockStatusForceFlushReceived:
 		return "ForceFlushCommandReceived"
+	case BlockStatusReady:
+		return "Ready to receive block"
```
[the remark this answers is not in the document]
iwanbk: done

tlog/stor/client.go
```diff
@@ -190,11 +190,11 @@ func (c *Client) processStoreData(data []byte, lastSequence uint64, timestamp in
 
 	c.storeNum++
 	// we don't store last sequence on each iteration because
-	// we can easily fetch it using `Walk` method while it certainly increase
+	// we can easily fetch it using `Walk` method, while it certainly increase
 	// processing time.
 	// We still store it because otherwise we need to walk the entire disk
 	// on startup.
-	if c.storeNum%100 == 0 {
+	if c.storeNum%25 == 0 {
```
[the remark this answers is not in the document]
iwanbk: changed to constant.
But not to external docs, because i don't see any reason why we need to do this for something private

iwanbk:
@GlenDC i've addressed  all your remarks

iwanbk:
@GlenDC let me know if you need something else to be fixed or improved

GlenDC:
Merged master for you into this branch, and changed the code where needed because of this merge. Will now give the PR a review it deserves. Thanks for waiting so long.

nbd/nbdserver/tlog/tlog.go
```diff
@@ -303,10 +344,43 @@ func (tls *tlogStorage) Close() (err error) {
...
+	defer tls.mux.Unlock()
+	tls.tlogReady = true
+
+	// update last sequence
+	tls.sequenceMux.Lock()
+	tls.sequence = tls.tlog.LastFlushedSequence() + 1
+	tls.sequenceMux.Unlock()
+
+	// resume all helds IO
+	for _, op := range tls.tlogNotReadyBuff {
```
GlenDC: Instead of these 2 loops I would do the following instead:

```go
for _, operation := range tls.tlogNotReadyBuff {
    // scheduele tlog transaction, to be sent to the server
    tls.transactionCh <- &transaction{
        Operation: op.op,
        Sequence:  op.sequence,
        Index:     op.blockIndex,
        Timestamp: op.timestamp,
        Content:   op.content,
    }
}
```

Seems a bit silly to evict something from the cache, to than add it again immediately...

nbd/nbdserver/tlog/tlog.go
```diff
@@ -119,6 +125,12 @@ type transaction struct {
 	Timestamp int64
 }
 
+type writeOp struct {
+	sequence   uint64
+	blockIndex int64
+	content    []byte
+}
```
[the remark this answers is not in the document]
GlenDC: But thinking about it, if you do that you can just as well remove this struct and cache directly the transaction struct.

tlog/stor/client.go
```diff
@@ -17,9 +17,14 @@ import (
 
 const (
 	capnpBufLen = 4096 * 4
+
+	// description of this constant can be found in the place where this constant
```
GlenDC: please summarize a description here instead, otherwise writing a comment is kinda useless

GlenDC: Or if you want declare it is as a local constant within the function body where you use it instead. I never said you had to declare it at a package level

tlog/tlogclient/client.go
```diff
@@ -239,6 +243,20 @@ func (c *Client) runReceiver(ctx context.Context, cancelFunc context.CancelFunc)
 					if len(tr.Sequences) > 0 { // should always be true, but we anticipate.
 						c.blockBuffer.SetSent(tr.Sequences[0])
 					}
+				case tlog.BlockStatusReady:
+					seq := tr.Sequences[0]
+					if seq != 0 {
+						c.blockBuffer.SetLastFlushed(seq)
+					}
+					if c.Ready() {
+						continue
```
GlenDC: is this even possible, how can we receive a `tlog.BlockStatusReady` sequence if we're already ready?

tlog/tlogclient/client.go
```diff
@@ -356,16 +374,22 @@ func (c *Client) connect() (err error) {
 
 	c.setCurServerFailedFlushStatus(false)
 
-	if err = c.createConn(); err != nil {
+	if err := c.createConn(); err != nil {
```
GlenDC: why?

tlog/tlogclient/client.go
```diff
@@ -356,16 +374,22 @@ func (c *Client) connect() (err error) {
...
+	ready, err := c.handshake()
+	if err != nil {
 		if errClose := c.conn.Close(); errClose != nil {
 			log.Debug("couldn't close open connection of invalid client:", errClose)
 		}
 		return fmt.Errorf("client handshake failed: %s", err.Error())
 	}
+
+	c.mux.Lock()
+	defer c.mux.Unlock()
```
GlenDC: this defer is a bit much, for simple things like this, where you can be sure it will not crash I would just do

```go
c.mux.Lock()
c.serverReady = ready
c.mux.Unlock()
```

Remember that defer can easily add about 100ms to the call. Not that you cannot use defer, as it is useful in plenty of places, but here it seems a bit overkill.

tlog/tlogclient/client.go
```diff
@@ -399,36 +423,41 @@ func (c *Client) resender() {
 	}
 }
 
-func (c *Client) handshake() error {
+// do handshaking process to tlogserver
+// it returns true if tlogserver is ready
+func (c *Client) handshake() (bool, error) {
```
GlenDC: do `(ready bool, err error)` instead, this makes it also clear from the handshake signature that the bool is the ready one, and this way you don't have to declare it on line 429.

tlog/tlogclient/client.go
```diff
@@ -399,36 +423,41 @@ func (c *Client) resender() {
...
+	ready := false
```
GlenDC: also.. do `var ready bool` when you want to use the nil value of a type, not sure if it's cheaper, but it seems a bit weird to explicitly assign it the nil value yourself.

tlog/schema/tlog_schema.capnp
```diff
@@ -15,6 +15,7 @@ struct HandshakeResponse {
 	version @0 :UInt32;
 	status @1 :Int8;
 	lastFlushedSequence @2: UInt64;
+	waitTlogReady @3: Bool;
```
GlenDC: I would rename this to `ready`, and thus assign the negated version of what you are assigning now. As it basically responds with the state of the server, so making this last property a commanding thing (WAIT!), seems like a bit weird. It aligns much better with the other properties of this handshake if it would just indicate whether or not the (tlog) server is ready.

tlog/tlogclient/client.go
```diff
@@ -562,6 +591,24 @@ func (c *Client) Close() error {
 	return nil
 }
 
+// Ready returns true if server is ready and thus this client
+// is ready to be used
+func (c *Client) Ready() bool {
+	c.mux.Lock()
+	defer c.mux.Unlock()
```
GlenDC: I wouldn't do a defer here. Instead do:

```go
c.mux.Lock()
ready := c.serverReady
c.mux.Unlock()
return ready
```

And even better would be if we can avoid this mutex all together, but I guess that's wishful thinking for now.

tlog/tlogserver/server/config.go
```diff
@@ -5,6 +5,12 @@ import (
 	"github.com/zero-os/0-Disk/tlog/tlogserver/aggmq"
 )
 
+const (
+	// WaitListenAddrRandom is wait listen addr which wait in random port.
+	// Only useful for test
+	WaitListenAddrRandom = "-"
```
GlenDC: I like that you add a constant char for this purpose, but I would use the `*` character, as it is convention to use this char for any wildcard purposes, and this seems defintely to be such a wildcard purpose.

tlog/tlogserver/server/server.go
```diff
@@ -74,6 +95,46 @@ func (s *Server) Listen(ctx context.Context) {
 	s.ctx = ctx
 	defer s.listener.Close()
 
+	if s.coordListener != nil {
+		log.Infof("s.coordListener listen at: %v", s.coordListener.Addr().String())
+		go func() {
```
GlenDC: I would defintely turn this into a seperate function and spawn that as a goroutine instead, because not it just makes the Listen function bigger and more complex for no good reason.

tlog/tlogserver/server/server.go
```diff
@@ -74,6 +95,46 @@ func (s *Server) Listen(ctx context.Context) {
...
+				// start listener for the coordination (wait-other-tlog) purposes
+				conn, err := s.coordListener.Accept()
+				if err != nil {
+					log.Errorf("couldn't accept coord connection:%v", err)
+					continue
+				}
+
+				log.Infof("accepted coordination connection from %s", conn.RemoteAddr().String())
+
+				go func(conn net.Conn) {
```
GlenDC: I don't think it is needed to add another local variable for `conn` here, fairly sure you can just use it directly from the created on in line 103.

tlog/tlogserver/server/server.go
```diff
@@ -74,6 +95,46 @@ func (s *Server) Listen(ctx context.Context) {
...
+		log.Infof("s.coordListener listen at: %v", s.coordListener.Addr().String())
+		go func() {
+			for {
+				// start listener for the coordination (wait-other-tlog) purposes
+				conn, err := s.coordListener.Accept()
+				if err != nil {
+					log.Errorf("couldn't accept coord connection:%v", err)
+					continue
+				}
+
```
GlenDC: I miss here some kind of handshake procedure. Something that would validate the connection, server versions, and anything else you may wish to validate. On top of that the handshake should also take into account future extensions. Because right now the connection can basically be anything, not just what you might expect.

tlog/tlogserver/server/server.go
```diff
@@ -74,6 +95,46 @@ func (s *Server) Listen(ctx context.Context) {
...
+					// start handler for tlog waiter
+					defer func() {
+						conn.Close()
+					}()
+					for {
+						select {
+						case <-ctx.Done():
+							return
+						default:
+							b := make([]byte, 1)
```
GlenDC: I would send something more meaningful than some random character that you don't even inspect.

tlog/tlogserver/server/server.go
```diff
@@ -74,6 +95,46 @@ func (s *Server) Listen(ctx context.Context) {
...
+							conn.SetReadDeadline(time.Now().Add(1 * time.Second))
```
GlenDC: I don't think 1 second is a reasonable deadline. Remember that network connections might not always be fast or reliable.

tlog/tlogserver/server/server.go
```diff
@@ -83,6 +144,23 @@ func (s *Server) Listen(ctx context.Context) {
...
+			log.Infof("connection request from %s", remoteAddr)
+			host, _, err := net.SplitHostPort(remoteAddr)
+			if err != nil {
+				log.Errorf("fail to parse remote address: %v", err)
+				conn.Close()
+				continue
+			}
+
+			if s.acceptAddr != "" && s.acceptAddr != host {
+				log.Infof("connection from %s refused, it does not match the accept address configured %s", remoteAddr, s.acceptAddr)
```
GlenDC: should be a `log.Errorf` seemingly.

tlog/tlogserver/server/server.go
```diff
@@ -83,6 +144,23 @@ func (s *Server) Listen(ctx context.Context) {
...
+			remoteAddr := conn.RemoteAddr().String()
+			log.Infof("connection request from %s", remoteAddr)
+			host, _, err := net.SplitHostPort(remoteAddr)
+			if err != nil {
+				log.Errorf("fail to parse remote address: %v", err)
+				conn.Close()
+				continue
+			}
+
+			if s.acceptAddr != "" && s.acceptAddr != host {
```
GlenDC: I think your simple equality check will not always work. Think about DNS, IPv4 vs IPv6 etc... Fairly sure you can't simply assume that the RemoteAddr String result will equal what you might have defined as acceptAddr.

tlog/tlogserver/server/server.go
```diff
@@ -83,6 +144,23 @@ func (s *Server) Listen(ctx context.Context) {
...
+				log.Infof("connection from %s refused, it does not match the accept address configured %s", remoteAddr, s.acceptAddr)
+				conn.Close()
+				continue
+			}
+			log.Infof("connection accepted from %s", remoteAddr)
+
 			tcpConn, ok := conn.(*net.TCPConn)
 			if !ok {
 				log.Info("received conn is not tcp conn")
```
GlenDC: old code I know, but it seems like a `conn.Close()` statement is missing here.

GlenDC: Also I think you can remove the `addr := conn.RemoteAddr()` a bit below this, and replace its uses (`addr.String()`) with the `remoteAddr` you have declared earlier.

tlog/tlogserver/server/vdisk.go
```diff
@@ -127,8 +133,13 @@ func newVdisk(parentCtx context.Context, vdiskID string, aggMq *aggmq.MQ, config
 		aggMq:           aggMq,
 		apc:             apc,
 
-		ctx:        ctx,
-		cancelFunc: cancelFunc,
+		ctx:              ctx,
+		cancelFunc:       cancelFunc,
+		coordConnectAddr: coordConnectAddr,
```
GlenDC: why not do `ready: coordConnectAddr == "",`, instead of some branching down below.

tlog/tlogserver/server/vdisk.go
```diff
@@ -153,6 +164,12 @@ func newVdisk(parentCtx context.Context, vdiskID string, aggMq *aggmq.MQ, config
 	return vd, nil
 }
 
+func (vd *vdisk) Ready() bool {
+	vd.mux.Lock()
```
GlenDC: Where you can avoid the runtime cost that a `defer` calls brings with, I would avoid it.

For example here you can remove the defer statement by doing the following instead:
```go
vd.mux.Lock()
ready := vd.ready
vd.mux.Unlock()
return ready
```

tlog/tlogserver/server/vdisk_conn.go
```diff
@@ -27,6 +27,23 @@ func (vd *vdisk) handle(conn *net.TCPConn, br *bufio.Reader, respSegmentBufLen i
 	// start response sender
 	go vd.sendResp(ctx, conn, respSegmentBufLen)
 
+	if !vd.Ready() && vd.coordConnectAddr != "" {
```
GlenDC: What if the vdisk is not ready but `vd.coordConnectAddr == ""`, isn't that a situation unhandled here?

tlog/tlogserver/server/vdisk_wait.go
```diff
@@ -0,0 +1,42 @@
...
+
+	if err := vd.createFlusher(); err != nil {
+		return 0, err
+	}
+	return vd.loadLastFlushedSequence()
+}
+
+// wait for other server to be finished
+func (vd *vdisk) waitOtherToFinish() {
+	defer func() {
```
GlenDC: no need to wrap it in another function though, you can just defer the log call as it is.

tlog/tlogserver/server/vdisk_wait.go
```diff
@@ -0,0 +1,42 @@
...
+		log.Infof("waitOther %v finished", vd.coordConnectAddr)
+	}()
+
+	// connect to the other server
+	conn, err := net.Dial("tcp", vd.coordConnectAddr)
+	if err != nil {
+		log.Errorf("failed to dial %v: %v", vd.coordConnectAddr, err)
+		return
+	}
+
```
GlenDC: I miss here some kind of handshake protocol, similar to my other comment.

tlog/tlogserver/server/vdisk_wait.go
```diff
@@ -0,0 +1,42 @@
...
+	log.Infof("vdisk %v waiting for remote server %v to dies", vd.id, vd.coordConnectAddr)
+
+	// wait for other server to dies
+	b := make([]byte, 1)
```
GlenDC: I would read something more meaningful than some random byte, and also check its content.

iwanbk:
>  your S2 remains unusable for that vdisk, or am I wrong?

True, we can't risk of having two tlog for same vdisk

nbd/nbdserver/tlog/tlog.go
iwanbk: we can't do this, the sequence number is not valid anymore at this moment

tlog/tlogclient/client.go
iwanbk: Yes, should be not possible, just additional checking, in case the server's response sender sent double response

tlog/tlogclient/client.go
iwanbk: you're right

tlog/tlogserver/server/server.go
iwanbk: yep,

tlog/tlogserver/server/server.go
iwanbk: the timeout is only for the code to still be able to catch `ctx.Done`

tlog/tlogclient/client.go
iwanbk: naked return is debatable, "go tour" also suggest it to only use it on short func

tlog/tlogclient/client.go
GlenDC: Sure, but shouldn't that be checked earlier than. Like as the first thing you do for this `case`?

tlog/tlogclient/client.go
GlenDC: It's not because you use named parameters that you have to do a naked return.

You can still do `return ready, err` or even, `return false, err`.

iwanbk: > It's not because you use named parameters that you have to do a naked return.
yep.

My reply was because i only saw you use named params along with naked return, until yesterday :)

Good point btw, will change it

tlog/tlogclient/client.go
iwanbk: you're right

tlog/tlogserver/server/config.go
iwanbk: yep

tlog/tlogserver/server/server.go
iwanbk: i'm adding simple handshaking now

tlog/tlogserver/server/server.go
iwanbk: it is good practice for code in loop, i saw super slow goroutine creation which could be easily fixed by this code style

tlog/tlogserver/server/server.go
iwanbk: It doesn't mean to send any data, it only mean to keep the connection

tlog/tlogserver/server/server.go
iwanbk: will be addressed at #539

tlog/tlogserver/server/server.go
iwanbk: fixed

tlog/tlogserver/server/server.go
iwanbk: fixed

tlog/tlogserver/server/vdisk.go
iwanbk: fixed

tlog/tlogserver/server/vdisk_conn.go
iwanbk: second condition is not needed, removed

tlog/tlogserver/server/vdisk_wait.go
iwanbk: yep

tlog/tlogserver/server/vdisk_wait.go
iwanbk: added

tlog/tlogserver/server/vdisk_wait.go
iwanbk: it is not reading, it is waiting to be closed. But i have added handshake now

tlog/tlogserver/server/server.go
```diff
@@ -318,3 +281,83 @@ func (s *Server) ReadDecodeClientMessage(r io.Reader) (*schema.TlogClientMessage
...
+		// start listener for the coordination (wait-other-tlog) purposes
+		conn, err := s.coordListener.Accept()
+		if err != nil {
+			log.Errorf("couldn't accept coord connection:%v", err)
+			continue
+		}
+
+		log.Infof("accepted coordination connection from %s", conn.RemoteAddr().String())
+
+		go s.handleTlogWait(conn)
```
GlenDC: This seems to make it possible for multiple tlogServers to connect to this tlogServer for the same vdisk? Or am I wrong? And if so, would that ever be a problem? Not that it should ever happen, but it's hard to know what sometimes happens in production (now or in the future).

tlog/tlogserver/server/vdisk.go
```diff
@@ -152,6 +160,12 @@ func newVdisk(parentCtx context.Context, vdiskID string, aggMq *aggmq.MQ, config
...
+	defer vd.mux.Unlock()
```
GlenDC: You still have the defer here.

nbd/nbdserver/tlog/tlog.go
```diff
@@ -171,11 +171,11 @@ func (tls *tlogStorage) set(blockIndex int64, content []byte) error {
 			blockIndex, sequence)
 	}
 
-	if tls.tlogReady {
-		// copy the content to avoid race condition with value in cache
-		transactionContent := make([]byte, len(content))
-		copy(transactionContent, content)
+	// copy the content to avoid race condition with value in cache
+	transactionContent := make([]byte, len(content))
+	copy(transactionContent, content)
```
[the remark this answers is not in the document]
iwanbk: fixed

tlog/tlogserver/server/server.go
iwanbk: >This seems to make it possible for multiple tlogServers to connect to this tlogServer for the same > vdisk? Or am I wrong?

Yeah

> And if so, would that ever be a problem?

No

tlog/tlogserver/server/vdisk.go
iwanbk: Changed

tlog/tlogserver/server/vdisk_mgr.go
```diff
@@ -51,6 +51,14 @@ func (vt *vdiskManager) Get(ctx context.Context, vdiskID string,
 	return
 }
 
+// exists returns true if vdisk with given ID is exists
+func (vt *vdiskManager) exists(vdiskID string) bool {
+	vt.lock.Lock()
+	defer vt.lock.Unlock()
```
GlenDC: got a `defer` call here that you can easily avoid.

Issue #543, opened 2017-10-12
add testable (godoc) examples for ardb package


GlenDC:
add missing reply examples to `reply_test.go`: https://github.com/zero-os/0-Disk/blob/df57a5623e5442559a30e112a5507413c2e58d09/nbd/ardb/reply_test.go#L295-L298

- [x] `ExampleError`
- [x] `ExampleBytes`
- [x] `ExampleOptBytes`
- [x] `ExampleInt64`
- [x] `ExampleInt64ToBytesMapping`

add missing cluster examples to `cluster_test.go`: https://github.com/zero-os/0-Disk/blob/df57a5623e5442559a30e112a5507413c2e58d09/nbd/ardb/cluster_test.go#L13-L16

- [x] `ExampleNewCluster`
- [ ] <s>`ExampleNewClusterPair`  </s>

chrisvdg:
@GlenDC , what does `ExampleNewClusterPair` have to show specifically?

GlenDC:
> what does ExampleNewClusterPair have to show specifically?

Nevermind that one, it doesn't exist anymore (for the time being). It might exist again once we figured out the entire primary-slave stuff, but for now it's non-existent.

PR #548, opened 2017-10-13
WIP Adds timeout for the tlog client reconnecting to servers

### LINKED ISSUE #541: tlogclient should limit the amount of times it tries to reconnect to
Right now the tlogclient can get stuck in an infinite loop if during a reconnect no tlogserver is available, and thus it will keep shifting and trying without success.

You can see it clearly for yourself how the infinite loop can happen (and has happened for some users): https://github.com/zero-os/0-Disk/blob/a60cf2b832cfa94ceafd4f0d2fec052cd47184fa/tlog/tlogclient/client.go#L323-L342

+ Even though the return parameter is `error`, and even though that result is handled by the callee, the function will in fact never return an error, and instead keep trying until it can return `nil`.
+ We also receive an input parameter which indicates the point in time at which we started retrying (as the callee uses `time.Now()`), this input parameter is not used;

There needs to be put some limit here:

+ Either we limit based on some maximum time of trying to reconnect;
+ Or we limit based on how many times we can retry each server or the total amount of retries;

But right now we simply keep trying, which is pointless, as in that scenario all hope is basically lost for the vdisk and we shouldn't block the entire system because of it, and instead exit with an error.

chrisvdg:
closes #541 

+ When a tlog server connection fails, the server is marked as failed;
+ The fail status for a specific server will be ignored when enough time has passed since the last time it failed;
+ The fail status or all servers will be erased when the config gets reloaded;
+ If no servers are available, the client will be come unusable and all methods will return an error when used;
  + However, before going so radical, it will try to reconnect, if it can find a new server (taking the new rules from above into account), to reconnect to, this server will be used and the client becomes usable once again;

### DIFF
--- a/tlog/tlogclient/client.go
+++ b/tlog/tlogclient/client.go
@@ -133,7 +133,7 @@ func (c *Client) run(ctx context.Context) {
 			case <-ctx.Done():
 				return
 			default:
-				err := c.reconnect(time.Now())
+				err := c.reconnect(5 * time.Second)
 				if err != nil {
 					log.Errorf("reconnect failed : %v", err)
 				} else {
@@ -321,23 +321,32 @@ func (c *Client) getCurServerFailedFlushStatus() bool {
 }
 
 // reconnect to server
-func (c *Client) reconnect(closedTime time.Time) error {
+// if timeout is 0 it will take the default timeout
+func (c *Client) reconnect(timeout time.Duration) error {
 	var err error
+	if timeout == 0 {
+		timeout = 1 * time.Minute
+	}
+	timeoutTick := time.After(timeout)
 
 	log.Info("tlogclient reconnect")
 
 	for {
-		// try other server
-		c.shiftServer()
+		select {
+		case <-timeoutTick:
+			return fmt.Errorf("timed out reconnecting to tlog servers")
+		default:
+			// try other server
+			c.shiftServer()
 
-		if err = c.connect(); err == nil {
-			// if reconnect success, sent all unflushed blocks
-			// because we don't know what happens in servers.
-			// it might crashed
-			c.blockBuffer.SetResendAll()
-			return nil
+			if err = c.connect(); err == nil {
+				// if reconnect success, sent all unflushed blocks
+				// because we don't know what happens in servers.
+				// it might crashed
+				c.blockBuffer.SetResendAll()
+				return nil
+			}
 		}
-
 	}
 }
 
--- a/tlog/tlogclient/client.go
+++ b/tlog/tlogclient/client.go
@@ -36,7 +36,7 @@ var (
 	// ErrFlushFailed returned when client failed to do flush
 	ErrFlushFailed = errors.New("tlogserver failed to flush")
 
-	// ErrServersExhausted respresnets an error where all servers have been exhausted
+	// ErrServersExhausted represents an error where all servers have been exhausted
 	// trying to (re)connect to them
 	ErrServersExhausted = errors.New("all servers have been exhausted")
 )
@@ -132,7 +132,8 @@ func newClient(serverAddresses []string, vdiskID string) (*Client, error) {
 		respCh: make(chan *Result, 3),
 
 		// channel for retry signal
-		retryCh: make(chan struct{}),
+		retryCh:    make(chan struct{}),
+		retryErrCh: make(chan error),
 	}
 
 	return client, client.connect()
@@ -153,7 +154,9 @@ func (c *Client) run(ctx context.Context) {
 	for !c.isStopped() {
 		// if failed state, wait till retry signal
 		if c.isFailedState() {
+			log.Debug("Waiting for signal before reconnecting")
 			<-c.retryCh
+			log.Debug("Received reconnecting signal")
 		}
 
 		curCtx, cancelFunc := context.WithCancel(ctx)
@@ -176,7 +179,6 @@ func (c *Client) run(ctx context.Context) {
 			}
 			c.setFailedState(true)
 		} else {
-
 			if c.isFailedState() {
 				c.retryErrCh <- nil
 			}
@@ -303,17 +305,13 @@ func (c *Client) runReceiver(ctx context.Context, cancelFunc context.CancelFunc)
 }
 
 // ChangeServerAddresses change servers to the given servers
-func (c *Client) ChangeServerAddresses(servers []string) error {
-	err := c.retryIfFailedState()
-	if err != nil {
-		return err
-	}
+func (c *Client) ChangeServerAddresses(servers []string) {
 
 	c.mux.Lock()
 	defer c.mux.Unlock()
 
 	if len(servers) == 0 {
-		return nil
+		return
 	}
 	log.Infof("tlogclient vdisk '%v' change server addrs to '%v'", c.vdiskID, servers)
 
@@ -333,7 +331,6 @@ func (c *Client) ChangeServerAddresses(servers []string) error {
 		log.Infof("tlogclient vdisk '%v' current server is not in new address, close it", c.vdiskID)
 		c.conn.Close()
 	}
-	return nil
 }
 
 // shift server address move current active client to the back
@@ -370,19 +367,22 @@ func (c *Client) getCurServerFailedFlushStatus() bool {
 func (c *Client) reconnect() error {
 	var err error
 
-	log.Info("tlogclient reconnect")
+	log.Info("tlogclient reconnecting to tlog server...")
 
 	for {
+		log.Debugf("trying to reconnect to server %s...", c.servers[0].address)
+
 		if err = c.connect(); err == nil {
 			// if reconnect success, sent all unflushed blocks
 			// because we don't know what happens in servers.
 			// it might crashed
 			c.blockBuffer.SetResendAll()
+			log.Debugf("succeeded to reconnect to %v", c.servers[0].address)
 			return nil
 		}
 
 		// failed to reconnect
-		log.Debugf("failed to reconnect to %v", c.servers[0].address)
+		log.Debugf("failed to reconnect to %v: %s", c.servers[0].address, err)
 		c.servers[0].failed = true
 		c.servers[0].lastTried = time.Now()
 
@@ -520,6 +520,8 @@ func (c *Client) createConn() error {
 		return err
 	}
 
+	c.mux.Lock()
+	defer c.mux.Unlock()
 	conn := genericConn.(*net.TCPConn)
 
 	conn.SetKeepAlive(true)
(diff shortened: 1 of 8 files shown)


tlog/tlogclient/client.go
```diff
@@ -321,23 +321,32 @@ func (c *Client) getCurServerFailedFlushStatus() bool {
 }
 
 // reconnect to server
-func (c *Client) reconnect(closedTime time.Time) error {
+// if timeout is 0 it will take the default timeout
+func (c *Client) reconnect(timeout time.Duration) error {
 	var err error
+	if timeout == 0 {
```
GlenDC: I don't think this logic is needed, it's an internal function, only ever called in one place, so it really shouldn't need a default value here.

tlog/tlogclient/client.go
```diff
@@ -321,12 +321,15 @@ func (c *Client) getCurServerFailedFlushStatus() bool {
...
-func (c *Client) reconnect(closedTime time.Time) error {
+// if timeout is 0 it will take the default timeout
+func (c *Client) reconnect() error {
 	var err error
 
 	log.Info("tlogclient reconnect")
 
-	for {
+	// try each server 5 times
+	attempts := len(c.servers) * 5
```
GlenDC: retrying each server 5 times is quite much o.o

also you're going to have a race condition here, as getting/setting c.servers should be protected (in the current design) by a `c.mux.Lock()`.

GlenDC:
Also please check out: https://github.com/zero-os/0-Disk/issues/541#issuecomment-336473703

tlog/tlogclient/client.go
```diff
@@ -19,6 +19,7 @@ const (
 	readTimeout          = 2 * time.Second
 	resendTimeoutDur     = 2 * time.Second // duration to wait before re-send the tlog.
 	failedFlushSleepTime = 10 * time.Second
+	retryReconnectAfter  = 10 * time.Second
```
GlenDC: this time is way too small. I would make it more something around 2 to 5 minutes.

The problem with 10 seconds is, that if you would have 10 servers in your tlogcluster, you would once again have an infinite loop.

tlog/tlogclient/client.go
```diff
@@ -48,10 +49,16 @@ type Result struct {
 	Err  error
 }
 
+type server struct {
```
GlenDC: oh I see, your name conflict is because of this vague type.

Rame this serverContext or something like that.

tlog/tlogclient/client.go
```diff
@@ -325,14 +343,23 @@ func (c *Client) getCurServerFailedFlushStatus() bool {
 func (c *Client) reconnect() error {
 	var err error
 
+	// reset fails on servers
+	for i := range c.servers {
+		c.servers[i].failed = false
```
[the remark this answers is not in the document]
chrisvdg: Removing this makes `TestReconnectFromSend` test become stuck on printing the servers  exhausted error, even after the timeout of the test forcing my to kill the process running the test. Will investigate this more tomorrow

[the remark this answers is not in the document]
GlenDC: While you're at it, also ensure all travis CI tests pass. As your latest commit seems to make them fail.

tlog/tlogclient/client.go
```diff
@@ -341,9 +368,10 @@ func (c *Client) reconnect() error {
 			return nil
 		}
 
+		// failed to reconnect
```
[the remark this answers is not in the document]
GlenDC: You still haven't moved these lines up. It should be done at the beginning of this for loop...

tlog/tlogclient/client.go
```diff
@@ -321,14 +346,22 @@ func (c *Client) getCurServerFailedFlushStatus() bool {
 }
 
 // reconnect to server
-func (c *Client) reconnect(closedTime time.Time) error {
+// if timeout is 0 it will take the default timeout
+func (c *Client) reconnect() error {
 	var err error
 
+	for i := range c.servers {
+		c.servers[i].failed = false
```
GlenDC: This is still in here? I told you that we do not want to reset manually each time we call reconnect. It should only be reset when we reload config. Otherwise it's pointless that we store the lastTried time for each server.

tlog/tlogclient/resend_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/zero-os/0-Disk/log"
 	"github.com/zero-os/0-Disk/tlog"
 	"github.com/zero-os/0-Disk/tlog/schema"
-	"github.com/zero-os/0-Disk/tlog/tlogserver/server"
+	tlogServer "github.com/zero-os/0-Disk/tlog/tlogserver/server"
```
GlenDC: You forgot to undo the changes to this file.

tlog/tlogclient/client.go
```diff
@@ -22,6 +22,10 @@ const (
 )
 
 var (
+	retryReconnectAfter = 2 * time.Minute
```
GlenDC: add some comment to this time to clarify further, also please turn this into a constant?

tlog/tlogclient/client.go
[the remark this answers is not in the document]
chrisvdg: so the logic in the forloop should be in this order?
```
-> mark failed
-> shiftServer
-> check if already failed
-> try and connect
```

tlog/tlogclient/client.go
chrisvdg: Removing this will make `TestReconnectFromSend` fail very badly, from what I get so far is that closing a conn and connecting to it again (for the next test) caused  the server to sometimes fail when reconnecting:  
`connection from 127.0.0.1:33932 dropped with an error: handshake failed: couldn't connect to vdisk 12345: there is already conencted client for vdisk '12345'` 
wich marks this server as failed and starts flooding with `errServersExhausted`messages, making the test timeout which I don't think closes the client anymore so the reconnect goroutine keeps on running until the process is killed.
Setting `retryReconnectAfter` to a really short time can fix it, but doesn't feel like a neat solution at all...

tlog/tlogclient/client.go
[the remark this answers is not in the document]
GlenDC: Yes.

tlog/tlogclient/client.go
GlenDC: OK, but you can't add stuff to production code just to make tests pass. Unless there is a good reason to have this in production code, which I do not see at all.

chrisvdg: I know, I'm looking into removing it, but can't seem to make the test pass without this

chrisvdg:
My solution for retrying to reconnect when a call is made doesn't work as it seems to deadlock occasionally.
Open for suggestions for other approaches/improvements.

GlenDC:
I'll check it out this weekend. As to not be blocked with work until than, please move your focus on some other issues for now. I assigned you #543, so you can start with that one, @chrisvdg.

iwanbk:
> closes #541
> 
> Not sure what (default) timeouts should be though.


@chrisvdg 
can you add the summary of your fixes to this PR description?

I have read #541, saw some alternative solution there, but not sure which one you choose.

GlenDC:
> can you add the summary of your fixes to this PR description?

I added the summary @iwanbk, feel free to check it out. Your review on the current state of this PR would definitely be good. I'm going to do a review myself tomorrow if needed.

The current code of this PR is good, except for the last part of the PR (adding the stuff to make the tlogclient in some kind of failed state when no servers are available). Besides that last part, it does look good.

iwanbk:
My initial review:
- instead of creating `serverContext`, i will simply do this
```go
lastTry := time.Now()
for  {
    for _, s := range c.servers {
          tryToConnectWith(s)
    }
   if lastTry < retryReconnect {
         return Err
   }
}
```
It is true that `lastTried` won't be as precise as original PR, but the `lastTried` there is only compared with `retryReconnectAfter` which value is come from nowhere

- instead of mark a server as `failed` when we can't `connect` to it, we might want to retry first. Not sure.

- how it will play with our upcoming `self healing` feature? While it is true that current code is silly: doing reconnect forever without delay and doesn't report the failed server to 0-orch, it relies on the fact/plan that we will have self healing feature in the future. In other words, it hopes that 0-orch will do it's job to recover/replace the failed tlogservers.

- If we are still want to go this way, we might also need to adapt the nbd code.
   When tlog is in failed state, `tlogclient.Send` will simply return an error, reject the transaction. What to do with the transaction? Don't we need to buffer it, in case 0-orch heal the tlogservers.
  Please note that the transaction is sent in async way.

My suggestion is to fix the silly things in current code
- report to 0-orch when a server failed
- add some delay before try to reconnect

iwanbk:
> instead of creating serverContext, i will simply do this

It couldn't be simply like that, that simple code won't play well when the servers address got updated

iwanbk:
> If we are still want to go this way, we might also need to adapt the nbd code.

I only mentioned about `tlogclient.Send` there, doesn't mean that all other APIs are safe, all other APIs need to be checked as well

GlenDC:
It is true that we have self-healing in place. But that doesn't mean we can block forever.

What we're handling here is a worst-case scenario where _all_ tlogservers configured for a vdisk are offline. Obviously we need to warn AYS about this, but if we can't connect to any of these servers there is not much we can do, except to simply exit the vdisk. We will lose some data, but at least the logged transactions would still be in sync with the ARDB. This is a worst case scenario, and there is not much we can do in this case. Even if you wait for a long time, _all_ servers might simply be dead forever.

Errors should be handled anyhow as failures. AFAIK the tlogClient already supports errors as return values, and thus that should already be handled as such. If not, that would simply have meant that we were incorrectly using the existing API in the first place.

iwanbk:
>  But that doesn't mean we can block forever.

Not forever, the API user will see that it's flush failed

> What we're handling here is a worst-case scenario where all tlogservers configured for a vdisk are offline. Obviously we need to warn AYS about this, but if we can't connect to any of these servers there is not much we can do, except to simply exit the vdisk.

As far as i can see, the solution design is not complete.
- there were no plan to notify orchestrator
- we should retry to connect before mark it as failed
- we can't simply exit the vdisk, because nbd's vdisk is currently can't be exited (see : #521)
- what nbd need to do if sequence X failed then X+1 succeed? I believe there are things to do related with sequenceCache

I strongly suggest for this PR to only tackle first two points, and let the rest points to be handled in separate issue. Because the rest points are not suitable for `mentor` type.

To summarize, current code is absolutely need to fixed, but fixing it using the plan mentioned in this PR will only make it worse.

GlenDC:
Closing this PR, as it's not a solution we desire for the current master branch.

On top of that it won't be needed anymore once we switch over to the standalone tlog repo. We'll keep the branch alive for now, for reference reasons, will delete it in the future once we really know for sure we won't need it anymore.

PR #554, opened 2017-10-18
move data/parity shard count to (yaml) config

### LINKED ISSUE #268: Add K and M values as optional values to storageCluster yaml config
Currently the `k`+`m` values are global. However it would be great if those values could also be customized per vdisk, which would result in a (`tlogserver`) config such as:

```yaml
storageClusters:
  tlogcluster:
    dataStorage: # k+m server configs
    - address: 192.168.122.76:2000
    - address: 192.168.122.76:2001
    - address: 192.168.122.76:2002
    - address: 192.168.122.76:2003
    tlog:
      k: 2
      m: 2
vdisks:
  myvdisk:
    tlogStorageCluster: tlogcluster
    k: 2 # optional, if not given the global k value is used
    m: 2 # optional, if not given the global m value is used
```

This feature would obviously only work, if we use the config-based redis pool factory (which is selected only when the `-memorystorage` and `-storage-addresses` flags are NOT specified).

Also I think it would make sense, if we make sure that we validate that if either `k` or `m` is given for a vdisk, that the other value of the pair `k+m` is also given, as it would be super confusing to allow one to be local and the other to be global.

GlenDC:
fixes #268

### DIFF
--- a/config/composition_api_test.go
+++ b/config/composition_api_test.go
@@ -168,8 +168,13 @@ func TestReadTlogStorageConfig(t *testing.T) {
 			ClientID:  "foo clientID",
 			Secret:    "foo secret",
 		},
-		Servers:         []ServerConfig{ServerConfig{"1.1.1.1:11"}},
 		MetadataServers: []ServerConfig{ServerConfig{"2.2.2.2:22"}},
+		DataServers: []ServerConfig{
+			ServerConfig{"1.1.1.1:11"},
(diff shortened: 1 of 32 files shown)


GlenDC:
@iwanbk did you manage to review this quickly?

iwanbk:
> @iwanbk did you manage to review this quickly?

missed to do it tomorrow, need to catch up on many things

PR #555, opened 2017-10-19
streamline storage config API

### LINKED ISSUE #542: streamline config/storage API in terms of pointer usage
We are not consistent in how we create, pass and use `config.StorageServerConfig` and `config.StorageClusterConfig`. Sometimes they are created and passed on by value, and other times as pointers. Even worse, often they go (multiple times) from pointer to value and perhaps even back to pointer.

+ The `Validate() error` method for those types should return an error in case the pointer version of that type is `nil`, this will help a lot;

After that I'm not too sure if we should either use them by value, or by pointer. I think using them by value and passing them by value is probably a lot safer, and in general it's also not too expensive.

+ `config.StorageClusterConfig` just contains a slice;
+ `config.StorageServerConfig` contains a string, int64 and uint8, so still manageable, especially as it's not really passed frequently in any performance critical location;

For now I leave this on the beta-1 backlog, but this can and should probably be done right after #483 is resolved and merged into master.

chrisvdg:
Steamlines passing `config.StorageServerConfig` and `config.StorageClusterConfig` in the config API

closes #542

### DIFF
--- a/config/composition_api.go
+++ b/config/composition_api.go
@@ -36,30 +36,28 @@ func ReadNBDStorageConfig(source Source, vdiskID string) (*NBDStorageConfig, err
 		log.Debugf("ReadNBDStorageConfig failed, invalid StorageClusterConfig: %v", err)
 		return nil, err
 	}
-	nbdStorageConfig.StorageCluster = storageClusterConfig
+	nbdStorageConfig.StorageCluster = *storageClusterConfig
 
 	// if template storage ID is given, fetch it
 	if nbdConfig.TemplateStorageClusterID != "" {
 		// Fetch Template Storage Cluster Config
-		templateStorageCluster, err := ReadStorageClusterConfig(
+		nbdStorageConfig.TemplateStorageCluster, err = ReadStorageClusterConfig(
 			source, nbdConfig.TemplateStorageClusterID)
 		if err != nil {
 			log.Debugf("ReadNBDStorageConfig failed, invalid TemplateStorageClusterConfig: %v", err)
 			return nil, err
 		}
-		nbdStorageConfig.TemplateStorageCluster = &templateStorageCluster
 	}
 
 	// if slave storage ID is given, fetch it
 	if nbdConfig.SlaveStorageClusterID != "" {
 		// Fetch Template Storage Cluster Config
-		slaveStorageCluster, err := ReadStorageClusterConfig(
+		nbdStorageConfig.SlaveStorageCluster, err = ReadStorageClusterConfig(
 			source, nbdConfig.SlaveStorageClusterID)
 		if err != nil {
 			log.Debugf("ReadNBDStorageConfig failed, invalid SlaveStorageClusterConfig: %v", err)
 			return nil, err
 		}
-		nbdStorageConfig.SlaveStorageCluster = &slaveStorageCluster
 	}
 
 	return nbdStorageConfig, nil
@@ -96,13 +94,12 @@ func ReadTlogStorageConfig(source Source, vdiskID string) (*TlogStorageConfig, e
 
 	// if slave storage ID is given, fetch it
 	if tlogConfig.SlaveStorageClusterID != "" {
-		slaveStorageCluster, err := ReadStorageClusterConfig(
+		tlogStorageConfig.SlaveStorageCluster, err = ReadStorageClusterConfig(
 			source, tlogConfig.SlaveStorageClusterID)
 		if err != nil {
 			log.Debugf("ReadTlogStorageConfig failed, invalid SlaveStorageCluster: %v", err)
 			return nil, err
 		}
-		tlogStorageConfig.SlaveStorageCluster = &slaveStorageCluster
 	}
 
 	return tlogStorageConfig, nil
--- a/config/composition_api.go
+++ b/config/composition_api.go
@@ -36,28 +36,30 @@ func ReadNBDStorageConfig(source Source, vdiskID string) (*NBDStorageConfig, err
 		log.Debugf("ReadNBDStorageConfig failed, invalid StorageClusterConfig: %v", err)
 		return nil, err
 	}
-	nbdStorageConfig.StorageCluster = *storageClusterConfig
+	nbdStorageConfig.StorageCluster = storageClusterConfig
 
 	// if template storage ID is given, fetch it
 	if nbdConfig.TemplateStorageClusterID != "" {
 		// Fetch Template Storage Cluster Config
-		nbdStorageConfig.TemplateStorageCluster, err = ReadStorageClusterConfig(
+		templateStorageCluster, err := ReadStorageClusterConfig(
 			source, nbdConfig.TemplateStorageClusterID)
 		if err != nil {
 			log.Debugf("ReadNBDStorageConfig failed, invalid TemplateStorageClusterConfig: %v", err)
 			return nil, err
 		}
+		nbdStorageConfig.TemplateStorageCluster = &templateStorageCluster
 	}
 
 	// if slave storage ID is given, fetch it
 	if nbdConfig.SlaveStorageClusterID != "" {
 		// Fetch Template Storage Cluster Config
-		nbdStorageConfig.SlaveStorageCluster, err = ReadStorageClusterConfig(
+		slaveStorageCluster, err := ReadStorageClusterConfig(
 			source, nbdConfig.SlaveStorageClusterID)
 		if err != nil {
 			log.Debugf("ReadNBDStorageConfig failed, invalid SlaveStorageClusterConfig: %v", err)
 			return nil, err
 		}
+		nbdStorageConfig.SlaveStorageCluster = &slaveStorageCluster
 	}
 
 	return nbdStorageConfig, nil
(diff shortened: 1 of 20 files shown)


config/config_api.go
```diff
@@ -76,16 +76,16 @@ func ReadVdiskTlogConfig(source Source, vdiskID string) (*VdiskTlogConfig, error
 
 // ReadStorageClusterConfig returns the requested StorageClusterConfig
 // from a given config source.
-func ReadStorageClusterConfig(source Source, clusterID string) (*StorageClusterConfig, error) {
+func ReadStorageClusterConfig(source Source, clusterID string) (StorageClusterConfig, error) {
```
GlenDC: We've discussed before that using pointer-type values is fine for situations where the values are optional. When you have a function with the return parameters `(T, error)`, than `T` is in fact optional. Conventionally it can be assumed that `T` is (only) defined when `error` is `nil`.

Thus please revert this (and other similar code).

config/format.go
```diff
@@ -449,7 +449,9 @@ type StorageServerConfig struct {
 // Validate this Storage Server Config,
 // returning an error in case this config is invalid.
 func (cfg *StorageServerConfig) Validate() error {
-	if cfg == nil || cfg.State == StorageServerStateRIP {
+	if cfg == nil {
+		return ErrInvalidStorageServerConfig
+	} else if cfg.State == StorageServerStateRIP {
```
GlenDC: just turn this into another `if`, as in:

```go
if cfg == nil {
    // ...
}
if cfg.State == StorageServerStateRIP {
   // ....
}
```

config/format.go
```diff
@@ -449,7 +449,9 @@ type StorageServerConfig struct {
 // Validate this Storage Server Config,
 // returning an error in case this config is invalid.
 func (cfg *StorageServerConfig) Validate() error {
-	if cfg == nil || cfg.State == StorageServerStateRIP {
+	if cfg == nil {
+		return ErrInvalidStorageServerConfig
```
GlenDC: Please change this error or create a new error constant, saying `ErrNilConfig`. It's both more accurate and more generic.

config/format.go
```diff
@@ -471,19 +473,7 @@ func (cfg *StorageServerConfig) Validate() error {
...
-			return true
-		}
-		return false
-	} else if other == nil {
-		return false
-	}
-
-	// both configs are given
-
+func (cfg StorageServerConfig) Equal(other StorageServerConfig) bool {
```
GlenDC: No I still want `StorageServerConfig` to be a pointer type here. If `cfg == nil` you can simply return `false`. So the signature should be: `func (cfg *StorageServerConfig) (other StorageServerConfig) bool`.

config/format_test.go
```diff
@@ -101,10 +101,9 @@ func TestNewStorageClusterConfig(t *testing.T) {
 	}
 
 	for _, invalidCase := range invalidStorageClusterConfigYAML {
-		cfg, err := NewStorageClusterConfig([]byte(invalidCase))
+		_, err := NewStorageClusterConfig([]byte(invalidCase))
```
GlenDC: ensure to revert the changes to this file again

nbd/ardb/storage/deduped.go
```diff
@@ -309,7 +309,7 @@ func copyDedupedSameServerCount(sourceID, targetID string, sourceCluster, target
 	for index, sourceCfg := range sourceCluster.Servers {
 		targetCfg = targetCluster.Servers[index]
 
-		if sourceCfg.Equal(&targetCfg) {
+		if sourceCfg.Equal(targetCfg) {
```
GlenDC: Nice looks much better indeed!

zeroctl/cmd/backup/import.go
```diff
@@ -120,7 +120,7 @@ func importVdisk(cmd *cobra.Command, args []string) error {
 	}
 
 	// copy nbd's tlog metadata
-	return nbdtlog.CreateMetadata(vdiskCmdCfg.VdiskID, lastFlushedSeq, clusterConf)
+	return nbdtlog.CreateMetadata(vdiskCmdCfg.VdiskID, lastFlushedSeq, &clusterConf)
```
GlenDC: You forgot to streamline here. The main task in this issue is ensuring that:

+ Optional config values use a pointer type for config values;
+ Required config values use a value type (the exception is where we require the values to be passed by reference, in those cases you still use pointer types);

right now `tlog.CreateMetadata` is a perfect example of how you can easily have a nil-pointer dereference panic on runtime by passing `nil` right now. As it simply dereferences the pointer to pass it to `ardb.NewCluster`. The fact that it simply dereferences it shows that the config value is required. And thus that function signature should change to `func CreateMetadata(vdiskID string, lastFlushedSequence uint64, clusterCfg config.StorageClusterConfig) error`.

zeroctl/cmd/copyvdisk/vdisk.go
```diff
@@ -91,11 +91,12 @@ func copyVdisk(cmd *cobra.Command, args []string) error {
 		}
 
 		if sourceVdiskNBDConfig.StorageClusterID != targetClusterID {
-			targetClusterConfig, err = config.ReadStorageClusterConfig(configSource, targetClusterID)
+			targetClusterConfigBuf, err := config.ReadStorageClusterConfig(configSource, targetClusterID)
```
GlenDC: Revert this file once you have reverted the signature of `config.ReadStorageClusterConfig`.

config/composition_api.go
```diff
@@ -41,23 +41,25 @@ func ReadNBDStorageConfig(source Source, vdiskID string) (*NBDStorageConfig, err
 	// if template storage ID is given, fetch it
 	if nbdConfig.TemplateStorageClusterID != "" {
 		// Fetch Template Storage Cluster Config
-		nbdStorageConfig.TemplateStorageCluster, err = ReadStorageClusterConfig(
+		templateStorageCluster, err := ReadStorageClusterConfig(
```
GlenDC: revert this change please

config/composition_api.go
```diff
@@ -41,23 +41,25 @@ func ReadNBDStorageConfig(source Source, vdiskID string) (*NBDStorageConfig, err
...
 			return nil, err
 		}
+		nbdStorageConfig.TemplateStorageCluster = templateStorageCluster
 	}
 
 	// if slave storage ID is given, fetch it
 	if nbdConfig.SlaveStorageClusterID != "" {
 		// Fetch Template Storage Cluster Config
-		nbdStorageConfig.SlaveStorageCluster, err = ReadStorageClusterConfig(
+		slaveStorageCluster, err := ReadStorageClusterConfig(
```
GlenDC: revert this change please

config/composition_api.go
```diff
@@ -94,12 +96,13 @@ func ReadTlogStorageConfig(source Source, vdiskID string) (*TlogStorageConfig, e
 
 	// if slave storage ID is given, fetch it
 	if tlogConfig.SlaveStorageClusterID != "" {
-		tlogStorageConfig.SlaveStorageCluster, err = ReadStorageClusterConfig(
+		slaveStorageCluster, err := ReadStorageClusterConfig(
```
GlenDC: revert this change please

config/format.go
```diff
@@ -192,7 +192,7 @@ func (cfg *VdiskTlogConfig) Validate() error {
 func NewStorageClusterConfig(data []byte) (*StorageClusterConfig, error) {
 	clustercfg := new(StorageClusterConfig)
 
-	err := yaml.Unmarshal(data, clustercfg)
+	err := yaml.Unmarshal(data, &clustercfg)
```
GlenDC: wrong, it's created using `new`, so it's already a pointer value. How this still makes your tests pass is a mystery to me.

config/format_test.go
```diff
@@ -156,30 +156,22 @@ func TestStorageClusterConfigEqual(t *testing.T) {
 func TestStorageServerConfigEqual(t *testing.T) {
 	assert := assert.New(t)
 
-	var a, b *StorageServerConfig
-	assert.True(a.Equal(b), "both are nil")
+	var a, b StorageServerConfig
+	assert.True(a.Equal(b), "both are empty")
```
GlenDC: `nil` is a better adverb than `empty` in this case

nbd/ardb/storage/deduped.go
```diff
@@ -302,14 +288,14 @@ func CopyDeduped(sourceID, targetID string, sourceCluster, targetCluster *config
 	return copyDedupedDifferentServerCount(sourceID, targetID, sourceCluster, targetCluster)
 }
 
-func copyDedupedSameServerCount(sourceID, targetID string, sourceCluster, targetCluster *config.StorageClusterConfig) error {
+func copyDedupedSameServerCount(sourceID, targetID string, sourceCluster config.StorageClusterConfig, targetCluster *config.StorageClusterConfig) error {
```
GlenDC: should be `func copyDedupedSameServerCount(sourceID, targetID string, sourceCluster, targetCluster config.StorageClusterConfig) error`

nbd/ardb/storage/deduped.go
```diff
@@ -344,7 +330,7 @@ func copyDedupedSameServerCount(sourceID, targetID string, sourceCluster, target
 	return nil
 }
 
-func copyDedupedDifferentServerCount(sourceID, targetID string, sourceCluster, targetCluster *config.StorageClusterConfig) error {
+func copyDedupedDifferentServerCount(sourceID, targetID string, sourceCluster config.StorageClusterConfig, targetCluster *config.StorageClusterConfig) error {
```
GlenDC: should be `func copyDedupedDifferentServerCount(sourceID, targetID string, sourceCluster, targetCluster config.StorageClusterConfig) error`

nbd/ardb/storage/storage.go
```diff
@@ -192,7 +192,7 @@ func VdiskExists(id string, t config.VdiskType, ccfg *config.StorageClusterConfi
 
 // ListBlockIndices returns all indices stored for the given storage.
 // This function will always either return an error OR indices.
-func ListBlockIndices(id string, t config.VdiskType, ccfg *config.StorageClusterConfig) ([]int64, error) {
+func ListBlockIndices(id string, t config.VdiskType, ccfg config.StorageClusterConfig) ([]int64, error) {
```
GlenDC: `s/ccfg/cfg`

PR #569, opened 2017-10-25
add --name <regexp> flag to zeroctl list vdisks

### LINKED ISSUE #462: make it possible to list template vdisks
Once we have the flow described in #460 in place, we'll need a way to list all the templates available in a storage cluster, this command should also works against the 0-hub to be able to list all templates available to imports.

GlenDC:
closes #462


GlenDC:
@zaibon can you please review this PR.

This is my solution as proposed in https://github.com/zero-os/0-Disk/issues/462#issuecomment-339473365 and https://github.com/zero-os/0-Disk/issues/462#issuecomment-339483229.

PR #570, opened 2017-10-26
Improve error handling config package

### LINKED ISSUE #568: Improve error handling config package
Part of #529, improve the error passing and handling in the config module.

* Import `pkg/errors`
* Add Err vars:
  * `ErrInvalidConfig`
  * `ErrInvalidStorageCluster`
  * `ErrNilStorage`
* Wrap errors with `pkg/errors`, e.g.:
  * wrap errors where context is added to to preserve cause of the error
  * `ETCDKey` returns error with ErrInvalidKey as Cause and the invalid `keytype` in the wrap
* Replace `NewInvalidConfigError` with `pkg/errors` and set `ErrInvalidConfig` as `Cause`
* Check errors with `pkg/errors.Cause()` where applicable (e.g.: `ReadConfig`)

chrisvdg:
Closes #568

### DIFF
--- a/config/format.go
+++ b/config/format.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	valid "github.com/asaskevich/govalidator"
-	"github.com/pkg/errors"
+	"github.com/zero-os/0-Disk/errors"
 	yaml "gopkg.in/yaml.v2"
 )
 
@@ -97,13 +97,13 @@ func (cfg *VdiskStaticConfig) Validate() error {
 	// validate properties in more detail
 
 	if !ValidateBlockSize(int64(cfg.BlockSize)) {
-		return errors.Errorf(
+		return errors.Newf(
 			"blockSize '%d' is invalid, should be equal-or-greater than 512 and be a power of 2",
 			cfg.BlockSize,
 		)
 	}
 	if (cfg.Size * gibibyteAsBytes) < cfg.BlockSize {
-		return errors.Errorf(
+		return errors.Newf(
 			"%d is an invalid size, has to be able to contain at least 1 block",
 			cfg.Size,
 		)
@@ -323,7 +323,7 @@ func (cfg *ZeroStorClusterConfig) Validate() error {
 
 	expectedServeCount := cfg.DataShards + cfg.ParityShards
 	if len(cfg.DataServers) != expectedServeCount {
-		return errors.Errorf(
+		return errors.Newf(
 			"invalid ZeroStorClusterConfig: expected %d (data+parity) servers, while %d servers were defined",
 			expectedServeCount,
 			len(cfg.DataServers),
@@ -484,7 +484,7 @@ func (cfg *StorageServerConfig) Validate() error {
 		return errors.New("storage server address not given while it is required")
 	}
 	if !valid.IsDialString(cfg.Address) {
-		return errors.Errorf("storage server address '%s' is an invalid dialstring", cfg.Address)
+		return errors.Newf("storage server address '%s' is an invalid dialstring", cfg.Address)
 	}
 
 	// all checks out
@@ -523,7 +523,7 @@ func (cfg StorageServerConfig) String() string {
 func isDialStringSlice(data []string) error {
 	for _, server := range data {
 		if !valid.IsDialString(server) {
-			return errors.Errorf("%s is not a valid dial string", server)
+			return errors.Newf("%s is not a valid dial string", server)
 		}
 	}
 	return nil
--- a/config/format.go
+++ b/config/format.go
@@ -1,10 +1,10 @@
 package config
 
 import (
-	"errors"
 	"fmt"
 
 	valid "github.com/asaskevich/govalidator"
+	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v2"
 )
 
@@ -67,7 +67,7 @@ func (cfg *NBDVdisksConfig) Validate() error {
 
 	_, err := valid.ValidateStruct(cfg)
 	if err != nil {
-		return fmt.Errorf("invalid NBDVdisksConfig: %v", err)
+		return errors.Wrap(err, "invalid NBDVdisksConfig")
 	}
 
 	return nil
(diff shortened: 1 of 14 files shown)


Gopkg.lock
```diff
@@ -446,6 +446,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "e331a16481d6db5ca4695bf757594409f4c1e1f77956143ffe1d156b080aa7e5"
+  inputs-digest = "c78bf3de753ff7fea25abac7c3eb6d0e5c7a52d09f4ea51ccbb0098ad227ea9e"
```
GlenDC: You updated the `Gopkg.lock`, but didn't update `Gopkg.toml` and you haven't added the new dependency to the `vendor/` folder either (`github.com/pkg/errors`).

config/format.go
```diff
@@ -471,14 +477,14 @@ func (cfg *StorageServerConfig) Validate() error {
 	}
 
 	if cfg.Database < 0 {
-		return errInvalidStorageDatabase
+		return errors.New("invalid database")
```
GlenDC: make this into a constant error

Gopkg.lock
chrisvdg: I guess one of our dependencies uses pkg/errors, it didn't add the new dependency because it was already there, I guess I can revert this

GlenDC: Oh yes better revert it it than for now.

errors/errors.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// ErrorSlice represents a slice of errors
+type ErrorSlice struct {
+	errors []error
+}
+
+// Add adds an error to ErrorSlice
+func (errs *ErrorSlice) Add(err error) {
+	errs.errors = append(errs.errors, err)
```
GlenDC: add a check `if err == nil { return }` please, this allows the user to return all errors, even if they are `nil`.

errors/errors.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Error implements the error interface
+func (errs *ErrorSlice) Error() string {
+	if errs.Len() == 0 {
```
GlenDC: no need for this check, you're result will be the same, as the for loop will not iterate. So you're only making the expected case (`Len() > 0`) slower, for no good reason.

errors/errors.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Cause returns the underlying cause of the error if possible.
+// If the error does not implement `Cause() error` it returns the full error
+func Cause(err error) error {
+	return errors.Cause(err)
+}
+
+// NewErrSlice returns an empty ErrorSlice
+func NewErrSlice() *ErrorSlice {
```
GlenDC: Rename this `NewErrorSlice` please.

errors/errors.go
```diff
@@ -0,0 +1,75 @@
...
+	return new(ErrorSlice)
+}
+
+// ErrorSlice represents a slice of errors
+type ErrorSlice struct {
```
GlenDC: This type was very much like the `ErrorSlice` type I implemented a couple of times in 0-Disk. However it always left me with a questioning feeling if this type is really good as it is.

One of the short comings I felt it had was what if you want to check if it contains a certain error. But perhaps that is not something to be done. Perhaps if you want one special error it would be much better to do something like:

```go
slice := errors.NewErrorSlice()
// add errors
return errors.WrapError(mySpecialErr, slice)
```

In which case perhaps the type is good enough.

That reminds me:

Implement a `WrapError` method, in case we want to wrap errors like this;

Also you expose the `ErrorSlice` type, which I guess is fine. But than you should make it work as a nil type without panics and errors. What is mean is that the following should (perhaps) work:

```
var errs ErrorSlice
// ...
errs.Add(myError)
// ...
return errs
```

errors/errors.go
```diff
@@ -0,0 +1,75 @@
...
+func Cause(err error) error {
+	return errors.Cause(err)
+}
+
+// NewErrSlice returns an empty ErrorSlice
+func NewErrSlice() *ErrorSlice {
+	return new(ErrorSlice)
+}
+
+// ErrorSlice represents a slice of errors
```
GlenDC: Would be good to also have an `AsError() error` which returns `nil` if the internal slice is empty, and returns itself otherwise. This is a useful method when you want to return this slice as an error result of a function. Because if you would simply return `ErrorSlice` you'll get a non-nil error, even when the internal slice is empty, which is not what you want.

Perhaps you have a better idea how you solve this though, but that is what I would propose.

errors/example_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	fmt.Println(err.Error())
+	fmt.Println(cause.Error())
+	// Output:
+	// annotation 3: annotation 2: annotation 1: cause of the error
+	// cause of the error
+}
+
+func ExampleErrorSlice() {
+	var errs errors.ErrorSlice
```
GlenDC: It's true that you can already use it as a nil type. However, because of how you implemented the `Error` method, you won't be able to return this local `errs` variable as an `error` type.

errors/errors.go
GlenDC: Without this method every user has to do `if errs.Len() > 0 { return errs}` when returning it, which is a bit cumbersome. Much cleaner to just be able to write `errs.AsError()`, I think.

chrisvdg: Didn't need this for the config package, but just saw I will need it for the nbd package ;)

GlenDC: You do also need this for the config package, as I can see that in this PR you are having that kind of `if` code that I mentioned in my previous comment (https://github.com/zero-os/0-Disk/pull/570#discussion_r147414411)

errors/errors.go
chrisvdg: `errors.WrapError(mySpecialErr, slice)` should return an error that has `mySpecialErr` as cause and slice.Error() as  annotation, or...?

GlenDC: No it would be a new struct `withError`, which also implement the `causer` and `Error` interface. Which would simply contain 2 errors (the cause and the other error). This way it doesn't have to do any work to convert the Error in a string, when it isn't needed. Because it's possible that the user only wants to use this error for its Cause in the end. So would be silly to have to do all that string conversion work for nothing.

errors/errors.go
```diff
@@ -41,35 +41,63 @@ func Cause(err error) error {
 	return errors.Cause(err)
 }
 
-// NewErrSlice returns an empty ErrorSlice
-func NewErrSlice() *ErrorSlice {
-	return new(ErrorSlice)
+// NewErrorSlice returns an empty error slice
+func NewErrorSlice() *errorSlice {
```
GlenDC: You cannot do this. When you have a public function all its types should be public as well.

PR #575, opened 2017-10-27
Add testable examples for root package

katia-e:


example_test.go
```diff
@@ -0,0 +1,75 @@
...
+	// Output:
+	// ae1c89d781f63c4dd6c8ec4703b711bed45966af278446749dbe0eed34eaedf3
+}
+
+func ExampleNewKeyedHasher() {
+	// given we have data and key ...
+	data := []byte("data to hash")
+	key := []byte("key")
+
+	// we can define a new unstance of default keyed hasher
```
GlenDC: instance*

example_test.go
```diff
@@ -0,0 +1,75 @@
...
+	hasher, err := NewKeyedHasher(key)
+	panicOnError(err)
+
+	// hasher is used to hash the data
+	h := hasher.HashBytes(data)
+	str := hex.EncodeToString(h)
```
GlenDC: I would do the spacing instead as following (as you did in your previous example):

```go
// hasher is used to hash the data
h := hasher.HashBytes(data)

str := hex.EncodeToString(h)
fmt.Println(str)
```

By the way, I think it would be even better if you define a helper function `printHex([]byte)` such that you can reduce the bottom of these examples to:

```go
// hasher is used to hash the data
h := hasher.HashBytes(data)
printHex(h)
```

example_test.go
```diff
@@ -0,0 +1,122 @@
...
+}
+func ExampleVersion_Compare() {
+	// given we have several versions of the zerodisk modules
+	versA := NewVersion(2, 3, 4, versionLabel("beta-2"))
+	versB := NewVersion(2, 3, 4, versionLabel("beta-2"))
+	versC := NewVersion(1, 1, 0, versionLabel("beta-1"))
+
+	// we can compare versions
+	diff := versA.Compare(versB)
+	fmt.Println(versionDiff(diff))
```
GlenDC: just do `fmt.Println(diff)`, remove that `versionDiff` function, it makes things less obvious.

Example is all about showing how to use the API. Knowing what a function returns is part of that API, which is right now something that you are hiding.

It is ok that the output will look like

```
// Output:
// 0
// 1
// -1
```

Our audience for these examples are our fellow programmers, so they should be used to interpret such values to their meanings. We're following a convention here, also applied in the Golang std code, so all is good.

example_test.go
```diff
@@ -0,0 +1,122 @@
...
+	// given we have a new version of the zerodisk modules
+	// defined by ...
+	var (
+		major, minor, patch uint8
+		label               *VersionLabel
+	)
+	major, minor, patch, label = 2, 3, 4, versionLabel("beta-2")
+
+	// we can define a new version of the zerodisk modules
+	ver := NewVersion(major, minor, patch, label)
```
GlenDC: I would replace all these lines by:

```go
label := VersionLabel{'b', 'e', 't', 'a', '-', '2'}
ver := NewVersion(2, 3, 4, versionLabel
```

This makes things more compact, doesn't use any private functions and makes the code copy-pastable. Also no need to make clear what is major, minor, patch and label, as that is clear from the function signature which is bundled with this example.

example_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+func printHex(h Hash) {
+	str := hex.EncodeToString(h)
+	fmt.Println(str)
+}
+
+func ExampleHashBytes() {
+	// given we have two sets of data ...
+	data := []byte("data to hash")
```
GlenDC: please rename these variable names to `dataA` and `dataB`.

Perhaps it's me, but reading diff makes me think about some data diff, not about data different. Instead of calling it `dataDiff` you would better have called it `otherData`. But even so. I think `dataA` and `dataB` is better, as it makes it more neutral.

example_test.go
```diff
@@ -0,0 +1,122 @@
+package zerodisk
+
+import (
+	"encoding/hex"
+	"fmt"
+)
```
GlenDC: also add `"github.com/zero-os/0-Disk"` as an imported dependency here. This way you examples become `zerodisk.HashBytes`, etc...

This also forces you to only use the public API, which is good, as your examples should be able to be used pretty much directly as a copy paste (except for the few helper functions we define locally in this example_test file, which should only ever be used to make things more compact/focussed.

example_test.go
```diff
@@ -0,0 +1,108 @@
...
+	label := zerodisk.VersionLabel{'b', 'e', 't', 'a', '-', '2'}
+	ver := zerodisk.NewVersion(2, 3, 4, &label)
+
+	fmt.Println(ver)
+	// Output: 2.3.4-beta-2
+}
+
+func ExampleVersion_Compare() {
+	// given we have several versions of the zerodisk modules
+	labelA := zerodisk.VersionLabel{'b', 'e', 't', 'a', '-', '2'}
```
GlenDC: I wouldn't use a label for this example, as we are trying to showcase Compare, and the label gets ignored for comparing versions. So no need to clutter these examples.

PR #576, opened 2017-10-27
integrate slave-cluster blockStorage in tlog slave syncer

### LINKED ISSUE #475: tlogserver syncs to primary cluster instead of slave cluster
While working on the self-healing storage code (with all the consequences it has for primary, slave and template cluster), I was going once again through the tlog code, this time focussing a bit more on the slave sync code.

Here I noticed that the code which is responsible for keeping the slave cluster in sync with the data the tlogserver has is writing to the primary cluster instead of the slave cluster.

Here is a break down of the current code:

In [`tlog/tlogserver/slavesync/slave_sync.go#L121-L127`](https://github.com/zero-os/0-Disk/blob/196c18e382ae839c9fee9bca457bea49890539ce/tlog/tlogserver/slavesync/slave_sync.go#L121-L127) we create a `tlog.Player`:
```go
func (ss *slaveSyncer) init() error {
	// tlog replay player
	player, err := player.NewPlayer(ss.ctx, ss.configSource, ss.apc.VdiskID, ss.apc.PrivKey, ss.apc.DataShards, ss.apc.ParityShards)
	if err != nil {
		return err
	}
	ss.player = player
// ..
```

This is used to write data to an internal blockstorage which is created as part of the `NewPlayer` logic using the given `ss.configSource`. This can be seen in [`tlog/tlogclient/player/player.go#L33-L66`](https://github.com/zero-os/0-Disk/blob/master/tlog/tlogclient/player/player.go#L33-L66):
```go
// ...
       // get config to create block storage
	vdiskCfg, err := config.ReadVdiskStaticConfig(source, vdiskID)
	if err != nil {
		return nil, err
	}
	nbdCfg, err := config.ReadNBDStorageConfig(source, vdiskID, vdiskCfg)
	if err != nil {
		return nil, err
	}

	// create static provider,
	// as the tlog player does not require hot reloading.
	ardbProvider, err := ardb.StaticProvider(*nbdCfg, nil)
	if err != nil {
		return nil, err
	}

	blockStorage, err := storage.NewBlockStorage(storage.BlockStorageConfig{
		VdiskID:         vdiskID,
		TemplateVdiskID: vdiskCfg.TemplateVdiskID,
		VdiskType:       vdiskCfg.Type,
		BlockSize:       int64(vdiskCfg.BlockSize),
	}, ardbProvider)
// ...
```

Thus you can see that the provider is created in the `ardb.StaticProvider` constructor using the given `nbdConfig` (which contains all clusters), for the used internal block storage in that player.

+ That static provider is as it says static and does not support hot-reloading, this is bad as this player is in this instance used throughout the entire lifetime of a tlogserver-vdisk object, and thus even if the config is updated in its lifetime, it won't be updated;
+ But this isn't even the issue I would like to address here. The real issue is that it will create a provider using the Primary Cluster, and if supported even template cluster. The slave cluster will never be used and data will be written to the Primary Cluster, to which the nbdserver is also writing;

This can be seen in [`nbd/ardb/provider.go`](https://github.com/zero-os/0-Disk/blob/196c18e382ae839c9fee9bca457bea49890539ce/nbd/ardb/provider.go):

```go
// StaticProvider creates a Static Provider using the given NBD Config.
func StaticProvider(cfg config.NBDStorageConfig, pool *RedisPool) (ConnProvider, error) {
	if pool == nil {
		pool = NewRedisPool(nil)
	}


	provider := &staticRedisProvider{redisPool: pool}
	provider.setConfig(&cfg)
	return provider, nil
}

// ...

func (rp *staticRedisProvider) setConfig(cfg *config.NBDStorageConfig) {
	rp.dataConnectionConfigs = cfg.StorageCluster.DataStorage
	rp.numberOfServers = int64(len(rp.dataConnectionConfigs))

	rp.metaConnectionConfig = cfg.StorageCluster.MetadataStorage

	if cfg.TemplateStorageCluster == nil {
		rp.templateDataConnectionConfigs = nil
		rp.numberOfTemplateServers = 0
	} else {
		rp.templateDataConnectionConfigs = cfg.TemplateStorageCluster.DataStorage
		rp.numberOfTemplateServers = int64(len(rp.templateDataConnectionConfigs))
	}
}
```

As you can see no Slave Cluster is used or accesses anywhere, it only deals with Primary and Template clusters.

So that's an issue as the tlogserver should sync data to the slave cluster, not the primary cluster. On top of that the configuration should support hot reloading as the config might be updated over the course of its life time.

GlenDC:
fixes #475


GlenDC:
To finish this PR I'll need your help on this @iwanbk, you'll need to put in a bit of work here as well, as the code as it is won't work. I'll divide the rest of this comment in sections. What I'm describing here is to get us started on what has to be done in order to finish this PR.

---

https://github.com/zero-os/0-Disk/blob/3edc4cb085e074664cb96877b12a8b7ceec274b8/tlog/tlogserver/server/vdisk.go#L531-L546

I disabled this code (commenting it out), even though it can just be removed.

Even if a slaveStorageClusterID isn't configured at startup, what if it gets configured on runtime (when the vdisk is already mounted)?

+ This does mean that we need to find a way to not make it sync if no slave cluster is currently defined (this can be checked by doing a check such as `slaveCluster.ServerCount() == 0`);
+ It also means that somehow we'll have to do a "massive" sync if we suddenly get a slave cluster defined, after first it not being defined (should we somehow also subscribe to the config to auto-detect this, or just do that sync next time we get an incoming command?);

---

https://github.com/zero-os/0-Disk/blob/3edc4cb085e074664cb96877b12a8b7ceec274b8/tlog/tlogserver/server/vdisk.go#L483-L496

I was very surprised to see this code. We don't use `SIGHUP` anywhere anymore (except for local (dev) config-reloading, and even this is done behind the scenes for you, so no need to have it somewhere else).

Why restart the slave sync when you receive a `SIGHUP`, is this because you arranged that kind of deal with the AYS team, or is it because you are thinking about config reloading?

Either way, I am puzzled to see this code.

---

Further:

+ I've removed the `slaveStorageClusterID` from the `VdiskTlogConfig` as it is no longer needed. It's a bit silly to require the user to configure that ID twice (and error-prone, because what if the ids are different by accident). Also the Slave Cluster Code is kept in the NBDConfig, as this is more about ardb storage logic than anything else, so I feel it naturally belongs there;
+ We'll need the slave_syncer to work in 2 modes:
   + `enabled`: this is the normal mode as it currently is, where it syncs incoming aggregations using the player;
   + `disabled`: this is possible when no slaveStorageClusterID is given. It is important that no syncing happens in this mode, as the storage Cluster will return a `ErrClusterNotDefined` when used in that mode;
   + When we switch to `enabled` mode (either because we just created it, or because we were `disabled` before), we'll need to do a sync using the `last_slave_sync_seq:<vdiskID>` meta variable that you are already using;

---

I ask for your help @iwanbk and doing this for me as much as you can, because I have a feeling that I'll break some stuff in that module if I would have to do it myself.

iwanbk:
> Even if a slaveStorageClusterID isn't configured at startup, what if it gets configured on runtime (when the vdisk is already mounted)?

Well, please take a look more closely, `manageSlaveSync` was supposed to be called on config reload.

> I was very surprised to see this code. We don't use SIGHUP anywhere anymore (except for local (dev) > config-reloading, and even this is done behind the scenes for you, so no need to have it somewhere else).

Yeah, i was surprised too when i saw it (but forgot to create issue about it), because i saw that `manageSlaveSync` already being modified on `etcd` addition but the `SIGHUP` part is not being removed
https://github.com/zero-os/0-Disk/blame/3edc4cb085e074664cb96877b12a8b7ceec274b8/tlog/tlogserver/server/vdisk.go#L532

> Why restart the slave sync when you receive a SIGHUP, is this because you arranged that kind of deal with the AYS team, or is it because you are thinking about config reloading?

It has nothing to do with AYS, it is only a matter of restarting goroutine.
Yes, about config reloading as restarting slave syncer is very cheap.

> I ask for your help @iwanbk and doing this for me as much as you can, because I have a feeling that I'll break some stuff in that module if I would have to do it myself.

Which part that you want me to do?

Also, #475 is only about primary/slave cluster, i think we can focus on it for now.

iwanbk:
> Also, #475 is only about primary/slave cluster, i think we can focus on it for now.

Next steps will be addressed at #586 and #587

nbd/ardb/storage/cluster.go
```diff
@@ -335,6 +335,359 @@ func (server primaryStorageServer) Config() config.StorageServerConfig {
...
+	}
+
+	// compute server index of first available server
+	serverIndex, err := ardb.FindFirstServerIndex(sc.serverCount, sc.serverOperational)
+	if err != nil {
+		sc.mux.RUnlock()
+		return nil, err
+	}
+	cfg := sc.servers[serverIndex]
+	sc.mux.RUnlock()
```
iwanbk: i can see that it is unwise to simply use `defer RUnlock` here, i can also see that the protected code shouldn't be crash.
But having three unlocks might become problem later.
However, i'm not really sure about this, so feel free to leave it as is.

nbd/ardb/storage/cluster.go
```diff
@@ -335,6 +335,359 @@ func (server primaryStorageServer) Config() config.StorageServerConfig {
...
+		return nil, err
+	}
+
+	return slaveCluster, nil
+}
+
+// SlaveCluster defines a vdisk's slave cluster.
+// It supports hot reloading of the configuration
+// and state handling of the individual servers of a cluster.
+type SlaveCluster struct {
```
iwanbk: just curious, why we need to have `PrimaryCluster` and `SlaveCluster`?

From my quick check, the `SlaveCluster` has this check in `Do` and `DoFor`
```
if sc.serverCount == 0 {
 +		sc.mux.RUnlock()
 +		return nil, ErrClusterNotDefined
 +	}
 +
```

nbd/ardb/storage/cluster.go
```diff
@@ -335,6 +335,359 @@ func (server primaryStorageServer) Config() config.StorageServerConfig {
...
+	// returning an error is required though,
+	// as a slave custer is required where we do use this type of cluster.
+	if sc.serverCount == 0 {
+		sc.mux.Unlock()
+		return nil, ErrClusterNotDefined
+	}
+
+	ch := make(chan ardb.StorageServer)
+	go func() {
+		defer sc.mux.Unlock()
```
iwanbk: I think having `defer` in this unusual place should deserve a comment to explain

nbd/ardb/storage/cluster.go
```diff
@@ -335,6 +335,359 @@ func (server primaryStorageServer) Config() config.StorageServerConfig {
...
+			}
+		}
+	}()
+	return ch, nil
+}
+
+// ServerCount implements StorageCluster.ServerCount.
+func (sc *SlaveCluster) ServerCount() int64 {
+	sc.mux.RLock()
+	defer sc.mux.RUnlock()
```
iwanbk: curious why you use `defer` here

iwanbk:
LGTM, i basically can merge this PR, but still curious about some pieces of code (commented there).
So, i think it is better for me to merge after you replied my comment.

nbd/ardb/storage/cluster.go
GlenDC: It's OK for a slave cluster not to be defined. As a slave cluster is optional.

The reason why we have a Primary Cluster and a Slave Cluster, is because this is just the start of that code. The primary Cluster will turn into a PrimarySlaveClusterPair, which is used for the nbdserver. Which will switch for a node to the slave cluster if the primary node goes temporarily offline. And the slave cluster will become the SlavePrimaryClusterPair, which has similar logic, but here we use the slave cluster as the primary cluster.

Both pairs will have different logic, as they are specialized for their individual purposes.

nbd/ardb/storage/cluster.go
GlenDC: I think it's fine for now, but I do see your concern. Will reevaluate when working again on this code.

GlenDC:
Thanks for the review @iwanbk, will take your feedback into account first thing tomorrow and merge it in myself when I did.

PR #577, opened 2017-10-30
Snapshot version

### LINKED ISSUE #574: add version property to metadata of snapshots (vdisk backups)
When creating a backup we also store some metadata. All that info can be retrieved for a snapshot using the `zeroctl describe snapshot` command. For example:

```
$ zeroctl describe snapshot bar_1508696133 --pretty
{
  	"snapshotID": "bar_1508696133",
  	"blockSize": 131072,
  	"size": 14024704,
  	"created": "2017-10-22T13:15:33-05:00",
  	"source": {
  	  	"vdiskID": "bar",
  	  	"blockSize": 4096,
  	  	"size": 10737418240
  	}
}
```

So the main 5 properties right now that we have are `snapshotID`, `blockSize`, `size`, `created` and `source`. The only required ones are `snapshotID`, `blockSize` and `size`.

What I would like is to have a 6th property added, `version`. Such that the command above would give the following output instead:

```json
{
  	"snapshotID": "bar_1508696133",
  	"blockSize": 131072,
  	"size": 14024704,
  	"created": "2017-10-22T13:15:33-05:00",
        "version": "1.1.0-alpha-8",
  	"source": {
  	  	"vdiskID": "bar",
  	  	"blockSize": 4096,
  	  	"size": 10737418240
  	}
}
```

This property is also optional (for backwards compatibility reasons), and it can be assumed that the version is `"1.1.0"` when not given. As that is the semver in which the entire versioning was added.

The goal of this version is that if in the future for some reason we would have some breaking change or some new functionality, we could check the version (and compare it with `zerodisk.CurrentVersion`) to remove any ambiguity and do different things based on those comparisons.

As part of this issue you'll have to do the following:

- [x] Add `func VersionFromString(string) (Version, error)` to `github.com/zero-os/0-Disk/version.go`;
  + when the given string is empty it can be assumed that the version is `1.1.0` (make a private var for this);
- [x] Add an example for the new `VersionFromString` func;
- [x] Add unit tests for the new `VersionFromString` func;
- [x] Add a new optional `Version` (`string`) property to the `Metadata` struct of the `ardb/backup` module (make sure to document that this is the version of the toolchain);
  + make sure to validate the version in the `(*Header).Validate` method;
- [x] Modify the backup code such that it properly supports this new `Version` property;
- [x] Add the new optional `Version` (`string`) property to `SnapshotInfo` struct of the `zeroctl describe command` code (don't forget the `omitempty` annotation);
- [x] Support this new optional `Version` property in the `zeroctl describe command` code;

muhamadazmy:
Fixes #574


nbd/ardb/backup/import.go
```diff
@@ -70,6 +70,16 @@ func importBS(ctx context.Context, src StorageDriver, dst storage.BlockStorage,
 		return err
 	}
 
+	version, err := zerodisk.VersionFromString(header.Metadata.Version)
+	if err != nil {
+		return err
+	}
+
+	//support only current or older versions
```
GlenDC: No, please remove this check. For now we don't require any check as any backup made is currently still supported by the newest toolchain.

And even if we would have another breaking change in the future, we'll still would prefer to read the backup with some backward-compatibility (if possible), rather than simply throwing an error.

Either way, right now there is no issue, so no need to do this check at all, especially if it's just to return an error for no good reason at all.

version.go
```diff
@@ -128,6 +138,35 @@ func (v Version) String() string {
...
 }
 
+//VersionFromString returns a Version object from the string
+//representation
+func VersionFromString(ver string) (Version, error) {
+	if ver == "" {
+		return defaultVersion, nil
+	}
+
+	match := verRegex.FindStringSubmatch(ver)
```
GlenDC: Is there any good reason that you use a regexp instead of a `https://godoc.org/fmt#Scanf`?
By using your current regexp, you still have to parse the numbers yourself...
And even if you used a regexp here (which I don't think you should), your current pattern would still be way to liberal.

version_test.go
```diff
@@ -67,3 +68,58 @@ func TestVersionCompare(t *testing.T) {
...
+	} else {
+		t.Fatal(err)
+	}
+
+	//faulty version numbers
+	bad := []string{
+		"1.1-alpha",     //no patch number
+		"abcd",          //rubbish
+		"1.1.1.alpha-2", //label separated by . instead of -
+	}
```
GlenDC: also add some underflows and overflows

version_test.go
```diff
@@ -67,3 +68,58 @@ func TestVersionCompare(t *testing.T) {
...
+		}
+	}
+}
+
+func ExampleVersionFromString() {
+	v, err := VersionFromString("1.0.0-alpha")
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(v)
```
GlenDC: I would print Major, Minor, Patch and label on separate lines, this better showcases that indeed some conversion to an in-memory model has taken place. Because right now from the example it looks like it was just some type casting or w/e.

GlenDC:
Thank you for putting in the work @muhamadazmy! 👍

GlenDC:
@muhamadazmy please also address https://github.com/zero-os/0-Disk/issues/574#issuecomment-340485640 as that's another thing you have left to do.

Also move your (testable) example code to `github.com/zero-os/0-Disk/example_test.go` and import `github.com/zero-os/0-Disk` as a dependency, such that you can write `zerodisk.VersionFromString`, this makes the example copy-pastable.

version.go
muhamadazmy: I usually prefer regexp over scanf for the control you get from regex. In this case, i used regex because i didn't know what the label format is, if it can contain spaces or special characters. If the label allows spaces, scanf will fail to capture the full string to the end of the line, while regex can. this is why I made it liberal on capturing the label string.

If label *must be* a single word string, I can happily switch to using `scanf` instead.

version_test.go
muhamadazmy: I will add a case with numeric numbers outside the uint8 range, but i don't get how to test the underflow since the scan will fail to scan the negative values from the start. Unless i got what u mean wrong.

muhamadazmy:
@GlenDC I will move the `Example` function to a separate file, but even if i do so, it won't be possible to import `github.com/zero-os/0-Disk` because self import is not allowed in go.

muhamadazmy:
@GlenDC I applied all your comments, except the one regarding switching to `scanf` (check my comment https://github.com/zero-os/0-Disk/pull/577#discussion_r147911217).

version.go
GlenDC: No I think it's OK to go for a regexp, even though it's more expensive than manual parsing, it's not logic that we'll be executing often, so this is OK.

version.go
```diff
@@ -16,6 +19,13 @@ var (
 	CommitHash string
 	// BuildDate represents the date when this tool suite was built
 	BuildDate string
+
+	//version parsing regex
+	verRegex = regexp.MustCompile(`^(\d+)\.(\d+)\.(\d+)(?:-(.+))?$`)
```
GlenDC: What about this regexp instead:

```regexp
^([01]?[0-9]?[0-9]|2[0-4][0-9]|25[0-5]).([01]?[0-9]?[0-9]|2[0-4][0-9]|25[0-5]).([01]?[0-9]?[0-9]|2[0-4][0-9]|25[0-5])(?:-([A-Za-z0-9\-]{1,8}))?$
```

It ensures that all numbers are `uint8`, and it limits the optional label to a string between 1 and 8 characters (selected from a custom character class). If you think some other characters should be allowed do tell me, but I don't think spaces make sense for example.

version.go
```diff
@@ -91,6 +101,22 @@ type (
 	VersionLabel [8]byte
 )
 
+func (n VersionNumber) Major() uint8 {
+	return uint8(n & 0xFF0000 >> 16)
```
GlenDC: Personally I would change this to `(n >> 16) && 0xFF`. It guarantees the order of operation regardless of the language and I guess it's more or less personal preference.

version.go
```diff
@@ -124,8 +150,36 @@ func (v Version) String() string {
...
+		v, err := strconv.ParseUint(n, 10, 8)
+		if err != nil {
+			return Version{}, fmt.Errorf("failed to parse version number (%d): %v", n, err)
+		}
+
+		num[i] = uint8(v)
+	}
+
+	var label *VersionLabel
+	if l := strings.TrimSpace(match[4]); len(l) != 0 {
```
GlenDC: with my new proposes regexp this (trim) won't be needed, you can pretty much just assign the string using the `versionLabel` helper.

version.go
```diff
@@ -124,8 +150,36 @@ func (v Version) String() string {
...
+		return defaultVersion, nil
+	}
+
+	match := verRegex.FindStringSubmatch(ver)
+	if len(match) == 0 {
+		return Version{}, fmt.Errorf("not a valid version format '%s'", ver)
+	}
+	num := make([]uint8, 3)
+	for i, n := range match[1:4] {
+		v, err := strconv.ParseUint(n, 10, 8)
```
GlenDC: with my new proposed regexp you can ignore the `error` result, as we already know for sure it's a valid unsigned 8 bit number.

version_test.go
```diff
@@ -67,3 +66,50 @@ func TestVersionCompare(t *testing.T) {
...
+	} else {
+		t.Fatal(err)
+	}
+
+	//faulty version numbers
+	bad := []string{
+		"1.1-alpha",                                  //no patch number
+		"abcd",                                       //rubbish
+		"1.1.1.alpha-2",                              //label separated by . instead of -
+		"123671.1231245.0-very-very-very-long-label", //numbers out of rage of uint8,
```
GlenDC: I would replaces this with multiple cases:

```go
"256.1.1",
"1.256.1",
"1.1.256",
"256.256.256",
"1.1.1-123456789", // we'll fail too long labels as well
```

example_test.go
```diff
@@ -7,6 +7,23 @@ import (
...
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(v.Number.Major())
+	fmt.Println(v.Number.Minor())
+	fmt.Println(v.Number.Patch())
+	fmt.Println(v.Label)
+
+	// Output: 1
```
GlenDC: put the `1` on a new line please

version.go
muhamadazmy: Yeah, this is just in case there were any garbage in the higher order bits, but from the code this should not be possible. Also it will make it look consistent with the `Minor` and `Patch` methods

GlenDC: Yeah I know why you do the masking, but in my version that is still being done as well.

example_test.go
muhamadazmy: Didn't know this was possible :D

PR #588, opened 2017-11-01
add ARDB failure scenarios spec

### LINKED ISSUE #355: document (and collect) all fatal scenarios
There are critical fail scenarios such as the fact that one (or multiple) of the ARDB servers used for a single vdisk, are suddenly not-reachable or not working any longer. This is an example of a fatal scenario which doesn't get handled. Some fatal scenarios might already be handled.

The goal of this issue is to document all these scenarios, regardless if they are already handled or not. This should help us to plan the issues to come regarding self-healing, which will be the biggest focus point for the 0-Disk services starting from this milestone.

This issue could be seen as part of #308.

GlenDC:
Closes #355.

Finishes the ARDB failure scenarios spec which was started a long time ago. A lot has changed since, but it seemed like a good time to fill in that spec, now that we're starting to (finally) integrate the slave sync feature.

Available in rendered markdown @ https://github.com/zero-os/0-Disk/blob/ardb-failure-scenarios/specs/ardbfailurescenarios.md

### DIFF
--- a/specs/ardbfailurescenarios.md
+++ b/specs/ardbfailurescenarios.md
@@ -1,27 +1,108 @@
-# ARDB failure scenario's
+# ARDB failure scenarios
 
 Disks fail, nodes fail and ardb's can crash. What actions does we have to take in case of ardb failure?
 
 It is not as simple as just switching to the slave since redis does asynchronous replication so there is a small window for data loss.
 
-## NBDServer
+## Failure Scenarios
 
-### Content
-1. Master ardb fails  
-**TODO**
-2. Slave ardb fails but the master is still alive  
-**TODO**
-3. Both failed  
-**TODO**
+### a master node fails
 
-### Metadata
-1. Master ardb fails  
-**TODO**
-2. Slave ardb fails but the master is still alive  
-**TODO**
-3. Both failed  
-**TODO**
+#### no slave cluster is supported or configured
 
-## TLogServer
+When a master node fails and no slave cluster is configured or available,
+the vdisk will fail as soon as an operation touches the master node which failed.
 
-**TODO**
+#### a slave cluster is supported and configured
+
+When a master node fails and a slave cluster is configured and available,
+the following will immediately happen:
+
+1. The TlogServer will be notified to no longer sync to that slave node of that pair until further notice, the configuration of that slave cluster gets reloaded or that slave cluster gets replaced;
+2. AYS will be notified via an STDERR log that this master node failed, so that it can start considering what to do next;
+
+As soon as AYS made a decision on what to do, it will update the configuration of the master cluster:
+
++ If the master node that previously failed was restored:
+    1. It will be marked as `repair`;
+    2. This will trigger to copy all data for that vdisk, from the slave node to the master node;
+    3. The TlogServer will be instructed that it can sync to the slave shard in question once again;
+    4. The master (primary) node will be marked as `online` oce again;
+    5. AYS will be notified via an STDERR that the primary cluster has been repaired;
++ If the master node was deemed to be offline forever:
+    1. It will be marked as `respread`;
+    2. This will trigger to respread all data of that node, from the slave server to the primary cluster;
+    3. Both the primary and slave node will be marked as `rip` and can be forgotten;
+    4. AYS will be notified that the primary and slave node has been marked `rip` for this vdisk;
+
+If in the middle of any of these processes something goes wrong the vdisk will become unusable and the AYS will be notified via an STDERR log to let it know what went wrong. The vdisk will be unmounted in the meanwhile.
+
+#### Summary
+
++ The server count of the primary and slave cluster will always be equal (anything else is a critical error);
++ When a primary node is marked as `rip` its linked slave node will always be marked as `rip` as wel, and vice versa;
++ Data can be copied between slave and master clusters for restore purposes;
++ Once a server is marked `rip`, it is expected that it will never change state again;
+
+### a slave node fails
+
+When a slave node fails the following will happen:
+
+1. The TlogServer will stop syncing to it and will notify the nbdserver;
+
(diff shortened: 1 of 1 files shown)


specs/ardbfailurescenarios.md
```diff
@@ -1,27 +1,119 @@
...
++ It a slave server (node) is marked `RIP`, than there would be no longer a "backup" server for that primary server. Meaning that if the primary server would go suddenly offline, this would result in a critical error as there is no paired server available. This is a shame as there is probably still an entire slave cluster of other servers available for usage;
+
+Allowing a difference is server count between primary and slave servers would:
+
++ Make it possible to attach a slave cluster with a bigger or smaller amount of servers, making it for example possible to have a slave cluster of just 5 servers, while you have a primary cluster of 20 servers (whether this is a good idea is another question, but at least it would be possible);
++ It would solve both issues that I mentioned above;
+
+However it makes things a lot more complex. And we would need to find a proven method which ensures that data is always available. No matter which and when slave and primary servers go online/offline. Finding such a method and proving it can do so might be the best way to answer this open question.
```
GlenDC: I've started prototyping the method I had in mind. You can see the result here: https://github.com/GlenDC/zerodisk/compare/master...GlenDC:prototype-primary-slave-server-index

It seems to work. The next step would be to prototype the primitives we have to applied in-between state changes to ensure the data is always available, when it is expected to be available.

specs/ardbfailurescenarios.md
```diff
@@ -1,27 +1,119 @@
...
+
+> Could we allow a different amount of primary and slave servers?
+
+Mapping a primary node directly to a slave node keeps everything very simple. On top of that it makes it also very cheap to restore a primary node, as you simply have to copy the data from one slave node to ensure the primary node is repaired. However it also comes at some issues:
+
++ If a primary server (node) is marked `RIP`, than this index will no longer be used and thus skipped. If however the paired slave node still functions fine, it would be a shame that from now on we could no longer use that slave node. Seems like wasted resources to me (even though we could forcefully bring it offline from an AYS level, to save resources, I suppose);
++ It a slave server (node) is marked `RIP`, than there would be no longer a "backup" server for that primary server. Meaning that if the primary server would go suddenly offline, this would result in a critical error as there is no paired server available. This is a shame as there is probably still an entire slave cluster of other servers available for usage;
+
+Allowing a difference is server count between primary and slave servers would:
```
zaibon: `difference is server count` => `difference in server count` ?

specs/ardbfailurescenarios.md
```diff
@@ -1,27 +1,119 @@
...
+
++ What if this list gets corrupted or lost?
++ How can we be sure that this list is complete at all times?
+
+It also means we have more data to manage, and we for example would also have to copy/delete this when we copy/delete together with any other data of a vdisk.
+
+The reason why this modification list (or resilver list) was proposed is so that we don't have to copy all data from a given server to another server. However I start to think it might be the safest to do so:
+
++ When an ardb server comes back online, how to be sure that all data that was previously on there is stil there?
++ Deleting the vdisk data from an ardb server when it comes back online, to than copy all data from the slave cluster to that primary server might be a bit more expensive but it would keep the logic much simpler and it would keep it all much more resilient;
```
zaibon: Would be nice indeed to try to keep this logic as simple as possible.

Now would it really be needed to delete vdisk data before doing the copy ? Can't we just erased previous data. That would save a bit of time and still rewrite the full vdisk data.

specs/ardbfailurescenarios.md
```diff
@@ -1,27 +1,119 @@
...
+
++ If a primary server (node) is marked `RIP`, than this index will no longer be used and thus skipped. If however the paired slave node still functions fine, it would be a shame that from now on we could no longer use that slave node. Seems like wasted resources to me (even though we could forcefully bring it offline from an AYS level, to save resources, I suppose);
++ It a slave server (node) is marked `RIP`, than there would be no longer a "backup" server for that primary server. Meaning that if the primary server would go suddenly offline, this would result in a critical error as there is no paired server available. This is a shame as there is probably still an entire slave cluster of other servers available for usage;
+
+Allowing a difference is server count between primary and slave servers would:
+
++ Make it possible to attach a slave cluster with a bigger or smaller amount of servers, making it for example possible to have a slave cluster of just 5 servers, while you have a primary cluster of 20 servers (whether this is a good idea is another question, but at least it would be possible);
```
zaibon: If we go that direction, I think we should also support dynamic resizing of a cluster. So being able to add/remove new ardb server from a cluster (master or slave). This would imply to respread the data over the new shards. This flow need to be thought out properly.

specs/ardbfailurescenarios.md
GlenDC: Erasing previous data is the same as deleting the vdisk data, no? I don't see what you mean differently if not that.

specs/ardbfailurescenarios.md
GlenDC: Not sure why it would have to mean that we need to support dynamic resizing. I think that would yet be another feature (which does influence the design a lot):

+ Bringing a server permanently offline should be simple enough. We would simply need to respread its data and than mark it as `RIP`. Pretty simple;
+ Bringing a server back online after it was dead is a lot harder. As you would need to somehow move the data that was now spread across the other servers. I suppose it should still be do-able as we can probably calculate where the data might have been, but I'm not too sure about this. Would need some prototyping/working out of the idea to know for sure;

On top of that we cannot simply add/remove servers:

+ As this would modify the total serverCount and thus affect how the in-memory model will fetch and store data. Doing so would mean we need to reshuffle a lot of data within the cluster to make all the previous data accessible on the newly assumed server indices;

I do see why that kind of feature would be desirable. But we better agree that this is something we want, because I do not think this is easily achievable with the current way we compute the server indices (and thus do the spreading).

PR #591, opened 2017-11-03
delete tlog data

### LINKED ISSUE #281: deleted any existing tlog-related (meta)data during delete vdisks
depends on:

+ #273;
+ functionality to delete tlogserver-dictated (meta)data;

open questions:

+ Should we also delete all aggregations? 
  + And what to do if the aggregations are used by multiple chains (see: forks)?

iwanbk:
Fixes #281

### DIFF
--- a/tlog/stor/delete.go
+++ b/tlog/stor/delete.go
@@ -1,0 +1,48 @@
+package stor
+
+import (
+	"github.com/zero-os/0-Disk/log"
+	"github.com/zero-os/0-Disk/tlog"
+	stormeta "github.com/zero-os/0-stor/client/meta"
+)
+
+// Delete tlog metadata and data of this vdisk
+func (c *Client) Delete() error {
+	storMetaCli, err := stormeta.NewClient(c.metaShards)
+	if err != nil {
+		return err
+	}
+
+	for wr := range c.Walk(0, tlog.TimeNowTimestamp()) {
+		if err := c.deleteData(storMetaCli, wr.StorKey, wr.Meta, wr.RefList); err != nil {
+			return err
+		}
+	}
+	// delete first meta
+	if err := c.metaCli.deleteMeta(c.firstMetaEtcdKey); err != nil {
+		return err
+	}
+	// delete last meta
+	return c.metaCli.deleteMeta(c.lastMetaEtcdKey)
+}
+
+func (c *Client) deleteData(storMetaCli *stormeta.Client, key []byte, md *stormeta.Meta, refList []string) error {
+	if len(refList) == 1 {
+		if refList[0] != c.vdiskID { // cheap strict checking, it should never happen
+			log.Errorf("unexpected reference list `%v` for vdisk `%v`", refList, c.vdiskID)
+			return nil
+		}
+		return c.storClient.DeleteWithMeta(md)
+	}
+	// remove meta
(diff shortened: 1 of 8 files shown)


zeroctl/cmd/delvdisk/vdisk.go
```diff
@@ -78,4 +84,10 @@ WARNING: until issue #88 has been resolved,
 	VdiskCmd.Flags().Var(
 		&vdiskCmdCfg.SourceConfig, "config",
 		"config resource: dialstrings (etcd cluster) or path (yaml file)")
+
+	VdiskCmd.Flags().StringVar(
+		&vdiskCmdCfg.PrivKey,
+		"priv-key", "12345678901234567890123456789012",
+		"tlog private key")
```
GlenDC: please provide more information about the key requirements (e.g. length, is it fixed size, what algorithm is used?)

tlog/stor/delete.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// Delete tlog metadata and data of this vdisk
+func (c *Client) Delete() error {
+	storMetaCli, err := stormeta.NewClient(c.metaShards)
+	if err != nil {
+		return err
+	}
+
+	for wr := range c.Walk(0, tlog.TimeNowTimestamp()) {
```
GlenDC: can't we give `0,0`? Presumably when you call delete it's because you want to delete everything from that vdisk, no? I guess this will also give everything, but just curious why you wouldn't keep it simpler by just using `0,0`.

tlog/stor/delete.go
```diff
@@ -0,0 +1,39 @@
...
+	// delete first meta
+	if err := c.metaCli.deleteMeta(c.firstMetaEtcdKey); err != nil {
+		return err
+	}
+	// delete last meta
+	return c.metaCli.deleteMeta(c.lastMetaEtcdKey)
+}
+
+func (c *Client) deleteData(storMetaCli *stormeta.Client, key []byte, md *stormeta.Meta, refList []string) error {
+	if len(refList) == 1 && refList[0] == c.vdiskID {
```
GlenDC: why is the 2nd part (`refList[0] == c.vdiskID`) needed? How would it ever be possible that we walk over data, which only has 1 reference, but that reference is somehow not our vdisk? O.O

tlog/stor/delete.go
iwanbk: No.
0,0 works with `zeroctl`, yes. But in the end it calls the same code as this.

tlog/stor/delete.go
iwanbk: we shouldn't need it. just cheap strict checking.

GlenDC: would be good to somehow log an error about this though (and even better would be if we can somehow handle this error case somehow, as now you just ignore it. Also what will happen with the RemoteReference in that case? will it return an error or so?

iwanbk: log error, good idea.
what you mean by `handle`?

iwanbk: > what you mean by handle?

I mean, what kind of handling?

iwanbk: > happen with the RemoteReference in that case? will it return an error or so?

It seems what you mean is `RemoveReference`, right?

It won't return error i guess, but better to handle it more clearly.
Thanks for catching this :)

PR #592, opened 2017-11-03
tlog : reuse capnp encoder and decoder

### LINKED ISSUE #423: tlog server : reuse canpn decoder

iwanbk:
Closes #423

### DIFF
--- a/tlog/tlogserver/server/vdisk_conn.go
+++ b/tlog/tlogserver/server/vdisk_conn.go
@@ -42,9 +42,10 @@ func (vd *vdisk) handle(conn *net.TCPConn, br *bufio.Reader, respSegmentBufLen i
 		}
 	}
 
+	capnpDec := capnp.NewDecoder(br)
 	for {
 		// decode message
-		msg, err := capnp.NewDecoder(br).Decode()
+		msg, err := capnpDec.Decode()
 		if err != nil {
 			if errors.Cause(err) == io.EOF {
 				return nil // EOF in this stage is not an error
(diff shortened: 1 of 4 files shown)


tlog/tlogserver/server/vdisk_conn.go
```diff
@@ -42,9 +42,10 @@ func (vd *vdisk) handle(conn *net.TCPConn, br *bufio.Reader, respSegmentBufLen i
 		}
 	}
 
+	capnpDec := capnp.NewDecoder(br)
```
GlenDC: I don't know the capnp code (haven't look over it), but from the docs it seems you might need to call https://godoc.org/zombiezen.com/go/capnproto2#Decoder.ReuseBuffer once, so that it will reuse the buffer on subsequent decodes. Or am I misunderstanding something here?

iwanbk: see https://github.com/zero-os/0-Disk/issues/423#issuecomment-333726138

GlenDC: You misunderstand me I think.

I think what you did here is good. Except instead of doing here:

```go
capnpDec := capnp.NewDecoder(br)
```

shouldn't you be doing instead:

```go
capnpDec := capnp.NewDecoder(br)
capnpDec.ReuseBuffer()
```

iwanbk: > But simply calling ReuseBuffer in the end of loop is not working.

> Because after decode, we hold the capnp object in memory.
> Which means the buffer is indeed couldn't be reused because it is still in use.

Except what you want is to call it once, in the first, which doesnt make sense:
- how we reuse something that not exist yet
- to reuse it only once? seems not really useful
- to always reuse the buffer on each iteration? see my above comment

iwanbk: OK, i just checked the code&docs again, we indeed need to only call it once.
But still, we can't reuse the buffer, see my comment above

GlenDC: Fair enough :)

Issue #596, opened 2017-11-08
proposal: move slave syncer out of tlogserver


GlenDC:
// TODO:
// add intro, description, problems with current approach and finally solutions (propositions)

iwanbk:
Very interesting proposal, some quick into will be very helpful especially about the solution.
As slave syncer is indeed not normally a task for a tlog server.

If we can see some raw solution proposal soon, we might make this issue and #587 to block each other.

GlenDC:
Sure thing. Will add it tomorrow as soon as possible. Today I was too busy with coding work. Sorry for that.

GlenDC:
@iwanbk I added the description. It's quite a lot to take in though, and it's a bit chaotic as it's written directly from my mind. However I hope it helps us to think and decide about the future of the slave sync feature.

Feel free to discuss it tomorrow with others such as @zaibon as well.

iwanbk:
> However I think we all agreed since the beginning that syncing (meta)data to the slave cluster shouldn't be the responsibility of the tlogserver. That is also something highlighted by @robvanmieghem's comment from back in May: #38 (comment). There was however not a bad idea back than, and so we continued with the idea of implementing this slave syncer component within the tlogserver, using the tlogplayer.

Yep, it is correct.
Please also note that slave syncer also make the tlogserver slower

> There is no more extra communication (start/stop slave sync) required between nbdserver and tlogserver

For now, the extra communication from nbdserver is only wait for syncing process to be finished

> The slave syncer becomes a lot cheaper as there is no longer an entire tlogreplay process (which has to decode, decrypt, decompress, etc...)

No
The entire tlogreplay process only happens when the data source is from 0-stor.
if the data comes from tlogclient directly, tlogreplay don't need to decrypt & decompress.

iwanbk:
> An alternative solution that i had in mind would be to have some kind of message queue service (similar to or RabbitMQ itself). 

Instead of rabbitmq, we might want something more lightweight like [nsq](nsq.io)

> This would be used to the nbdserver for all data write calls. Than it would write all that data to the primary cluster, slave cluster and tlogserver.

Sounds reasonable for slave and tlog, but not sure about writing to primary cluster from an MQ.

GlenDC:
> Please also note that slave syncer also make the tlogserver slower

[...]

PR #597, opened 2017-11-09
Reactivate slave syncer

### LINKED ISSUE #586: tlog : reactivate slave syncer
slave syncer hasn't been maintained since long time ago as the other side of it (nbd part) also being disabled.
We need to reactivate (and maintain) it since we will use this feature again.

We can only start it after #475 finished.

iwanbk:
Fixes #586 

The main goal of this PR is to only re-activate the slave syncer feature.
It also do some refactoring to improve the code as the old code is unnecessary complex and outdated.


iwanbk:
new feature, redesign, and any other improvements will be addressed at #587  or #596

GlenDC:
Blocked by #596.

iwanbk:
> Blocked by #596.

I think you misunderstood my comment at https://github.com/zero-os/0-Disk/issues/596#issuecomment-343033099.

It is not this PR that should be blocked, but #587.
This PR should be not blocked because there is no point to block it. It mainly contains two things:
- refactor current code
- reactivate slave syncer (which is only small code in main.go) -> if you want, i can undo this change

iwanbk:
> reactivate slave syncer (which is only small code in main.go) -> if you want, i can undo this change

I have disabled this code, so i think it is good to be merged.

PR #599, opened 2017-11-09
fix zeroctl import doc

khaledkbadr:
Signed-off-by: Khaled K. Badr <khaledkbadr@outlook.com>

### DIFF
--- a/docs/zeroctl/commands/describe.md
+++ b/docs/zeroctl/commands/describe.md
@@ -16,6 +16,7 @@ The FTP information is given using the `--storage` flag,
 here are some examples of valid values for that flag:
 + `localhost:22`;
 + `ftp://1.2.3.4:200`;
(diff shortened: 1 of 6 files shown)


khaledkbadr:
>  However you made some mistakes by overwriting environment-defined variables with your results.

Yup, I noticed the changes but I thought they're good to go since I just copied the output of the `zeroctl import vdisk --help`

GlenDC:
Yeah that's also how I keep the docs up to date, but than you have to manually edit those environment-defined variables. I think it makes the docs much more accurate, as otherwise a user might for example think the default job count is _always_ `1`. While that's only true for you, as that's apparently how many CPU (cores) you have. (How come you work with a single core machine? o.O)

khaledkbadr:
> How come you work with a single core machine? o.O

I use 0-disk on top of an zero-os VM not on my real machine :)

PR #600, opened 2017-11-09
Simplify storage constructors API

### LINKED ISSUE #594: simplify storage constructors API, by relying on cached ConfigSource
Right now we allow variants of constructors within the storage API, to create something either from a BlockStorage, or from some kind of config directly.

> example: https://godoc.org/github.com/zero-os/0-Disk/nbd/ardb/storage#BlockStorageFromConfig VS. https://godoc.org/github.com/zero-os/0-Disk/nbd/ardb/storage#BlockStorageFromConfigSource

this is done for places (mostly tooling) where we want to use static configs (as in ones that do no need to ever be hot-reloaded), such that we can re-use configs.

However we could simplify that API a lot, by introducing a `ConfigSource` Wrapper Type, which would provide an in-memory cache, such that it would only ever load a config (for a given key) once, and after that it would serve that config directly from the (in-memory) cache.

chrisvdg:
- [x] Adds onceSource
- [x] Simplify storage API
- [x] Zeroctl used onceSource

Closes #594

### DIFF
--- a/config/source_once.go
+++ b/config/source_once.go
@@ -1,0 +1,70 @@
+package config
+
+import (
+	"context"
+	"sync"
+)
+
+// NewOnceSource creates a config source,
+// that fetches the key's value from the provided source and saves in memory
+// if the key is requested again, the value from cache will be returned.
+// WARNING: this source is only to be used for a config that should be static,
+// calling the Watch function on this source will result in a panic.
+func NewOnceSource(source Source) Source {
+	if source == nil {
+		panic("a nil source was provided")
+	}
+
+	return &onceSource{
+		source: source,
+		cache:  make(map[Key][]byte),
+	}
+}
+
+type onceSource struct {
+	source Source
+	cache  map[Key][]byte
+	mux    sync.RWMutex
+}
+
+// Get implements Source.Get
+func (o *onceSource) Get(key Key) ([]byte, error) {
+	o.mux.RLock()
+	val, ok := o.cache[key]
+	o.mux.RUnlock()
+	if ok {
+		return val, nil
+	}
+
+	// if not in cache, get from internal source
+	// and store it in the cache
+	val, err := o.source.Get(key)
+	if err != nil {
+		return nil, err
+	}
+	o.mux.Lock()
+	o.cache[key] = val
+	o.mux.Unlock()
+
+	return val, nil
+}
+
+// Watch implements Source.Watch
+func (o *onceSource) Watch(ctx context.Context, key Key) (<-chan []byte, error) {
+	panic("tried to call Watch on a static config source")
+}
+
+// MarkInvalidKey implements Source.MarkInvalidKey
+func (o *onceSource) MarkInvalidKey(key Key, vdiskID string) {
+	o.source.MarkInvalidKey(key, vdiskID)
+}
+
+// SourceConfig implements Source.SourceConfig
+func (o *onceSource) SourceConfig() interface{} {
+	return o.source.SourceConfig()
+}
+
+// Type implements Source.Type
+func (o *onceSource) Type() string {
+	return o.source.Type()
+}
--- a/config/source_once_test.go
+++ b/config/source_once_test.go
@@ -15,8 +15,11 @@ func TestOnceSource(t *testing.T) {
 	}, "Providing NewOnceSource with a nil source should cause a panic")
 
 	// create stubsource with data
-	vdiskID := "testVdisk"
-	clusterID := "testCluster"
+	const (
+		vdiskID   = "testVdisk"
+		clusterID = "testCluster"
+	)
+
 	originalSource := NewStubSource()
 	clusterCfg := StorageClusterConfig{
 		Servers: []StorageServerConfig{
@@ -85,8 +88,10 @@ func TestDifferentKeyTypeSameID(t *testing.T) {
 	require := require.New(t)
 
 	// create stubsource with data
-	vdiskID := "testVdisk"
-	clusterID := "testCluster"
+	const (
+		vdiskID   = "testVdisk"
+		clusterID = "testCluster"
+	)
 	originalSource := NewStubSource()
 
 	// create StaticVdiskConfig
--- a/config/source_once_test.go
+++ b/config/source_once_test.go
@@ -1,0 +1,127 @@
+package config
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestOnceSource(t *testing.T) {
+	require := require.New(t)
+
+	// nil source should panic
+	require.Panics(func() {
+		NewOnceSource(nil)
+	}, "Providing NewOnceSource with a nil source should cause a panic")
+
+	// create stubsource with data
+	vdiskID := "testVdisk"
+	clusterID := "testCluster"
+	originalSource := NewStubSource()
+	clusterCfg := StorageClusterConfig{
+		Servers: []StorageServerConfig{
+			StorageServerConfig{
+				Address: "1.2.3.4:5678",
+			},
+		},
+	}
+	originalSource.SetPrimaryStorageCluster(vdiskID, clusterID, &clusterCfg)
+
+	// pass it into a onceSource
+	source := NewOnceSource(originalSource)
+
+	// fetch key
+	cfgBS, err := source.Get(Key{ID: clusterID, Type: KeyClusterStorage})
+	require.NoError(err)
+	require.NotEmpty(cfgBS)
+
+	clusterCfgFromSource, err := NewStorageClusterConfig(cfgBS)
+	require.NoError(err)
+	require.True(clusterCfg.Equal(*clusterCfgFromSource),
+		"cluster config from source should be equal to original cluster config")
+
+	// twice (check for cache fetch trough code coverage)
+	cfgBS, err = source.Get(Key{ID: clusterID, Type: KeyClusterStorage})
+	require.NoError(err)
+	require.NotEmpty(cfgBS)
+
+	clusterCfgFromSource, err = NewStorageClusterConfig(cfgBS)
+	require.NoError(err)
+	require.True(clusterCfg.Equal(*clusterCfgFromSource),
+		"cluster config from source should be equal to original cluster config")
+
+	// update config and check if config is still as before the update
+	newCluster := StorageClusterConfig{
+		Servers: []StorageServerConfig{
+			StorageServerConfig{
+				Address: "8.7.6.5:4321",
+			},
+		},
+	}
+	originalSource.SetPrimaryStorageCluster(vdiskID, clusterID, &newCluster)
+
+	cfgBS, err = source.Get(Key{ID: clusterID, Type: KeyClusterStorage})
+	require.NoError(err)
+	require.NotEmpty(cfgBS)
+
+	clusterCfgFromSource, err = NewStorageClusterConfig(cfgBS)
+	require.NoError(err)
+	require.True(clusterCfg.Equal(*clusterCfgFromSource),
+		"cluster config from source should be equal to original cluster config")
+	require.False(newCluster.Equal(*clusterCfgFromSource),
+		"cluster config should not equal the updated cluster config")
+
+	// get an invalid key ID
+	_, err = source.Get(Key{ID: "an invalid ID", Type: KeyClusterStorage})
+	require.Error(err)
+
+	// try to watch
+	require.Panics(func() {
+		source.Watch(nil, Key{})
+	}, "calling Watch on a onceSource should cause a panic")
+}
+
+func TestDifferentKeyTypeSameID(t *testing.T) {
+	require := require.New(t)
+
+	// create stubsource with data
+	vdiskID := "testVdisk"
+	clusterID := "testCluster"
+	originalSource := NewStubSource()
+
+	// create StaticVdiskConfig
+	vdiskCfg := &VdiskStaticConfig{
+		BlockSize: 4096,
+		Size:      10,
+		Type:      VdiskTypeDB,
+	}
+	originalSource.SetVdiskConfig(vdiskID, vdiskCfg)
+
+	// create vdiskNBDConfig
+	clusterCfg := &StorageClusterConfig{
+		Servers: []StorageServerConfig{
+			StorageServerConfig{
+				Address: "1.2.3.4:5678",
+			},
+		},
+	}
+	originalSource.SetTemplateStorageCluster(vdiskID, clusterID, clusterCfg)
+
+	// pass it into a onceSource
+	source := NewOnceSource(originalSource)
+
+	// test if different key types with same ID are cashed separately
+	bsVdisk, err := source.Get(Key{ID: vdiskID, Type: KeyVdiskStatic})
+	require.NoError(err)
+
+	cacheCntBefore := len(source.(*onceSource).cache)
+
+	bsNBD, err := source.Get(Key{ID: vdiskID, Type: KeyVdiskNBD})
+	require.NoError(err)
+
+	cacheCntAfter := len(source.(*onceSource).cache)
+	require.True(cacheCntBefore < cacheCntAfter,
+		"cache count should be increased after getting a different key type with same ID")
+	require.NotEqual(bsVdisk, bsNBD,
+		"Returned config from source with different key types should be different")
+}
--- a/nbd/ardb/backup/backup.go
+++ b/nbd/ardb/backup/backup.go
@@ -25,8 +25,8 @@ type Config struct {
 	// NOTE: only used by export funcs (ignored by import funcs)
 	BlockSize int64
 
-	// Required: SourceConfig to configure the storage with
-	BlockStorageConfig config.SourceConfig
+	// Required: config Source to configure the storage with
+	ConfigSource config.Source
 
 	// Optional: BackupStoragDriverConfig used to configure the backup storage driver.
 	//  - When not given, defaults to LocalStorageDriver, using the DefaultLocalRoot as the path.
@@ -76,16 +76,11 @@ func (cfg *Config) validate() error {
 		}
 	}
 
-	err := cfg.BlockStorageConfig.Validate()
-	if err != nil {
-		return err
-	}
-
 	if cfg.JobCount <= 0 {
 		cfg.JobCount = runtime.NumCPU()
 	}
 
-	err = cfg.CompressionType.validate()
+	err := cfg.CompressionType.validate()
 	if err != nil {
 		return err
 	}
@@ -370,19 +365,13 @@ func isNilBlock(block []byte) bool {
 }
 
 // Create a block storage ready for importing/exporting to/from a backup.
-func createStorageConfig(vdiskID string, sourceConfig config.SourceConfig) (*storageConfig, error) {
-	storageConfigCloser, err := config.NewSource(sourceConfig)
-	if err != nil {
-		return nil, err
-	}
-	defer storageConfigCloser.Close()
-
-	vdiskConfig, err := config.ReadVdiskStaticConfig(storageConfigCloser, vdiskID)
+func createStorageConfig(vdiskID string, configSource config.Source) (*storageConfig, error) {
+	vdiskConfig, err := config.ReadVdiskStaticConfig(configSource, vdiskID)
 	if err != nil {
 		return nil, err
 	}
 
-	nbdStorageConfig, err := config.ReadNBDStorageConfig(storageConfigCloser, vdiskID)
+	nbdStorageConfig, err := config.ReadNBDStorageConfig(configSource, vdiskID)
 	if err != nil {
 		return nil, err
 	}
--- a/nbd/ardb/backup/backup.go
+++ b/nbd/ardb/backup/backup.go
@@ -364,24 +364,6 @@ func isNilBlock(block []byte) bool {
 	return true
 }
 
-// Create a block storage ready for importing/exporting to/from a backup.
-func createStorageConfig(vdiskID string, configSource config.Source) (*storageConfig, error) {
-	vdiskConfig, err := config.ReadVdiskStaticConfig(configSource, vdiskID)
-	if err != nil {
-		return nil, err
-	}
-
-	nbdStorageConfig, err := config.ReadNBDStorageConfig(configSource, vdiskID)
-	if err != nil {
-		return nil, err
-	}
-
-	return &storageConfig{
-		NBD:   *nbdStorageConfig,
-		Vdisk: *vdiskConfig,
-	}, nil
-}
-
 var (
 	errNilVdiskID    = errors.New("vdisk's identifier not given")
 	errStreamBlocked = errors.New("stream block fetcher is blocked waiting for next expected block")
--- a/nbd/ardb/backup/export.go
+++ b/nbd/ardb/backup/export.go
@@ -23,7 +23,7 @@ func Export(ctx context.Context, cfg Config) error {
 		return err
 	}
 
-	storageConfig, err := createStorageConfig(cfg.VdiskID, cfg.BlockStorageConfig)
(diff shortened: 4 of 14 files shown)


config/source_once.go
```diff
@@ -0,0 +1,75 @@
...
+	o.mux.Lock()
+	o.cache[key.ID] = val
+	o.mux.Unlock()
+
+	return val, nil
+}
+
+// Watch implements Source.Watch
+func (o *onceSource) Watch(ctx context.Context, key Key) (<-chan []byte, error) {
+	return nil, ErrWatchNotSupported
```
GlenDC: this shouldn't return anything, I wasn't joking when I said it should panic. If this is triggered it means the programmer is abusing this type. This method should never be called, returning an error makes it as if you would expect it's possible

config/source_once.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Type implements Source.Type
+func (o *onceSource) Type() string {
+	return o.source.Type()
+}
+
+// Type implements SourceCloser.Close
+func (o *onceSource) Close() error {
+	if closableSource, ok := o.source.(SourceCloser); ok {
```
GlenDC: no just return `nil`. It's up to the user to close the source if it wraps around a `SourceCloser`. So nothing to do here.

config/source_once_test.go
```diff
@@ -0,0 +1,44 @@
...
+	// fetch key
+	key, err := source.Get(Key{ID: clusterID, Type: KeyClusterStorage})
+	require.NoError(err)
+	require.NotEmpty(key)
+
+	// twice (check for cache fetch trough code coverage)
+	key, err = source.Get(Key{ID: clusterID, Type: KeyClusterStorage})
+	require.NoError(err)
+	require.NotEmpty(key)
+
```
GlenDC: somewhere here you should also try to update that config, and try to get it once again. It should still return the cached version (not that the user should ever do something stupid like this, but still, nice to ensure that the onceSource type returns the original cached version at all times

config/source_once_test.go
```diff
@@ -0,0 +1,44 @@
...
+				Address: serverAddress,
+			},
+		},
+	})
+
+	// pass it into a onceSource
+	source := NewOnceSource(originalSource)
+
+	// fetch key
+	key, err := source.Get(Key{ID: clusterID, Type: KeyClusterStorage})
```
GlenDC: you're fetching a `yaml` config, not a key, also please validate that the config is the one you've set

config/source_once_test.go
```diff
@@ -0,0 +1,44 @@
...
+	require.NoError(err)
+	require.NotEmpty(key)
+
+	// twice (check for cache fetch trough code coverage)
+	key, err = source.Get(Key{ID: clusterID, Type: KeyClusterStorage})
```
GlenDC: you're fetching a `yaml` config, not a key, also please validate that the config is the one you've set

config/source_once_test.go
```diff
@@ -0,0 +1,44 @@
...
+	require.NoError(err)
+	require.NotEmpty(key)
+
+	_, err = source.Get(Key{ID: "an invalid ID", Type: KeyClusterStorage})
+	require.Error(err)
+
+	// try to watch
+	_, err = source.Watch(nil, Key{})
```
GlenDC: you can still test this, by ensuring you are catching the panic (using a defer `recover` call). I think the assertify package might even have a helper function for this to help you test this easily

config/source_once.go
```diff
@@ -0,0 +1,75 @@
...
+import "context"
+import "sync"
+
+// NewOnceSource creates a config source,
+// that fetches the key's value from the provided source and saves in memory
+// if the key is requested again, the value from cache will be returned.
+// WARNING: this source is only to be used for a config that should be static
+// calling the Watch function on this source will return an error
+func NewOnceSource(source Source) SourceCloser {
+	o := &onceSource{
```
GlenDC: add a panic call in case `source == nil`

config/source_once.go
```diff
@@ -66,10 +70,8 @@ func (o *onceSource) Type() string {
 }
 
 // Type implements SourceCloser.Close
+// The user should directly call Close on the internal source
```
GlenDC: No need to mention this, as we don't even know a Source can have a `Close` function. All we know is the `Source interface`. So in fact you can and should even delete this Close function, as we do not care about it.

config/source_once.go
```diff
@@ -6,9 +6,13 @@ import "sync"
 // NewOnceSource creates a config source,
 // that fetches the key's value from the provided source and saves in memory
 // if the key is requested again, the value from cache will be returned.
-// WARNING: this source is only to be used for a config that should be static
-// calling the Watch function on this source will return an error
+// WARNING: this source is only to be used for a config that should be static,
+// calling the Watch function on this source will result in a panic.
+// Also providing a nil source will result in a panic.
 func NewOnceSource(source Source) SourceCloser {
```
GlenDC: Should return a `Source`, not a `SourceCloser`

config/source_once.go
```diff
@@ -6,9 +6,13 @@ import "sync"
 // NewOnceSource creates a config source,
 // that fetches the key's value from the provided source and saves in memory
 // if the key is requested again, the value from cache will be returned.
-// WARNING: this source is only to be used for a config that should be static
-// calling the Watch function on this source will return an error
+// WARNING: this source is only to be used for a config that should be static,
+// calling the Watch function on this source will result in a panic.
+// Also providing a nil source will result in a panic.
```
GlenDC: You can remove this (last) line, as it can be implied that we require a source as we use the provided one, and and no point do we mention that the given source is optional.

config/source_once.go
```diff
@@ -0,0 +1,69 @@
...
+// NewOnceSource creates a config source,
+// that fetches the key's value from the provided source and saves in memory
+// if the key is requested again, the value from cache will be returned.
+// WARNING: this source is only to be used for a config that should be static,
+// calling the Watch function on this source will result in a panic.
+func NewOnceSource(source Source) Source {
+	if source == nil {
+		panic("a nil source was provided")
+	}
+	o := &onceSource{
```
GlenDC: just return here directly, I don't see the point of assigning it to a local variable first.

config/source_once.go
```diff
@@ -0,0 +1,69 @@
...
+		source: source,
+		cache:  make(map[string][]byte),
+	}
+
+	return o
+}
+
+type onceSource struct {
+	source Source
+	cache  map[string][]byte
```
GlenDC: It should be `cache map[Key][]byte` instead of `cache map[string][]byte`.

Right now you can get name conflicts as it is not said that the ID part of a key is unique, and it is perfectly acceptable that different Key Types have the same ID.

config/source_once_test.go
```diff
@@ -0,0 +1,79 @@
+package config
+
+import "testing"
+import "github.com/stretchr/testify/require"
+
+func TestOnceSource(t *testing.T) {
```
GlenDC: ensure to add test code which validates that keys of different types with the same ID (e.g. a StaticVdiskConfig and VdiskNBDConfig using the same `vdiskID`), are not counted as one and the same config due to having the same ID

config/source_once.go
```diff
@@ -0,0 +1,69 @@
+package config
+
+import "context"
```
GlenDC: change to:

```go
import(
   "context"
   "sync"
)
```

It's a bit weird to see it as separate import calls.

nbd/ardb/backup/backup.go
```diff
@@ -55,6 +55,9 @@ type Config struct {
 	// or the data was encrypted/compressed using a different
 	// key/compressionType than the one given.
 	Force bool
+
+	// Optional: For functions that can use the onceConfig for working with static config
+	ConfigSource config.Source
```
GlenDC: Good idea, but don't make it optional, make it required. And you can than delete the `BlockStorageConfig` property (as it effectively should replace this property). Not need to make this backup config too complex.

tlog/copy/generator.go
```diff
@@ -70,7 +70,7 @@ func (g *Generator) GenerateFromStorage(parentCtx context.Context) (uint64, erro
 	defer pool.Close()
```
GlenDC: A bit in this code you still have:

```go
staticConf, err := config.ReadVdiskStaticConfig(g.configSource, g.sourceVdiskID)
	if err != nil {
		return 0, err
	}

	storageConf, err := config.ReadNBDStorageConfig(g.configSource, g.sourceVdiskID)
	if err != nil {
		return 0, errors.Wrap(err, "failed to ReadNBDStorageConfig")
	}

	// create (primary) storage cluster
	// TODO: support optional slave cluster here
	// see: https://github.com/zero-os/0-Disk/issues/445
	cluster, err := ardb.NewCluster(storageConf.StorageCluster, nil) // not pooled
	if err != nil {
		return 0, errors.Wrapf(err,
			"cannot create storage cluster model for primary cluster of vdisk %s",
			g.sourceVdiskID)
	}

	indices, err := storage.ListBlockIndices(g.sourceVdiskID, staticConf.Type, cluster)
	if err != nil {
		return 0, errors.Wrapf(err, "ListBlockIndices failed for vdisk `%v`", g.sourceVdiskID)
	}
```

And that's not the only place where we have to do this common task (well there's only 2 places where we do this for now, but still...).

So we should probably add a wrapper function for this primitive to the storage code, which could have the following signature:

```go
// package storage

// wraps around the original primitive (ListBlockIndicesInCluster)
func ListBlockIndices(vdiskID string, cs config.Source) ([]int64, error)

// the original primitive has been renamed from ListBlockIndices to ListBlockIndicesInCluster
func ListBlockIndicesInCluster(vdiskID string, vt config.VdiskType, cluster ardb.StorageCluster) ([]int64, error)
```

This way in places like this you could replace code such as the one above with:

```go
indices, err := storage.ListBlockIndices(g.sourceVdiskID, g.configSource)
if err != nil {
	return 0, errors.Wrapf(err,
            "couldn't generate tlog data as ListBlockIndices failed for vdisk `%s`", g.sourceVdiskID)
}
```

On top of using this new function where applicable, you'll also have to make sure to rename the places that currently use `storage.ListBlockIndices`, to make them use `storage.ListBlockIndicesInCluster` instead (incase they shouldn't be using the new wrapper function that is).

nbd/ardb/backup/export.go
```diff
@@ -23,7 +23,7 @@ func Export(ctx context.Context, cfg Config) error {
 		return err
 	}
 
-	storageConfig, err := createStorageConfig(cfg.VdiskID, cfg.BlockStorageConfig)
+	storageConfig, err := createStorageConfig(cfg.VdiskID, cfg.ConfigSource)
```
GlenDC: I thought when you added this new optional `cfg.ConfigSource` I would have seen code at the top of import/export, such as:

```go
// create optional ConfigSource if it wasn't given
if cfg.ConfigSource == nil {
   storageConfigCloser, err := config.NewSource(cfg.BlockStorageConfig)
    if err != nil {		
        return nil, err		
    }		
    defer storageConfigCloser.Close()
    cfg.ConfigSource = storageConfigCloser
}
```

Because otherwise you're just ignoring an already existing API, by ignoring an existing (previously required) property, which would be very bad if there were actually other people using this code. Luckily for you we're the only user, so it's not too bad to do so here.

Either way, you should delete that old `cfg.BlockStorageConfig` property.

tlog/copy/generator.go
GlenDC: You can also provide similar changes for the `storage.VdiskExists` and `storage.DeleteVdisk` primitives, as their more common use case is that we just use the clusters they've configured. Still having the the API to do it from any cluster should be possible, hence we'll also rename the original primitives by adding the `InCluster` suffix.

These changes will be very useful! Thank you for doing this 👍

GlenDC:
Last added commits look good.

config/source_once_test.go
```diff
@@ -0,0 +1,127 @@
...
+	require.Panics(func() {
+		source.Watch(nil, Key{})
+	}, "calling Watch on a onceSource should cause a panic")
+}
+
+func TestDifferentKeyTypeSameID(t *testing.T) {
+	require := require.New(t)
+
+	// create stubsource with data
+	vdiskID := "testVdisk"
```
GlenDC: turn these into constants if you're not planning to mutate them

nbd/ardb/backup/backup.go
```diff
@@ -370,19 +365,13 @@ func isNilBlock(block []byte) bool {
...
 // Create a block storage ready for importing/exporting to/from a backup.
-func createStorageConfig(vdiskID string, sourceConfig config.SourceConfig) (*storageConfig, error) {
-	storageConfigCloser, err := config.NewSource(sourceConfig)
-	if err != nil {
-		return nil, err
-	}
-	defer storageConfigCloser.Close()
-
-	vdiskConfig, err := config.ReadVdiskStaticConfig(storageConfigCloser, vdiskID)
+func createStorageConfig(vdiskID string, configSource config.Source) (*storageConfig, error) {
```
GlenDC: You can delete this function, it is no longer needed. The only thing we still need is the StaticVdiskConfig, and that is something we can do directly in the `Import/Export` functions, no need to have a separate function for that any longer.

When you refactor/change code, also make sure to see what code you can omit. Trust me it shows more skill to be able to delete code, than it is to add code. And this PR is a great opportunity to remove quite some code.

nbd/ardb/backup/export.go
```diff
@@ -37,7 +37,7 @@ func Export(ctx context.Context, cfg Config) error {
 	}
 
 	log.Debugf("collecting all stored block indices for vdisk %s, this might take a while...", cfg.VdiskID)
-	indices, err := storage.ListBlockIndices(cfg.VdiskID, storageConfig.Vdisk.Type, storageCluster)
+	indices, err := storage.ListBlockIndicesInCluster(cfg.VdiskID, storageConfig.Vdisk.Type, storageCluster)
```
GlenDC: No need to use `storage.ListBlockIndicesInCluster`, just use `storage.ListBlockIndices`. Perhaps you think you want to use `storage.ListBlockIndicesInCluster` because you see that the pool was being used to create the cluster here. But don't worry, that pool is really not needed for this use case. So simply using the easier-to-use `storage.ListBlockIndices` should be fine here.

nbd/ardb/storage/storage.go
```diff
@@ -79,33 +79,25 @@ func BlockStorageFromConfigSource(vdiskID string, cs config.Source, dialer ardb.
...
-	return BlockStorageFromConfig(vdiskID, *vdiskConfig, *nbdStorageConfig, dialer)
-}
-
-// BlockStorageFromConfig creates a block storage from the given config.
-// It is the simplest way to create a BlockStorage,
-// but it also has the disadvantage that
-// it does not support SelfHealing or HotReloading of the used configuration.
-func BlockStorageFromConfig(vdiskID string, vdiskConfig config.VdiskStaticConfig, nbdConfig config.NBDStorageConfig, dialer ardb.ConnectionDialer) (BlockStorage, error) {
-	err := vdiskConfig.Validate()
+	err = vdiskConfig.Validate()
```
GlenDC: You can remote this validate call, this is not needed when you read it yourself from the source config, as it is guaranteed when creating a config (from source/bytes) that it is validated for us already before returning it to us.

When refactoring code, make sure to understand it. If you don't understand it yet, read through it and soak it in.

nbd/ardb/storage/storage.go
```diff
@@ -79,33 +79,25 @@ func BlockStorageFromConfigSource(vdiskID string, cs config.Source, dialer ardb.
...
 	if err != nil {
 		return nil, err
 	}
-	err = nbdConfig.Validate()
+	err = nbdStorageConfig.Validate()
```
GlenDC: This validate call is not neither anymore either, same as previous comment.

nbd/ardb/storage/storage.go
```diff
@@ -173,16 +165,51 @@ func NewBlockStorage(cfg BlockStorageConfig, cluster, templateCluster ardb.Stora
 
 // VdiskExists returns true if the vdisk in question exists in the given ARDB storage cluster.
 // An error is returned in case this couldn't be verified for whatever reason.
-func VdiskExists(id string, t config.VdiskType, cluster ardb.StorageCluster) (bool, error) {
+// Also return vdiskType and ardb cluster from config
+func VdiskExists(vdiskID string, source config.Source) (bool, config.VdiskType, *ardb.Cluster, error) {
```
GlenDC: Please don't do that, this overcomplicates the API for no good reason. It should only return `(bool, error)`. If you wish to reuse cluster and configs, than you shouldn't be using this function, as this function is meant for places where you just quickly need the cluster and configs just for this call.

I know that we also use it in places where this is not the case, but those cases are the zeroctl, and we honestly not going to care too much that we're creating a cluster object multiple times there. And as the config is cached in those instances, we also don't care that we read and parse an object multiple times. All that is really neglectable compared to the actual cost of network communication.

nbd/ardb/storage/storage.go
```diff
@@ -240,28 +267,52 @@ func CopyVdisk(source, target CopyVdiskConfig, sourceCluster, targetCluster ardb
 
 // DeleteVdisk returns true if the vdisk in question was deleted from the given ARDB storage cluster.
 // An error is returned in case this couldn't be deleted (completely) for whatever reason.
-func DeleteVdisk(id string, t config.VdiskType, cluster ardb.StorageCluster) (bool, error) {
+func DeleteVdisk(vdiskID string, configSource config.Source) (bool, error) {
```
GlenDC: this function should also delete the data from the slave cluster if it is defined for the vdisk in question. You can simply merge the 2 cluster config objects together in that case, so that you still only need one cluster and one DeleteVdiskInCluster call

zeroctl/cmd/backup/import.go
```diff
@@ -170,7 +136,7 @@ func checkVdiskExists(vdiskID string) error {
 	}
 
 	// delete vdisk, as it exists and `--force` is specified
-	deleted, err := storage.DeleteVdisk(vdiskID, staticConfig.Type, cluster)
+	deleted, err := storage.DeleteVdiskInCluster(vdiskID, vdiskType, cluster)
```
GlenDC: Simply use `DeleteVdisk` here, remember we cache the configs in this scenario, and creating a cluster object is very  cheap, so really don't worry about that.

zeroctl/cmd/backup/import.go
```diff
@@ -123,42 +122,9 @@ func importVdisk(cmd *cobra.Command, args []string) error {
...
-	if err != nil {
-		return errors.Wrapf(err,
-			"cannot create storage cluster model for cluster %s",
-			nbdStorageConfig.StorageClusterID)
-	}
-
+func checkVdiskExists(vdiskID string, configSource config.Source) error {
 	// check if vdisk exists
-	exists, err := storage.VdiskExists(vdiskID, staticConfig.Type, cluster)
+	exists, vdiskType, cluster, err := storage.VdiskExists(vdiskID, configSource)
```
GlenDC: This is a bad example of designing an API just to fit one specific scenario. That would be fine if it's just an internal (helper) function within this file, but it's not. `storage.VdiskExists` is a very generic storage function not just meant for this specific scenario, even though it might be only used here for now. `VdiskExists` should only ever return `(bool, error)`. If you want to share clusters and configs between different methods, we have the `InCluster` versions for those cases.

But for zeroctl we want to keep things as simple as possible, as it is not meant to have any real logic, except for calling the Golang API for the user.

zeroctl/cmd/delvdisk/vdisk.go
```diff
@@ -30,42 +28,14 @@ func deleteVdisk(cmd *cobra.Command, args []string) error {
...
 
-	// create config source
 	source, err := config.NewSource(vdiskCmdCfg.SourceConfig)
-	if err != nil {
-		return err
-	}
 	defer source.Close()
+	configSource := config.NewOnceSource(source)
 
-	argn := len(args)
```
GlenDC: Why did you delete these pos-arg validations? o.O They are still required. Did you delete them by mistake? Even worse your code as it is can result in a (OOB) panic.

zeroctl/cmd/delvdisk/vdisk.go
```diff
@@ -30,42 +28,14 @@ func deleteVdisk(cmd *cobra.Command, args []string) error {
...
-		return err
-	}
-
-	cluster, err := ardb.NewCluster(*clusterConfig, nil)
-	if err != nil {
-		return err
-	}
-
-	_, err = storage.DeleteVdisk(vdiskID, staticCfg.Type, cluster)
+	_, err = storage.DeleteVdisk(vdiskID, configSource)
```
GlenDC: I love how simple this file now looks :) Perfect.

zeroctl/cmd/restore/vdisk.go
```diff
@@ -116,7 +89,7 @@ func checkVdiskExists(vdiskID string, configSource config.Source) error {
 	}
 
 	// delete vdisk, as it exists and `--force` is specified
-	deleted, err := storage.DeleteVdisk(vdiskID, staticConfig.Type, cluster)
+	deleted, err := storage.DeleteVdiskInCluster(vdiskID, vdiskType, cluster)
```
GlenDC: Simply use `DeleteVdisk` here.

zeroctl/cmd/restore/vdisk.go
```diff
@@ -125,7 +98,7 @@ func checkVdiskExists(vdiskID string, configSource config.Source) error {
 	}
 
 	// delete 0-Stor (meta)data for this vdisk
-	if staticConfig.Type.TlogSupport() {
+	if vdiskType.TlogSupport() {
```
GlenDC: You do also want to on top of that read the staticVdiskConfig for this part of the code, but as the config is cached, that's a fairly cheap thing to do. Certainly 100% neglectable compared to the real work this command has to do.

zeroctl/cmd/restore/vdisk.go
```diff
@@ -77,34 +76,8 @@ func restoreVdisk(cmd *cobra.Command, args []string) error {
 // checkVdiskExists checks if the vdisk in question already/still exists,
```
GlenDC: Make sure to change:

```go
// create config source
configSource, err := config.NewSource(vdiskCmdCfg.SourceConfig)
if err != nil {
	return err
}
defer configSource.Close()
```

to:

```go
// create config source
cs, err := config.NewSource(vdiskCmdCfg.SourceConfig)
if err != nil {
	return err
}
defer cs.Close()
configSource := config.NewOnceSource(cs)
```

zeroctl/cmd/backup/import.go
```diff
@@ -46,28 +46,27 @@ func importVdisk(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	err = checkVdiskExists(vdiskCmdCfg.VdiskID)
+	source, err := config.NewSource(vdiskCmdCfg.SourceConfig)
```
GlenDC: don't ignore this returned error, make sure to handle it!

nbd/ardb/storage/storage.go
chrisvdg: I don't fully get what needs to happen, and I'm running out of time to work on this today so will leave this one to you!

chrisvdg:
Your comments have been applied as best I could on this Saturday morning and zeroctl commands should now be using onceConfig.
The only thing I couldn't get done was to make `storage.DeleteVdisk` also delete the data from the slave cluster if it is defined for the given vdiskID, as commented before.

GlenDC:
Thank you very much for this effort @chrisvdg!

Don't worry, I'll take it from here. I'll add any needed changes in a commit myself, and merge it in afterwards. Make sure to review the commit when you get to it Monday, so it can hopefully have some more educational value.

PR #606, opened 2017-11-13
Disk buffers

### LINKED ISSUE #351: proposal: dirty block cache
Right now we send all blocks directly to the `ardb` servers. We do this for all storage types (`deduped`, `nondeduped` and `semideduped`). I propose we instead store them in a dirty block cache instead, which would only be written to by write actions (`delete` and `set`). When `getting` a block, it will be checked if the index is available in the cache, and if so we can simply return that one. The cache would be emptied either because a call to the `Flush` method, or because the cache has reached its limit, in which case all content is flushed to the `ardb` servers as well. This should allow us to save some roundtrips.

muhamadazmy:
Fixes #351

> This is only done for the deduped disks atm, once the approach is approved it can be done for nondeduped disks as well.

This introduce to ideas to improve the disk write speed.
- A disk buffer (currently set to 5MB) we probably should make it configurable per disk
- A multiple writers to commit buffered blocks to the backend storage (ardb)

## How it works
A limited size cache is attached to a disk, the disk items are flushed if they expire or the buffer is getting full using a LRU mechanism.

A workers pool is started, which handles all evict operation from the cache (due to buffer limit reached, or expiration), as fast as possible.

## Naive performance tests 
I tried to compare the performance between `master` and this branch using a simple `dd` command. I can see a slight improvement in performance. But a better perftests should be done
I noticed that writing a file first time goes much faster than rewriting (on both master and this branch)

```bash
# BUFFERED WRITES (new file, test didn't exist)
➜  root sudo dd if=/dev/urandom of=test bs=2MB count=500
500+0 records in
500+0 records out
1000000000 bytes (1.0 GB, 954 MiB) copied, 4.68936 s, 213 MB/s

# BUFFERED (RE)WRITES (file `test` already exists)
➜  root sudo dd if=/dev/urandom of=test bs=2MB count=500
500+0 records in
500+0 records out
1000000000 bytes (1.0 GB, 954 MiB) copied, 44.2696 s, 23.6 MB/s

# NON-BUFFERD  (master branch) (new file)
➜  root sudo dd if=/dev/urandom of=test2 bs=2MB count=500
500+0 records in
500+0 records out
1000000000 bytes (1.0 GB, 954 MiB) copied, 7.39393 s, 135 MB/s

# NON-BUFFERD (master branch) (file exists)
➜  root sudo dd if=/dev/urandom of=test bs=2MB count=500
             
500+0 records in
500+0 records out
1000000000 bytes (1.0 GB, 954 MiB) copied, 45.1192 s, 22.2 MB/s
```


GlenDC:
Travis still fails...

nbd/ardb/storage/cache.go
```diff
@@ -0,0 +1,106 @@
...
+
+import (
+	"context"
+	"github.com/bluele/gcache"
+	"github.com/zero-os/0-Disk"
+	"math"
+	"time"
+)
+
+const (
```
GlenDC: Even though these might be sane defaults, I think they should also be configurable somehow. Especially the cache size.

nbd/ardb/storage/cache.go
```diff
@@ -0,0 +1,106 @@
...
+	CacheSize              = 5 * 1024 * 1024 //CacheSize in MB
+	cacheDefaultExpiration = 10 * time.Second
+	cacheDefaultCleanup    = 5 * time.Second
+	cacheDefaultMaxSize    = 100 //each chunk is 512K so this is around 2.5M of memory
```
GlenDC: how you compute that each chunk is 512KiB? Also I'm fairly sure that 100*512KiB = 50MiB, and no 2.5 MiB. So I'm not sure where you get your numbers. Either that or your comments really don't help me understand these values at all.

nbd/ardb/storage/cache_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func TestCacheTimeLimit(t *testing.T) {
+	key := zerodisk.Hash("test-key")
+
+	cache := NewCache(nil, 2*time.Second, 0, 5)
+
+	cache.Set(key, nil)
+
+	<-time.After(3 * time.Second)
```
GlenDC: why you go so fancy, just use a time.Sleep in this case... Even though, I would really prefer if we could test this without having to sleep at all.

nbd/ardb/storage/cache_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	if ok := assert.Equal(t, 0, cache.Count()); !ok {
```
GlenDC: use `require.Equal` if you want to do a fatal in case of failure

nbd/ardb/storage/deduped.go
```diff
@@ -180,15 +188,34 @@ func (ds *dedupedStorage) getPrimaryOrTemplateContent(hash zerodisk.Hash) (conte
 	return
 }
 
+func (ds *dedupedStorage) commitBlock(in interface{}) interface{} {
+	request := in.(*writeRequest)
+	hash := request.hash
+	cmd := ardb.Command(command.Set, hash.Bytes(), request.content)
+	if err := ardb.Error(ds.cluster.DoFor(int64(hash[0]), cmd)); err != nil {
```
GlenDC: why are these errors not propagated?

nbd/ardb/storage/deduped.go
```diff
@@ -180,15 +188,34 @@ func (ds *dedupedStorage) getPrimaryOrTemplateContent(hash zerodisk.Hash) (conte
...
+		log.Errorf("couldn't store content in primary/slave storage: %s", err.Error())
+	}
+
+	return nil
+}
+
+func (ds *dedupedStorage) evictCache(hash zerodisk.Hash, content []byte) {
+	if err := ds.wpool.Do(&writeRequest{hash: hash, content: content}, nil); err != nil {
```
GlenDC: why are these errors not propagated?

nbd/ardb/storage/pool.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	p.m.Unlock()
+	p.ch <- payload{in, cb}
+	return nil
+}
+
+func (p *Pool) apply(work payload) {
+	defer func() {
+		if err := recover(); err != nil {
```
GlenDC: we really don't want to recover, ever, ...
If there's a panic, it means you really have a bug in your code that should be fixed ASAP.

nbd/ardb/storage/pool.go
```diff
@@ -0,0 +1,146 @@
+package storage
```
GlenDC: There is not much in this pool.go file that I like. I don't see why we need it, it's very generic, and I honestly don't see why we need it (as to repeat myself)...

nbd/ardb/storage/pool.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"context"
+	"fmt"
+	"github.com/zero-os/0-Disk/log"
+	"runtime"
+	"sync"
+)
+
+type Work func(in interface{}) (out interface{})
+type Callback func(out interface{})
```
GlenDC: These names are really not good. You're adding these public types to a package called storage. Thus I suddenly see a name `storage.Callback`. From that name, what the hell is that callback type supposed to do? Same for the Work func..

nbd/ardb/storage/pool_test.go
```diff
@@ -0,0 +1,65 @@
...
+			wg.Done()
+			if err := p.Do(i, nil); err != nil {
+				t.Fatal(err)
+			}
+		}(i)
+	}
+
+	wg.Wait() //wait until we sure all go routines started
+	//extra wait just to be sure
+	<-time.After(100 * time.Millisecond)
```
GlenDC: If you need to sleep, it really start to smell. No matter how big this time is, this will eventually fail on travis or any other slow machine. Why is this needed?

nbd/ardb/storage/cache.go
muhamadazmy: Ah, apparently i changed the value and didn't update the comment, sorry

nbd/ardb/storage/cache_test.go
muhamadazmy: Sleep locks the full thread, which makes go start another thread to run the routines in runnable state. It's encourage to use time.After() to avoid this issue and allow the thread to be reused by the first runnable routine.

nbd/ardb/storage/deduped.go
muhamadazmy: One of the things i needed to discuss with you. This entire commitBlock runs inside a worker (go routine) and we can't actually (return) this error to anyone, unless we can push it to some error handler. 

I used the same approach done here https://github.com/zero-os/0-Disk/pull/606/files/bfcae70dc6bab364114f2348790038237fae4e08#diff-db84061ee48f242621e14c6dc6e60e55L164
to log the error.

nbd/ardb/storage/pool.go
muhamadazmy: Yeah, i agree. I first thought was to place this under a different utilities package. But 0-disk didn't have one, and wasn't sure we should create a new package for just this one. May be we change it all to private types? or move it. Your call.

nbd/ardb/storage/pool.go
muhamadazmy: The workers pool is used to have parallel writers to the ardb storage from the cache. so we don't slow down the write operations (as possible) when your cache is full. Once the cache is full it will start evicting the least accessed blocks. If the eviction speed is not good enough your call to SetBlock will get blocked until a slot is available in the cache.

The worker tries to ensure this by having multiple (managed) workers trying to handle all the evicted blocks. But it also granted that all "scheduled" blocks are processed before the disk Close() returns.

nbd/ardb/storage/pool.go
muhamadazmy: True. I will remove it.

nbd/ardb/storage/pool_test.go
muhamadazmy: it's how the test case is written that require this, i will change the test case to avoid sleep.

nbd/ardb/storage/cache.go
muhamadazmy: Yeah, i already suggested that in the PR description

nbd/ardb/storage/deduped.go
muhamadazmy: The thing is, SetBlock succeed once the block is pushed to the buffer, this will never fail. But when a block is about to be evicted due to a timeout or a full buffer, the evictMethod will not able to return the error the SetBlock caller which has already returned of course.

muhamadazmy:
@GlenDC Your review has raised few concerns after introducing disk buffers that we need to discuss 
- When exactly SetBlock should return? when the block is written to the cache, or when the block is written to actual physical disk (in our case ardb)?
- If the writing to buffer had succeeded, but the actual writing to physical storage had failed, what should be done? This is strongly related to the first question.
- Does nbd call SetBlock concurrently, or sequentially (for multiple blocks)?

Once having solid answers to the above questions, we can decide if we can:
- Support disk buffers in the first place, since disk buffers comes with the possibility of data loss (in case of power failure)
- Have pool of flushers to handle concurrent calls to `SetBlock`
- Or Have a mixture of both as we have now (we probably need to handle ardb write failures differently not just log)

GlenDC:
Going to close PR, but let's keep the branch alive for now.

PR #613, opened 2017-11-16
ndb with tlog : handle failed ardb.

iwanbk:
Parts of #491 
- add `storageFailed` flag, true if the storage considered in failed state.
- write operation will return `syscall.EIO` in case of `storageFailed` is false

Notes:
- ` waitStorageHealth` https://github.com/zero-os/0-Disk/commit/905d6751de6054855a6b09d4287aa5f7c9fa84af#diff-fb7477f5962f72d74355e5c822fae9d4R580 might not elegant, but can't find better way to do this
- `inMemorySequenceCache.Evict` now returns element ordered by sequence, so we have valid `metadata.LastFlushedSequence`

### DIFF
--- a/nbd/nbdserver/tlog/tlog.go
+++ b/nbd/nbdserver/tlog/tlog.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/zero-os/0-Disk/config"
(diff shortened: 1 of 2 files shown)


GlenDC:
For now I think we can just freeze this PR, or even close it. As we won't continue work on this repo, until we're happy with the other layers of zero-os. After that we'll do a big review and redesign of the parts that should be changed within 0-Disk.

When integrating the new tlog repo, we'll for example going to also see if we can do a big redo of tlogStorage and see how we can do better.

GlenDC:
Closing this PR for now. Most of this code might not be needed anymore once we switch over to the standalone tlog repo. We'll keep the branch alive for now, for reference reasons, will delete it in the future once we really know for sure we won't need it anymore.

Or perhaps we can re-open this PR if we do decide that we might want to have it in master (in some form or another), even _after_ switching to the new tlog code.

But for now I would keep this branch alive.

